PR #5, opened 2018-12-14
Adds smtp pipelining test

catenacyber:
I created a pcap with postfix as a server and a dummy python script as a client.
I did not find a one-liner client with pipelining support : curl does not seem to support smtp pipelining for instance

This test fails without merging https://github.com/OISF/suricata/pull/3568 (and succeeds afterwards) because of email.status

The dummy script is 
```
import socket

sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
sock.connect(("127.0.0.1", 25))

data = sock.recv(2000)
print "1", data
a = sock.send("EHLO ehlo.fr\r\n")
data = sock.recv(2000)
print "2", data
a = sock.send("MAIL FROM:<username@domain.com>\r\nRCPT TO:<john.doe@example.com>\r\nDATA\r\nThis is a test\r\n.\r\n")
data = sock.recv(2000)
print "3", data

sock.close()
```


victorjulien:
Merged in #8

PR #9, opened 2019-02-15
smtp: add raw extraction test

maurizioabba:
pcap downloaded from https://osqa-ask.wireshark.org/questions/33094/extract-an-attachment-email-smtp-cap


tests/smtp-raw-extraction/suricata.yaml
```diff
@@ -0,0 +1,54 @@
+%YAML 1.1
+---
+
+include: ../../etc/suricata-4.0.3.yaml
```
victorjulien: this should not be necessary (in fact I think we should remove it from the instructions)

tests/smtp-raw-extraction/test.yaml
```diff
@@ -0,0 +1,10 @@
+requires:
+  features:
+    - HAVE_NSS
+  files:
+    - src/output-filestore.c
+
```
victorjulien: think we need a 'min-version: 5.0.0' here in the requires, otherwise this will break for 4.0 and 4.1

victorjulien:
Can you add a README with a link to the pcap source?

maurizioabba:
Replaced by https://github.com/OISF/suricata-verify/pull/11

PR #10, opened 2019-02-15
eve/http: add request/response headers test

maurizioabba:
pcap downloaded from https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=http.cap


tests/http-all-headers/suricata.yaml
```diff
@@ -0,0 +1,13 @@
+%YAML 1.1
+---
+
+include: ../../etc/suricata-4.0.3.yaml
```
victorjulien: shouldn't be necessary

tests/http-all-headers/test.yaml
```diff
@@ -0,0 +1,9 @@
+requires:
```
victorjulien: min-version: 5.0.0

tests/http-all-headers/check.sh
```diff
@@ -0,0 +1,21 @@
+#! /bin/sh
```
victorjulien: @jasonish could this be expressed in the test.yaml itself instead?

victorjulien:
Can you add a README that links to the pcap source?

maurizioabba:
Replaced by https://github.com/OISF/suricata-verify/pull/12

PR #13, opened 2019-02-21
Windows/v3

victorjulien:
Fixes and updates to support running on Windows. Tested with mingw32 and mingw64 on python2 and python3.

Also adds test for Suricata bug 2769.


run.py
```diff
@@ -621,7 +638,10 @@ def default_args(self):
         # Find rules.
         rules = glob.glob(os.path.join(self.directory, "*.rules"))
         if not rules:
-            args += ["-S", "/dev/null"]
+            if sys.platform == "win32":
+                args += ["--disable-detection"]
+            else:
+                args += ["-S", "/dev/null"]
```
jasonish: Any reason not to just use --disable-detection for both cases here?

victorjulien: We could do that. The code paths are different though, so I guess it makes some sense to test both. In the past we've had large perf differences between the opts.

victorjulien: I guess for the purpose of suricata-verify we can make this change indeed.

victorjulien: See #27

victorjulien:
Replaced by #27

PR #17, opened 2019-03-02
Suricata-verify tests for http.{location,server}

jlucovsky:
Add test cases to suricata-verify for new HTTP sticky buffers: location and server


victorjulien:
Merged through #48, thanks Jeff!

PR #19, opened 2019-03-04
Adds a testcase for HTTP multiple content-length

catenacyber:
This follows #17 
This is still not right yet.

We get the expected behavior here.
Except if we turn off the `to_server` keyword, we have two alerts : a good one and a strange one...


catenacyber:
Ok, I think I got it.
app-layer-event are set by transaction, so they are both ways.
So it is normal to get two alerts if we do not set the `to_server` keyword.

For #18, we do not get the alert because we start a second request before finishing receiving the first response. So it looks like a pipelining problem

PR #22, opened 2019-03-05
Adds a testcase for http invalid request line

catenacyber:
From https://redmine.openinfosecfoundation.org/issues/2655


tests/http-request-invalid/check.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/sh
+
+. ${TOPDIR}/util/functions.sh
+
+# Should have two alerts sid 2221034.
+n=$(jq_count eve.json 'select(.alert.signature_id == 2221034)')
```
victorjulien: test conditions in test.yaml are preferred. Only if the logic can't be expressed there, a check script should be used.

tests/http-request-invalid/test.rules
```diff
@@ -0,0 +1,58 @@
...
+# Request line started with whitespace
+alert http any any -> any any (msg:"SURICATA HTTP Request line with leading whitespace"; flow:established,to_server; app-layer-event:http.request_line_leading_whitespace; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221031; rev:1;)
+
+
+alert http any any -> any any (msg:"SURICATA HTTP Request too many encoding layers"; flow:established,to_server; app-layer-event:http.too_many_encoding_layers; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221032; rev:1;)
+alert http any any -> any any (msg:"SURICATA HTTP Request abnormal Content-Encoding header"; flow:established,to_server; app-layer-event:http.abnormal_ce_header; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221033; rev:1;)
+
+alert http any any -> any any (msg:"SURICATA HTTP Invalid Request line"; flow:established,to_server; app-layer-event:http.request_line_invalid; flowint:http.anomaly.count,+,1; classtype:protocol-command-decode; sid:2221034; rev:1;)
```
victorjulien: can we have a rules file with just that one rule that you determined is supposed to match?

tests/http-request-invalid/test.yaml
```diff
@@ -0,0 +1,7 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+
```
victorjulien: we should set the min-version here, probably 4.1.0 as we'll backport a fix to the master-4.1.x branch in Suricata.

tests/http-request-invalid/check.sh
catenacyber: Ok, I got inspired by dnp3-dnp3_func-alert

tests/http-request-invalid/test.rules
catenacyber: Ok

tests/http-request-invalid/test.yaml
catenacyber: Ok

PR #25, opened 2019-03-05
Adds a testcase for flow_direction

fojac:
The addition of this testcase follows the request on the flow_direction addition (see https://github.com/OISF/suricata/pull/3521).
Hope it is OK for you, as it is my first PR.


tests/flow-direction/test.yaml
```diff
@@ -0,0 +1,23 @@
+command: |
```
victorjulien: with a simple pcap check like this, there is no need to use 'command'.

fojac:
Thanks, so I just have to keep the "checks" part and remove the "command" and "pre-check" lines?
So I understand the suricata.yaml as well as the .pcap should be removed too?

catenacyber:
> So I understand the suricata.yaml as well as the .pcap should be removed too?
The pcap file should be kept.
I think that the suricata.yaml can be removed.

You can add a README.md file to describe the test and explain where the pcap file comes from (or how we can create it from scratch)

tests/flow-direction/test.yaml
```diff
@@ -0,0 +1,19 @@
+pre-check: |
```
catenacyber: I think that we do not need this

fojac:
The README.md file was added. The flow direction feature should create the flow direction metadata in nearly any TCP flow. So the PCAP is just there as a test, any usual TCP flow pcap should properly work with this feature, I just added one as requested for the suricata MR of this feature to be accepted.

catenacyber:
> The README.md file was added.

The readme looks clear to me :-)

catenacyber:
@fojac could you rebase this with a minimal yaml file as requested ?

catenacyber:
ping @fojac ?

fojac:
It seems to me the flow_direction feature doesn't work anymore in suricata 7.0 dev.
I've tried to add the patch in output-json.c cloning the latest version of suricata but in eve.json I don't see the flow_direction anymore, so the check fails.

catenacyber:
> doesn't work anymore in suricata 7.0 dev

It does neither with 6.0 or 5.0
Is there a suricata PR related to this ?

fojac:
There was one (https://github.com/OISF/suricata/pull/3521) but now closed.

jasonish:
Closing as it looks like the feature this test is for has gone stale with no recent work. Please re-open if this is not true. Thanks.

PR #28, opened 2019-03-30
FTP active/passive mode file extraction tests

jlucovsky:
This changeset adds test cases for FTP active and passive mode using
the pcaps from issue [2527](https://redmine.openinfosecfoundation.org/issues/2527)

This PR verifies functionality added in https://github.com/OISF/suricata/pull/3758.


tests/filestore-ftp-active-mode/ftp.rule
```diff
@@ -0,0 +1 @@
+alert tcp any 20 -> any any (msg:"File Found within FTP and stored"; filestore;  sid:1; rev:1;)
```
victorjulien: if unused it can be removed

tests/filestore-ftp-active-mode/suricata.yaml
```diff
@@ -0,0 +1,12 @@
+%YAML 1.1
+---
+
+include: ../../etc/suricata-4.0.3.yaml
```
victorjulien: shouldn't be needed

victorjulien:
Can you also make sure the pcaps only contain the ftp traffic we care about?

victorjulien:
These will also need: 'requires' to make sure we only run this on 5.0.0 for passive, 4.1.x+ for active (or other way around).

Next I think you need to pass
```
args:
  - --runmode=single
```
as the pcap processing of the control flow and data flow can be so fast that there is a race between them.

tests/filestore-ftp-active-mode/suricata.yaml
jlucovsky: Will remove.

tests/filestore-ftp-active-mode/ftp.rule
jlucovsky: Will remove.

jlucovsky:
> I'd prefer file-store v2 tests, could you update the tests?

Are you asking for the test to be renamed or something else?

victorjulien:
> > I'd prefer file-store v2 tests, could you update the tests?
> 
> Are you asking for the test to be renamed or something else?

No, the file store output comes in 2 variants. An older one (v1) and a newer one (v2). I would prefer the test to use v2. https://github.com/OISF/suricata/blob/master/suricata.yaml.in#L460

jlucovsky:
> > > I'd prefer file-store v2 tests, could you update the tests?
> > 
> > 
> > Are you asking for the test to be renamed or something else?
> 
> No, the file store output comes in 2 variants. An older one (v1) and a newer one (v2). I would prefer the test to use v2. https://github.com/OISF/suricata/blob/master/suricata.yaml.in#L460

Will ensure v2 is used.

jlucovsky:
> Can you also make sure the pcaps only contain the ftp traffic we care about?

Will update both pcaps.

jlucovsky:
Continued in #69

PR #33, opened 2019-04-04
alert: verify test for distance/within keywords

jmtaylor90:
* this is based on https://redmine.openinfosecfoundation.org/issues/1


tests/alert-distance-within-1/suricata.yaml
```diff
@@ -0,0 +1,17 @@
+%YAML 1.1
+---
+
+include: ../../etc/suricata-4.0.3.yaml
```
victorjulien: this shouldn't be necessary

tests/alert-distance-within-1/test.yaml
```diff
@@ -0,0 +1,11 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+
+checks:
+
+  # Check that we only have one alert event type in eve.
```
victorjulien: this checks that there are 2 alert records, but doesn't say anything about other record types

jmtaylor90: Ah, good point, I will update. Thanks!

tests/alert-distance-within-1/suricata.yaml
jmtaylor90: will fix

jmtaylor90:
followed up in #36

PR #38, opened 2019-04-15

shivan1b:
appropriate message.

Sample conf
-----------
```
requires:
  features:
    - HAVE_LIBJANSSON

checks:

  # Check that we only have one alert event type in eve.
  - filter:
      version: 4.0.2
      count: 1
      match:
        event_type: alert

  - shell:
      expect: 1
```

version 5.0.0-dev.
```
===> alert-testmyids: SKIPPED: Suricata v4.0.2 not found
```

```
```

### DIFF
--- a/run.py
+++ b/run.py
@@ -112,43 +112,6 @@ def get_suricata_version():
     output = subprocess.check_output(["./src/suricata", "-V"])
     return parse_suricata_version(output)
 
-def version_equal(a, b):
-    """Check if version a and version b are equal in a semantic way.
-
-    For example:
-      - 4 would match 4, 4.x and 4.x.y.
-      - 4.0 would match 4.0.x.
-      - 4.0.3 would match only 4.0.3.
-    """
-    if not a.major == b.major:
-        return False
-
-    if a.minor is not None and b.minor is not None:
-        if a.minor != b.minor:
-            return False
-
-    if a.patch is not None and b.patch is not None:
-        if a.patch != b.patch:
-            return False
-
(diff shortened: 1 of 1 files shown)


shivan1b:
@inliniac please check if this is in accordance to your request.

victorjulien:
When I add:
```patch
diff --git a/tests/alert-testmyids/test.yaml b/tests/alert-testmyids/test.yaml
index b6ce41d..6c8834f 100644
--- a/tests/alert-testmyids/test.yaml
+++ b/tests/alert-testmyids/test.yaml
@@ -1,6 +1,7 @@
 requires:
   features:
     - HAVE_LIBJANSSON
+  min-version: 5.0.0
 
 checks:
 
```
I get:
```
===> alert-testmyids: Traceback (most recent call last):
  File "../suricata-verify/run.py", line 804, in <module>
    sys.exit(main())
  File "../suricata-verify/run.py", line 777, in main
    if test_runner.run():
  File "../suricata-verify/run.py", line 488, in run
    self.check_requires()
  File "../suricata-verify/run.py", line 414, in check_requires
    suri_version=suri_version, expr="gte"):
  File "../suricata-verify/run.py", line 239, in is_version_compatible
    if not func(suri_version, config_version):
TypeError: unbound method is_gte() must be called with Version instance as first argument (got SuricataVersion instance instead)
```

shivan1b:
I'm sorry, I had not tested it with Python 2. Please check the new PR #39

PR #39, opened 2019-04-15

shivan1b:
appropriate message.

Sample conf
-----------
```
requires:
  features:
    - HAVE_LIBJANSSON

checks:

  # Check that we only have one alert event type in eve.
  - filter:
      version: 4.0.2
      count: 1
      match:
        event_type: alert

  - shell:
      expect: 1
```

version 5.0.0-dev.
```
===> alert-testmyids: SKIPPED: Suricata v4.0.2 not found
```

```
```

### DIFF
--- a/run.py
+++ b/run.py
@@ -112,43 +112,6 @@ def get_suricata_version():
     output = subprocess.check_output(["./src/suricata", "-V"])
     return parse_suricata_version(output)
 
-def version_equal(a, b):
-    """Check if version a and version b are equal in a semantic way.
-
-    For example:
-      - 4 would match 4, 4.x and 4.x.y.
-      - 4.0 would match 4.0.x.
-      - 4.0.3 would match only 4.0.3.
-    """
-    if not a.major == b.major:
-        return False
-
-    if a.minor is not None and b.minor is not None:
-        if a.minor != b.minor:
-            return False
-
-    if a.patch is not None and b.patch is not None:
-        if a.patch != b.patch:
-            return False
-
-    return True
-
-def version_gte(v1, v2):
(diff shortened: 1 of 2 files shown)


victorjulien:
Hmm based on the output it looks like we now skip the whole test if one of the filter checks versions doesn't match?

victorjulien:
I get:
```
===> alert-testmyids: SKIPPED: Suricata v4 not found
```
with
```patch
diff --git a/tests/alert-testmyids/test.yaml b/tests/alert-testmyids/test.yaml
index b6ce41d..e56a25e 100644
--- a/tests/alert-testmyids/test.yaml
+++ b/tests/alert-testmyids/test.yaml
@@ -6,6 +6,13 @@ checks:
 
   # Check that we only have one alert event type in eve.
   - filter:
+      version: 4
+      count: 1
+      match:
+        event_type: alert
+  # Check that we only have one alert event type in eve.
+  - filter:
+      version: 5
       count: 1
       match:
         event_type: alert
```
In this case the version: 4 check should have been skipped on the Suricata git master, but not the version: 5 check.

shivan1b:
Current logging system for all tests is like this. One test fails or is skipped, an exception is raised leaving the control of filter checks. I can change this. Is there a particular format you'd like to see the results in?
Current results are per test suite (incorrect term? Take `alert-testmyids` as an example to what I'm referring as a test suite) basis. The count of failed, skipped and passed tests is also determined per test suite. Should that be changed too?

victorjulien:
Yes this should be changed. The point of the feature request is to allow having a single 'test suite' that can define multiple check conditions that can be different for different Suricata versions.

PR #40, opened 2019-04-16
pcap: convert pcapng capture to pcap.

eroullit:
Some platforms have a libpcap without pcapng support activated.
In this situation, the tests fail because the capture file cannot be
opened.
Those tests do not require pcapng therefore they can be safely converted
to the pcap standard.


victorjulien:
Which platforms would this be? My openbsd qa builders work w/o issue.

victorjulien:
Nevermind, I see that my mingw has the same issue.

victorjulien:
It seems this fix is incomplete, it just removes the pcap-ng files, but does not replace them with the pcap ones.

eroullit:
Indeed. Here they are.

catenacyber:
Looks good to me now

victorjulien:
All pcaps are now old pcap format, and a travis check was added. So this PR is not relevant anymore. I guess I fixed it a while back as part of some other work, and forgot to include this PR.

PR #41, opened 2019-04-17
Adds test case for JSON anomaly logging

jlucovsky:
This changeset adds tests for JSON anomaly logging (part of [issue 2282](https://redmine.openinfosecfoundation.org/issues/2282)


tests/output-eve-anomaly/test.yaml
```diff
@@ -0,0 +1,26 @@
+requires:
```
victorjulien: requires should check for either: suri version 'min-version: 5.0.0' or the anomaly logger:
```
requires:
  features:
    - HAVE_LIBJANSSON
  files:
    - src/output-json-anomaly.c
```

tests/output-eve-anomaly/test.rules
```diff
@@ -0,0 +1 @@
+alert tcp any 20 -> any any (msg:"Simple rule"; filestore;  sid:31; rev:1;)
```
victorjulien: no need to add a rules file if you're not checking the results or it matching anyway

tests/output-eve-anomaly/test.yaml
```diff
@@ -0,0 +1,26 @@
...
+  features:
+    - HAVE_LIBJANSSON
+
+checks:
+  - filter:
+      filename: eve.json
```
victorjulien: no need to have this filename setting, eve.json is the default

tests/output-eve-anomaly/test.yaml
victorjulien: To clarrify: as-is the test will break on 4.1.x. It is supposed to be skipped on that version.

jlucovsky: Will add the a `files:` section.

tests/output-eve-anomaly/test.rules
jlucovsky: Will remove.

tests/output-eve-anomaly/test.yaml
jlucovsky: Will remove.

jlucovsky:
Continued in #45

PR #46, opened 2019-04-20

shivan1b:
appropriate message. This feature allows to run different underlying
tests individually and observe their individual outputs.

Sample conf
-----------
```
requires:
  features:
    - HAVE_LIBJANSSON

checks:
  - filter:
     version: 2
      count: 1
      match:
        event_type: alert
  - filter:
      version: 5.0
      count: 1
      match:
          event_type: alert

  - shell:
      expect: 1
```

Sample output
-------------
version 5.0.0-dev.
```
===> alert-testmyids:
|
 --> Sub test #1: SKIPPED: Suricata v2 not found

|
 --> Sub test #2: OK

|
 --> Sub test #3: OK
OK
```

Case handling
-------------
Case 1

```
```

Case 2

In case someone provides an old style `check.sh` as well as `checks` in
test.yaml, the test only performs the operations mentioned in tests.yaml
and ignores the presence of `check.sh`.

Case 3

In case someone provides a faulty `check.sh`, entire test fails.

This is a rework of #39

### DIFF
--- a/run.py
+++ b/run.py
@@ -112,43 +112,6 @@ def get_suricata_version():
     output = subprocess.check_output(["./src/suricata", "-V"])
     return parse_suricata_version(output)
 
-def version_equal(a, b):
-    """Check if version a and version b are equal in a semantic way.
-
-    For example:
-      - 4 would match 4, 4.x and 4.x.y.
-      - 4.0 would match 4.0.x.
-      - 4.0.3 would match only 4.0.3.
-    """
-    if not a.major == b.major:
-        return False
-
-    if a.minor is not None and b.minor is not None:
-        if a.minor != b.minor:
-            return False
-
-    if a.patch is not None and b.patch is not None:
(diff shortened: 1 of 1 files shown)


shivan1b:
@inliniac please have a look at the PR above and let me know if any changes are required. Also, I'm not very sure about the way output looks now. If you have something in mind, please let me know.

Also, @jasonish @inliniac I have questions about certain behaviors.
1. In the code above if someone defines `checks` section in `test.yaml` as well as a `check.sh` then `test.yaml` shall be given preference and `checks.sh` shall be ignored. Is this correct?
2. If `count` is defined in the configuration file then the count of success, failure and skipped shall be reset with commencement of new iteration. Is this correct? I'm referring to this part of the code:  https://github.com/OISF/suricata-verify/blob/master/run.py#L490

victorjulien:
Thanks Shivani, I've merged this through #48.

PR #51, opened 2019-05-02

catenacyber:
Modifies #31 :
- Adds min version 5.0.0


catenacyber:
Needs https://github.com/OISF/libhtp/pull/264 and https://github.com/OISF/suricata/pull/3825 about https://redmine.openinfosecfoundation.org/issues/2881

PR #52, opened 2019-05-03
filestore: add testcases for filecontainer with http and smb

norg:
This adds the requested suricata-verify tests as mentioned in the PR https://github.com/OISF/suricata/pull/3683 for suricata itself.

Both tests will fail due to the fact that besides the filemagic in the rule file also other files (with different filemagic) will be stored. With the changes from the PR 3683 the test for smb will pass but the test for http still fails until this protocol parser gets updated to do a proper filestore as well.


victorjulien:
Thanks Andreas!

PR #56, opened 2019-05-15
tests/tls: add testcases for tls.cert keyword

thus:
This PR contains testcases that @victorjulien wanted for https://github.com/OISF/suricata/pull/3855.


thus:
"tls.cert" is being renamed to "tls.certs", so I'm closing this until the new PR is ready.

PR #58, opened 2019-05-15
tests/tls: add testcases for tls.certs keyword

thus:
This PR contains testcases that @victorjulien wants for OISF/suricata#3859.


tests/tls-certs-alert/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+  min-version: 5.0.0
+  features:
+    - HAVE_LIBJANSSON
+
+args:
+ - -k none
+
+checks:
+  # Check that we have the following events in eve.json
+  # 2 - alert events
```
victorjulien: 2 or 3 like the actual test below?

Might be good to split the alert test into a check for each specific sig.

thus:
Replaced by #60

PR #61, opened 2019-05-17
Add invalid rule tests #317 v2

jmtaylor90:
This PR takes the rules attached to redmine issue #317 and makes tests for them.
There is one invalid rule that is being loaded by suricata master branch, redmine #2982 is tracking that.

This PR has updates from v1 that resolve python parsing issues of escaped json strings.

### DIFF
--- a/tests/test-bad-content-dsize-rule-1/suricata.yaml
+++ b/tests/test-bad-content-dsize-rule-1/suricata.yaml
@@ -1,0 +1,10 @@
+%YAML 1.1
+---
+
+logging:
+  default-log-level: info
+  outputs:
+  - file:
+      enabled: yes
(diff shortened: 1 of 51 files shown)


victorjulien:
Passes on 5.0.0-dev, but fails on 4.1.x. This is because in 4.1.x the config:
```yaml
logging:
  default-log-level: info
  outputs:
  - file:
      enabled: yes
      filename: eve.json
      type: json
```
the ```filename``` variable is not taking ```-l``` or the ```default-log-dir``` variable.

In this case I think the simplest thing is just to set a ```min-version: 5.0.0``` requirement to the tests.

jmtaylor90:
continued in #63

PR #66, opened 2019-05-21
Adds test case four uri double encoding

catenacyber:
Modifies #62 :
- Fixes suricata.yaml for double-decode activation

Cf https://github.com/OISF/suricata/pull/3876


victorjulien:
Merged in #72, thanks!

PR #68, opened 2019-05-23
Tests: add more coverage for invalid rule syntax

jmtaylor90:
This PR adds some more tests that validate logic in the detect engine for various bad rule scenarios.

v2 adds more tests from v1 and fixed a potential run time issue with a couple of the tests.


jmtaylor90:
These tests were validated against suricata built in https://github.com/OISF/suricata/pull/3875

victorjulien:
Merged in #72, thanks Jason!

PR #67, opened 2019-05-21
tests: add testcases for SNMP v2c, v3 (unauth and encrypted)

chifflier:
This PR adds testcases for SNMP (see https://github.com/OISF/suricata/pull/3879)

PR #73, opened 2019-06-06
Snmp add test files 02

chifflier:
Update of #67 to use snake_case for PDU types


victorjulien:
Merged with the following fixes:
1. rebased to master
2. converted pcap-ng to pcap (mingw and some openbsd doesn't support pcap-ng yet)
3. added requirements to make sure suricata 4.1 skips these tests.

PR #71, opened 2019-05-29
tests: dns midstream reversed tests for tcp and udp

jasonish:
Test for:
https://github.com/OISF/suricata/pull/3894

The TCP one currently fails as the DNS response is logged twice.

PR #74, opened 2019-06-11
tests: add testmyids variant with midstream

victorjulien:
@jlucovsky this is the http test to show that midstream currently acts weirdly
```
===> alert-testmyids-midstream: Sub test #2: FAIL : expected 1 matches; got 2 for filter {'count': 1, 'match': {'event_type': 'http', 'dest_port': 80}}
```
2 http records are logged, while there is just one.

Related to #71


victorjulien:
Merged in #125

PR #76, opened 2019-06-18
Add tests for TFTP read and write requests

shivan1b:
Closes redmine ticket #3043.


catenacyber:
Looks good to me

tests/tftp-tx-handling-rrq/suricata.yaml
```diff
@@ -0,0 +1,14 @@
+%YAML 1.1
```
victorjulien: you should be able to leave this out as the default config has everything enabled already AFAICS

tests/tftp-tx-handling-wrq/suricata.yaml
```diff
@@ -0,0 +1,14 @@
+%YAML 1.1
```
victorjulien: same here

shivan1b:
Replaced by #88

PR #77, opened 2019-06-20
Http evader ok v1

catenacyber:
As requested by @victorjulien 

Here are all http evader tests that already pass in master+0.5.x
The ones that do not pass are skipped due to missing feature HTTP_EVADER_OK
This is a separate commit to be reverted later when all tests pass


victorjulien:
Thanks @catenacyber. @jasonish @shivan1b how do you feel about dropping these 501 tests in? Each tests something different, so it's useful. But it also adds quite a bit to the run time and generates quite a bit more lines of output than before.

jasonish:
> Thanks @catenacyber. @jasonish @shivan1b how do you feel about dropping these 501 tests in? Each tests something different, so it's useful. But it also adds quite a bit to the run time and generates quite a bit more lines of output than before.

I'm OK with it, I think there is value in having it.  Has me wondering though if we should think about a way to group these tests under a single directory.  It reports a single OK or failure, unless in verbose mode.  That can be done later though, and won't speed it up anyways.

catenacyber:
> we should think about a way to group these tests under a single directory

Sounds a good idea to me

shivan1b:
I propose we add a `skip-test <folder name>` option. That way we can keep running other unrelated tests fast.
For this particular test, all of this should probably go under `http-evade` folder as Jason suggested. But, `suricata-verify` won't ATM browse through directories looking for tests, it will skip all these tests if we put these in a folder. I think we should modify `suricata-verify` to accept folders inside folders [1 level deep only] for tests and then merge this one. This is huuuge.

catenacyber:
So, should I work on this directory feature as part of this PR ?

victorjulien:
On 03-07-19 07:06, Catena cyber wrote:
> So, should I work on this directory feature as part of this PR ?

No, I think we can merge as-is. Could you open a ticket for it though?

catenacyber:
Redmine seems to list only suricata and suricata-udpate, but not suricata-verify as project
Should I create the ticket anyways ?

victorjulien:
On 03-07-19 07:52, Catena cyber wrote:
> Redmine seems to list only suricata and suricata-udpate, but not
> suricata-verify as project
> Should I create the ticket anyways ?

Yeah in Suricata with the QA target.

victorjulien:
I've merged this, but I didn't properly test. Some of the targets I run my CI on is mingw, and it does not support pcap-ng. Could you convert the pcaps to plain old pcap and submit a new PR?

catenacyber:
Do you want to cherry-pick 11e3e1f ?

Should we add a test for this requirement ?
This is not the first tile this mistake is done, and likely not the last one

victorjulien:
@catenacyber pushed your commit to master, thanks!

PR #79, opened 2019-06-22
Tests for uricontent matching regression

shivan1b:
Closes redmine ticket: https://redmine.openinfosecfoundation.org/issues/3055


victorjulien:
I think the suricata.yamls can be removed. In general, only add them when they are really needed.

shivan1b:
Replaced by #87

PR #80, opened 2019-06-24
tests: add invalid dsize distance usage (#2982)

jmtaylor90:
This is the validation/test case for redmine issue #2982 and passes with the PR https://github.com/OISF/suricata/pull/3969

### DIFF
--- a/tests/test-bad-dsize-distance-rule-1/test.rules
+++ b/tests/test-bad-dsize-distance-rule-1/test.rules
@@ -1,0 +1,1 @@
+alert udp any any -> any any (msg:"TEST SUCCESFULL - dsize/distance INVALID combination "; dsize:10; content:"boom"; content:"loom"; distance:10; sid:6666663; rev:1;)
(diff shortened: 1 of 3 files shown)


tests/test-bad-dsize-distance-rule-1/test.rules
```diff
@@ -0,0 +1 @@
+alert udp any any -> any any (msg:"TEST SUCCESFULL - dsize/distance INVALID combination "; dsize:10; content:"boom"; content:"loom"; distance:10; sid:6666663; rev:1;)
```
victorjulien: any 'distance' value above 2 should fail I think: 4 bytes 'boom', then the match of 'loom' needs to fit in the max 6 bytes that are left. So we can have at max a 2 byte 'distance' before 'loom' can't fit anymore. Did you test this too?

jmtaylor90: I only test this rule. I can add a couple more tests and rework the suricata PR to be better based on your feedback. Thanks for the comments!

jmtaylor90:
continued in https://github.com/OISF/suricata-verify/pull/93

PR #84, opened 2019-06-30
Add tests for #130: content + nocase issue

shivan1b:
Closes redmine ticket #3057.


tests/bug-130/suricata.yaml
```diff
@@ -0,0 +1,14 @@
+%YAML 1.1
```
victorjulien: I don't think it will be needed to provide a yaml in this case. The default yaml should also have the relevant output enabled

shivan1b:
Replaced by #89

PR #93, opened 2019-07-11
tests: add invalid dsize distance usage (#2982)

jmtaylor90:
This is the validation/test case for redmine issue #2982 and passes with the PR https://github.com/OISF/suricata/pull/4023


jmtaylor90:
continued in #104

PR #94, opened 2019-07-11
Dir grouping/v1

shivan1b:
Closes redmine ticket: https://redmine.openinfosecfoundation.org/issues/3069


shivan1b:
Hey @catenacyber !
Could you please have a look at this?

run.py
```diff
@@ -803,8 +803,11 @@ def main():
         if dirpath == tdir:
             continue
 
-        # We only want to go one level deep.
-        dirnames[0:] = []
+        # Check if there are sub-test directories
+        if "test.yaml" in filenames or "check.sh" in filenames:
+            dirnames[0:] = []
```
catenacyber: To me, this is python dark magic (`dirnames` value seems not used) but it was already like this

run.py
```diff
@@ -764,6 +764,8 @@ def main():
                         help="Exit on test failure")
     parser.add_argument("--testdir", action="store",
                         help="Runs tests from custom directory")
+    parser.add_argument("--skip-tests", nargs="?", default=None,
+                        help="Runs tests from custom directory")
```
catenacyber: The long description is copy-pasted 
It should be updated to `Skips tests with this pattern`

run.py
catenacyber: Maybe we can add a comment that `dirnames` value reference gets used by `os.walk`

run.py
shivan1b: Oops. Fixing this. Thanks.

run.py
shivan1b: OK. Thanks.

shivan1b:
Repalced with #115

PR #97, opened 2019-07-14

jlucovsky:


jlucovsky:
Continued in #98

PR #98, opened 2019-07-15

jlucovsky:
Continuation of #97 


jlucovsky:
Continued in #101

PR #101, opened 2019-07-17

jlucovsky:
Continuation of #98


tests/detect-domain-01/suricata.yaml
```diff
@@ -0,0 +1,12 @@
+%YAML 1.1
```
victorjulien: Do the tests not work if this is left out? A yaml should only be added to the tests if the default config is not sufficient.

tests/detect-tld-01/suricata.yaml
```diff
@@ -0,0 +1,12 @@
+%YAML 1.1
```
victorjulien: Same here, is the yaml really needed?

jlucovsky:
Continued in #102

tests/detect-tld-01/suricata.yaml
jlucovsky: No; will remove.

tests/detect-domain-01/suricata.yaml
jlucovsky: No; will remove.

PR #102, opened 2019-07-20

jlucovsky:
Continuation of #101

This PR adds tests for new transformations: tld and domain and tests code added in https://github.com/OISF/suricata/pull/4063


victorjulien:
Looks like the pcap files are pcap-ng. Unfortunately not all our supported platforms support that yet. Can you convert them to regular pcap?

jlucovsky:
Continued in #105

PR #103, opened 2019-07-21
tests: add test cases for FTP logging

jlucovsky:
This PR adds tests for FTP logging. See (3093)[https://github.com/OISF/suricata/pull/3093]


tests/output-eve-ftp/suricata.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.1
```
victorjulien: why is this needed? The default config has ftp enabled.

tests/output-eve-ftp/test.yaml
```diff
@@ -0,0 +1,65 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
```
victorjulien: we will need a check for the version as well, as 4.1.x will fail this test

tests/output-eve-ftp/suricata.yaml
jlucovsky: Will remove.

tests/output-eve-ftp/test.yaml
jlucovsky: Will add a `files` requirement for `src/output-json-ftp.c` which was added for FTP logging.

jlucovsky:
Continued in #106

PR #105, opened 2019-07-22

jlucovsky:
Continuation of #102

This PR adds tests for new transformations: tld and domain and tests code added in https://github.com/OISF/suricata/pull/4063

This update modifies the domain test pcap file type to `pcap` from `pcap-ng`


jlucovsky:
Continued in #111

PR #108, opened 2019-08-01
Geoip test

regit:
This patch adds:
- detection of Geoip support
- basic geoip test

### DIFF
--- a/tests/geoip/suricata.yaml
+++ b/tests/geoip/suricata.yaml
@@ -1,0 +1,13 @@
+%YAML 1.1
+---
+
+geoip-database: /var/lib/GeoIP/GeoLite2-Country.mmdb
+
+outputs:
+  - eve-log:
+      enabled: yes
+      filetype: regular
+      filename: eve.json
+      types:
+        - alert
+
--- a/tests/geoip/test.yaml
+++ b/tests/geoip/test.yaml
@@ -1,0 +1,19 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+    - GeoIP
+
+checks:
+
+  # Check that we only have one alert event type in eve.
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+
+  # Check alert is the good one
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 2
--- a/run.py
+++ b/run.py
@@ -196,9 +196,18 @@ class SuricataConfig:
 
     def load_build_info(self):
         output = subprocess.check_output([suricata_bin, "--build-info"])
+        start_support = False
         for line in output.splitlines():
             if line.decode().startswith("Features:"):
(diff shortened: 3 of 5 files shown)


tests/geoip/test.yaml
```diff
@@ -0,0 +1,19 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+    - GeoIP
```
victorjulien: this looks suspicious. Should it be something like HAVE_GEOIP?

tests/geoip/suricata.yaml
```diff
@@ -0,0 +1,13 @@
+%YAML 1.1
+---
+
+geoip-database: /var/lib/GeoIP/GeoLite2-Country.mmdb
```
victorjulien: I don't think we can assume all OSs will put it here.

victorjulien:
@jasonish can you check the run.py additions/changes?

tests/geoip/test.yaml
regit: We don't have this key. Should I add it in Suricata ?

The code is using an extension of features detection added in 0733c4e.

tests/geoip/suricata.yaml
regit: @jasonish any idea on how to to this cross OSes ?

jasonish: > @jasonish any idea on how to to this cross OSes ?

Unfortunately not. And this is not the location used on CentOS/Fedora, so we can't even settle on a location across the most popular distros.

Maybe a pre-script that hunts it down in various known locations and copies it into a temporary directory, somehow skipping the test if it can't be found? This would require some extra support to suricata-verify I think.

tests/geoip/test.yaml
jasonish: Suricata in git master uses `GeoIP2`.

The changes to run.py look OK.  Its probably a good idea to extract information like this and put it in the feaures for use like this.  But GeoIP support probably does deserve a flag in the `Features:` in the first line of output.

tests/geoip/suricata.yaml
regit: It looks like we can build our own maxmindb easily. Which means we could have a small one embedded in the tests. I'm gonna try to create one.

regit: Got a success on that. One last thing to fix and gonna open a new PR.

tests/geoip/test.yaml
regit: Indeed GeoIP2 change was needed.

jasonish: > Indeed GeoIP2 change was needed.

I'm not sure I like the use of GeoIP2 in the --build-info output.  The feature is still GeoIP to me.  The fact that its using v2 of the database is just an implementation detail that I'm not sure should bleed into the feature description.

regit:
Closing in favor of #108

PR #113, opened 2019-08-21
tests: add test case for file_data depth inspection

norg:
This is a test case which currently doesn't generate an alert although it should. PCAP is from https://redmine.openinfosecfoundation.org/issues/2395 as well as the rule but removed some unnecessary parts from the rule (like classification).


victorjulien:
Merged, thanks Andreas!

PR #129, opened 2019-09-19
Give test count as output not check count

shivan1b:
With the current setup, after running suricata-verify, the output would
give stats about the passed, failed or skipped checks. However, it is
cleaner to get the total stats about tests only.

The priority order for the output is:
FAILED
SKIPPED
PASSED

i.e. if a check or sub-test has failed, the entire test will be counted
as failed in the final output. Similarly, if a check has been skipped,
the entire test will be counted as skipped in the final output.

Closes redmine ticket: https://redmine.openinfosecfoundation.org/issues/3172

### DIFF
--- a/run.py
+++ b/run.py
@@ -606,6 +606,10 @@ class TestRunner:
         count = StatsCheck(check, self.output).run()
         return count
 
+    def reset_count(self, dictionary):
+        for k in dictionary.keys():
+            dictionary[k] = 0
+
     def check(self):
         pdir = os.getcwd()
         os.chdir(self.output)
@@ -621,6 +625,7 @@ class TestRunner:
                 for check_count, check in enumerate(self.config["checks"]):
                     for key in check:
                         if key in ["filter", "shell", "stats"]:
+                            self.reset_count(count)
                             func = getattr(self, "perform_{}_checks".format(key))
                             count = func(check=check[key], count=count,
                                     test_num=check_count + 1)
@@ -847,9 +852,12 @@ def main():
             cwd, dirpath, outdir, suricata_config, args.verbose)
         try:
             results = test_runner.run()
-            passed += results["success"]
-            failed += results["failure"]
-            skipped += results["skipped"]
+            if results["failure"]:
+                failed += 1
+            elif results["skipped"]:
+                skipped += 1
+            elif results["success"]:
+                passed += 1
         except UnsatisfiedRequirementError as ue:
             print("SKIPPED: {}".format(ue))
             skipped += 1


run.py
```diff
@@ -847,9 +852,12 @@ def main():
             cwd, dirpath, outdir, suricata_config, args.verbose)
         try:
             results = test_runner.run()
-            passed += results["success"]
-            failed += results["failure"]
-            skipped += results["skipped"]
+            if results["failure"]:
```
catenacyber: `results["failure"] > 0` would be better, wouldn't it ?

run.py
```diff
@@ -621,6 +625,7 @@ def check(self):
                 for check_count, check in enumerate(self.config["checks"]):
                     for key in check:
                         if key in ["filter", "shell", "stats"]:
+                            self.reset_count(count)
```
catenacyber: What has this to do with the goal of this PR, ie Give test count as output rather than check count

catenacyber:
Thanks Shivani, it looks good.
I just have some code style questions...

run.py
shivan1b: Since there is no subtraction operation being performed on results and they are initialized with 0, it is guranteed to be positive.

catenacyber: Yes, your code works.
This is a suggestion of code style.
The goal is to ensure that results will still be correct even if the above assumptions (like there is no subtraction) no longer are.
And it looks more readable/meaningful to me this way.

shivan1b: Oh OK. Sorry for the misunderstanding. Sure, will do that.

run.py
shivan1b: Let's say there's a test `SampleTest` with 4 checks `check1`, `check2`, `check3` and `check4`.
Previously (before the checks count was introduced), the structure of the code was such that if `check2` failed, the code would return and the entire test will show up as failed without running the `check3` and `check4`.
However, while check count was introduced, the actual request was to be able to run all the checks and point out exactly the check which failed so in that that if `check2` failed, `check3` and `check4` would still run and the output would clearly point that `check2` failed and not all the other checks. In order to implement this, every check's state was to be recorded i.e. whether it `failed`, `passed` or was `skipped`. Alongwith this, I was storing the count of states of the checks to be able to give a final result (which was check count). Now that state is reset (the line above) with every test as we're only considering the count of tests in the final output.
The state could be made boolean but that would also have to be reset on ever test.

catenacyber: I am not sure I understand all of this code (maybe I need more coffee)
So, it may deserve some more comments.

Why is `self.reset_count(count)` called on every iteration of the loop `for key in check:` ?
So that count has only one key whose value is 1 (and not more) ?

Maybe my confusion comes from the fact that : ok we should have he final stats about tests... but I would also expect to have detailed stats about checks, especially when running only one test...

shivan1b: @catenacyber that was a bug. Thanks for finding this out. You work great even without coffee. :)

PR #130, opened 2019-09-23
tests: file_data depth inspection should keep working with other rules

gsomlo:
see https://redmine.openinfosecfoundation.org/issues/3190


jasonish:
Fails on 4.1.

PR #132, opened 2019-09-26
filemd5: test md5 rule triggers without filestore keyword - v1

jasonish:
Redmine issue:
https://redmine.openinfosecfoundation.org/issues/2490

Requires https://github.com/OISF/suricata/pull/4234 to pass.


jasonish:
Merged. Thanks!

PR #134, opened 2019-10-01
run: handle moved classificaton/reference config

jasonish:
First look for these configuration files in ./etc, then in the
top directory to handle the change of location in Suricata 5.0,
as well as work with older versions.


jasonish:
Required for
https://github.com/OISF/suricata/pull/4258

victorjulien:
Merged after rebasing.

PR #135, opened 2019-10-03
test: datasets state isnotset test

jasonish:
A dataset state test to test for the first occurrence
of something being seen.

### DIFF
--- a/tests/datasets-state-isnotset/test.yaml
+++ b/tests/datasets-state-isnotset/test.yaml
@@ -1,0 +1,13 @@
+requires:
+  min-version: 5.0.0
+
+pcap: ../datasets-05-state/input.pcap
+
+args:
+  - --data-dir ./output
+
+checks:
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 1 
--- a/tests/datasets-state-isnotset/README.md
+++ b/tests/datasets-state-isnotset/README.md
@@ -1,0 +1,3 @@
+Test dataset isnotseen for state. The idea is to check the first occurrence of
+something happening. Then once recorded in the state, isnotset should not
+match.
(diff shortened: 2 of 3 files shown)


tests/datasets-state-isnotset/test.yaml
```diff
@@ -0,0 +1,13 @@
+requires:
+  min-version: 5.0.0
+
+pcap: ../datasets-05-state/input.pcap
+
+args:
+  - --data-dir ./output
```
victorjulien: This is not portable
```
[15991] 5/10/2019 -- 08:24:13 - (suricata.c:1876) <Error> (ParseCommandLine) -- [ERRCODE: SC_ERR_LOGDIR_CMDLINE(117)] - The data directory "./output" supplied at the commandline (-d ./output) doesn't exist. Shutting down the engine.
```

victorjulien:
Merged after some fixes for the output dir and missing requirements.

tests/datasets-state-isnotset/test.yaml
jasonish: Ah, I see.  Maybe we should consider supporting environment vars in the test.yaml, should simplify this case.

victorjulien:
On 05-10-19 22:02, Jason Ish wrote:
> *@jasonish* commented on this pull request.
> 
> ------------------------------------------------------------------------
> 
> In tests/datasets-state-isnotset/test.yaml
> <https://github.com/OISF/suricata-verify/pull/135#discussion_r331760090>:
> 
>> @@ -0,0 +1,13 @@
> +requires:
> +  min-version: 5.0.0
> +
> +pcap: ../datasets-05-state/input.pcap
> +
> +args:
> +  - --data-dir ./output
> 
> Ah, I see. Maybe we should consider supporting environment vars in the
> test.yaml, should simplify this case.

Yes this would be very useful. There are quite a few tests that could be
written without an explicit 'command' then.

PR #137, opened 2019-10-10
tests: test that triggers a rule with established though 3whs missing

norg:
This creates a test for issue https://redmine.openinfosecfoundation.org/issues/2639 where the 3whs is missing but the rule with established keyword still triggers.


jasonish:
This test current fails. Does it depend on an issue being fixed?

norg:
Yes as described in the issue you **don't** want to have the alert

catenacyber:
There is a typo in the directory name : `estabilshed`

victorjulien:
Merged in #291, thanks!

PR #138, opened 2019-10-13
base64: add basic tests

regit:
Implement some tests for base64 operations.


tests/base64/test.yaml
```diff
@@ -0,0 +1,31 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+  min-version: 5.0.0
```
victorjulien: is there a specific reason not to run these on 4.1?

jasonish:
Merged with #151. Includes fixup for 4.1.

PR #140, opened 2019-10-14
test: add byte_math test

travisbgreen:


victorjulien:
I don't think we're supposed to redistribute TALOS rules. Could you craft your own rules for a test case?

PR #143, opened 2019-10-24
Adds test for http async parsing

catenacyber:
Redmine ticket is https://redmine.openinfosecfoundation.org/issues/2312

Follows #141 :
- added `--set stream.async-oneside=true`
- merged together two filters for http-async-srv

These tests pass now


catenacyber:
By the way, Should not I have got a warning for trying unknown option `stream.async` instead of `stream.async-oneside` ?

jasonish:
> By the way, Should not I have got a warning for trying unknown option `stream.async` instead of `stream.async-oneside` ?

No really. Most of the config works by getting a specific value, and it will get it set, or not. It doesn't enumerate all values looking for ones it knows and doesn't.

jasonish:
Merged. Thanks!

PR #145, opened 2019-10-30
option: adds debugfailed option

catenacyber:
Prints contents of files in output directory of failed tests
Intent is to debug flasky behaviors

cf https://redmine.openinfosecfoundation.org/issues/3096

We should then update .travis.yml in suricata repository

### DIFF
--- a/run.py
+++ b/run.py
@@ -797,6 +797,8 @@ def main():
                         help="Outputs to custom directory")
     parser.add_argument("--valgrind", dest="valgrind", action="store_true",
                         help="Run tests in with valgrind")
+    parser.add_argument("--debugfailed", dest="debugfailed", action="store_true",
+                        help="Prints debug output for failed tests")
     parser.add_argument("patterns", nargs="*", default=[])
     args = parser.parse_args()
 
@@ -855,6 +857,7 @@ def main():
 
     # Sort alphabetically.
     tests.sort()
+    failedLogs = []
 
     for dirpath in tests:
         name = os.path.basename(dirpath)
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -883,6 +888,18 @@ def main():
...
+            print("")
+            print("Failed tests debug output:")
+        for dirpath in failedLogs:
+            print("- Test %s:" % os.path.basename(dirpath))
+            for r, d, f in os.walk(dirpath+"/output"):
+                for fname in f:
+                    print("  - %s" % fname)
+                    fcontents = open(dirpath+"/output/"+fname, "r")
+                    print(fcontents.read())
+
```
inashivb: So, just to be sure, we want to dump the information about why a test failed after running all the tests and not right when they fail?

run.py
```diff
@@ -883,6 +888,18 @@ def main():
...
+    if args.debugfailed:
+        if len(failedLogs) > 0:
+            print("")
+            print("Failed tests debug output:")
+        for dirpath in failedLogs:
+            print("- Test %s:" % os.path.basename(dirpath))
+            for r, d, f in os.walk(dirpath+"/output"):
+                for fname in f:
+                    print("  - %s" % fname)
+                    fcontents = open(dirpath+"/output/"+fname, "r")
```
inashivb: This file pointer would have to be closed. Maybe a context manager with a [`with` statement](https://www.python.org/dev/peps/pep-0343/) would be better.
```
with open(filename) as fp:
    # do something with file
# unindent shall close the file now
```

inashivb: nit: spaces around operators

run.py
catenacyber: Hmm... we could do it right away
This is a design choice about how a human would better read it

run.py
catenacyber: ok for both

PR #146, opened 2019-11-06

catenacyber:


catenacyber:
Test should pass only when https://github.com/OISF/suricata/pull/4347 gets merged

PR #147, opened 2019-11-06
Test modifications for source code reorg - v1

jasonish:


tests/output-eve-anomaly-01/test.yaml
```diff
@@ -1,9 +1,7 @@
 requires:
   features:
     - HAVE_LIBJANSSON
-
-  files:
-    - src/output-json-anomaly.c
+  min-version: 5.0
```
jlucovsky: Should `min-version` be `major.minor` or `major.minor.patch` with `5.0` (`5.0.0`)? Both are used; are `5.0` and `5.0.0` equivalent at this stage?

inashivb: Seems like it won't matter after https://github.com/OISF/suricata-verify/pull/158 is merged. We can be flexible about minor and patch if only we are OK with them being set to 0.

tests/filestore-filecontainer-http/test.yaml
```diff
@@ -2,8 +2,7 @@ requires:
   features:
     - HAVE_NSS
     - MAGIC
-  files:
-    - src/output-filestore.c
```
inashivb: @jasonish have you removed all the files options because the features are guaranteed to be in the specified minimum versions? If yes, should `files` option be removed altogether since all such cases can be covered by `min-version` or `version`. If no, please let me know what's the thought behind this.

victorjulien: I have a use for this for long running dev branches that remain out of tree for a long time.

jasonish: Or even features that are added, but we want to have tests for without having had a version bump in the master branch yet.

jasonish: > have you removed all the files options because the features are guaranteed to be in the specified minimum versions?

Yes. Using 4.1 as the lowest supported version now.

jasonish:
> Should we close this @jasonish ?

Yup. Thanks.

victorjulien:
I'd still like to do the reorg btw.

PR #150, opened 2019-11-13
eve/dns: test eve/dns filtering - v1

jasonish:
To confirm ticket:
https://redmine.openinfosecfoundation.org/issues/3231


victorjulien:
Looks good. Didn't know about the filename use in filters. That is neat.

PR #153, opened 2019-11-19
http-headers-alert: test http headers logging in alert

regit:
Add test for HTTP headers logging in alert.


regit:
Dependent of https://github.com/OISF/suricata/pull/4372

catenacyber:
Should be merged in https://github.com/OISF/suricata-verify/pull/290, right ? as they both depend upon the same suricata PR

PR #157, opened 2019-11-20
tests: fix sip tests

glongo:
The correct numbers of alerts logged is not correct and OISF/suricata#4330
fixes this issue, so this commit fixes tests that are broken.

I removed the test added previously, because it's a duplicate of sip-method test.

Previous PR: #156


victorjulien:
I've merged this. It seems to be fixed with https://github.com/OISF/suricata/pull/4393 as well.

Thanks Giuseppe!

PR #158, opened 2019-11-20
run: use 0 for undefined versioned components

jasonish:
Fix the version parser to use 0 for undefined version
components instead of None to support looser version
specifications:

5 => 5.0.0
5.1 => 5.1.0
5.1.1 => 5.1.1

### DIFF
--- a/run.py
+++ b/run.py
@@ -93,20 +93,9 @@ SuricataVersion = namedtuple(
 def parse_suricata_version(buf):
     m = re.search("(\d+)\.?(\d+)?\.?(\d+)?.*", str(buf).strip())
     if m:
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -96,17 +96,17 @@ def parse_suricata_version(buf):
         if m.group(1) is not None:
             major = int(m.group(1))
         else:
-            major = None
+            major = 0
```
inashivb: Minor comment about style: We could probably write all these blocks as `major = int(m.group(1)) if m.group(1) else 0`.

jasonish: That does clean it up a little. Thanks.

PR #159, opened 2019-11-20
test: dhcp request flood test

jasonish:
Tests that a DHCP request flood doesn't take an excessive amount of time.

   Skipped by default.

   Related ticket #3345: https://redmine.openinfosecfoundation.org/issues/3345


run.py
```diff
@@ -392,6 +392,10 @@ def setup(self):
     def check_skip(self):
         if not "skip" in self.config:
             return
+        if isinstance(self.config["skip"], bool):
+            if self.config["skip"]:
```
inashivb: Nit: Maybe this if check can be skipped since we're checking in the line above the instance type for this var then it must exist.

run.py
```diff
@@ -392,6 +392,10 @@ def setup(self):
...
+                raise UnsatisfiedRequirementError("skipped by default")
+            return
```
inashivb: nit: This return statement can be skipped too since we're using a `raise` in the statement above this.

run.py
```diff
@@ -509,8 +510,9 @@ def run(self):
         sys.stdout.write("===> %s: " % os.path.basename(self.directory))
         sys.stdout.flush()
 
-        self.check_requires()
-        self.check_skip()
+        if not self.force:
+            self.check_requires()
+            self.check_skip()
```
inashivb: Just to be clear if there is `force` argument then we're not supposed to check for `requires` or `skip`?

run.py
jasonish: Without a return here the next line will try to iterate on a boolean.

run.py
jasonish: Yeah, I think it should. Sometimes you just need to force a test to run without modifying it.

run.py
jasonish: I'm not so sure.. With this change we have to support 3 datatypes for skip.  It could be None, a boolean or a list.

inashivb: Yes, you're right. Sorry about this.

run.py
inashivb: Correct. I did not test the False case at all. Sorry about this as well.

jasonish:
Merged.

PR #164, opened 2019-12-06
Ipv6 evasions

norg:
This is a first batch of suricata-verify tests for the ipv6 evasion issue https://redmine.openinfosecfoundation.org/issues/3323 which is currently WIP.

Notes:

- rules can be improved, especially if there is a way to detect the details of the evasion/attacks
- rules that trigger don't focus on the actual evasion/attack
- some are skipped as there is nothing to match on currently


tests/ipv6-evasion/ipv6-dos-with-ext-headers-1/test.yaml
```diff
@@ -0,0 +1,10 @@
+requires:
+    features:
+        - HAVE_LIBJANSSON
+
+checks:
+    - filter:
+        count: 34114
```
victorjulien: is there a reason for testing such a large number? I assume these are not all unique packets?

norg: well that's one of the things to discuss, the pcap tries to DoS by using those ext headers in a way that can increase decoding load. So one might think about a more advanced rule or decoder events.

PR #167, opened 2019-12-16
Add createst script

inashivb:
createst is a script to produce a test directory with test.yaml as per
the PCAP and configuration provided.

This currently implements the functionality of creating the "checks" block
in `test.yaml` from a given `eve.json`. You can add other configuration
in the file thus created.

Usage
=====
```
usage: createst.py [-h] [--output-path <output-path>] [--eventtype-only]
                   [--allow-events [ALLOW_EVENTS]]
                   <test-name> <pcap-file>

Create tests with a given PCAP. Execute the script from a valid Suricata
source directory.

positional arguments:
  <test-name>           Name of the test folder
  <pcap-file>           Path to the PCAP file

optional arguments:
  -h, --help            show this help message and exit
  --output-path <output-path>
                        Path to the folder where generated test.yaml should be
                        put
  --eventtype-only      Create filter blocks based on event types only
  --allow-events [ALLOW_EVENTS]
                        Create filter blocks for the specified events
```

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -1,0 +1,405 @@
+#! /bin/python
+#
+# Copyright (C) 2019 Open Information Security Foundation
+#
+# You can copy, redistribute or modify this Program under the terms of
+# the GNU General Public License version 2 as published by the Free
+# Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# version 2 along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA
+# 02110-1301, USA.
+
+import argparse
+import glob
+import json
+import logging
+import os
+import subprocess
+import sys
+from collections import defaultdict
+from shutil import copyfile
+
+import yaml
+from yaml.representer import Representer
+
+# Get a logger instance
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.DEBUG)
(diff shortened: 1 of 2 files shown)


jasonish:
This seems to be working very well now, and I'm finding it useful as is while I add more detail to the DNS tests.

Some ideas for future enhancements, as I'm adding these to the generated tests:
- Removing flow. Maybe I should leave them in, but am a little worried about stability of the flow end time. Thoughts on disabling flow by default? Or suppressing the flow timestamp from the output? @victorjulien ?
- I'm adding a filter that just counts the number of dns records to make sure more than expected are not being generated, for example:
```
- filter:
    count: 4
    match:
      event_type: dns
```
If these seems like reasonable features, I could create tickets for them, but I don't think they should hold up getting this in.

victorjulien:
I'd like the flow tests to be created, but I agree we can skip the timestamps (at least by default).

createst.py
```diff
@@ -0,0 +1,407 @@
...
+yaml.add_representer(defaultdict, Representer.represent_dict)
+
+WIN32 = sys.platform == "win32"
+suricata_bin = "src\suricata.exe" if WIN32 else "./src/suricata"
+suricata_yaml = "suricata.yaml" if WIN32 else "./suricata.yaml"
+CUR_DIR = os.path.dirname(os.path.realpath(__file__))
+DEFAULT_TEST_DIR = os.path.join(CUR_DIR, "tests")
+
+# Fields to exclude from the filter block
+skip_fields = ["timestamp", "flow_id", "last_reload"]
```
inashivb, line 46: @jasonish @victorjulien these are the skipped fields.
I'll check again just to be sure.

jasonish: We'll also want to skip "flow.start" and "flow.end".

inashivb: Done.

inashivb:
Replaced with #227

PR #171, opened 2020-01-19
Adds test for http range file

catenacyber:
Follows #163 
Yaml may certainly be improved to add more tests
But this is a valid basic test for the functionality


catenacyber:
Closing for now, as suricata PR needs rework

PR #175, opened 2020-02-12
Adds test for SMB EICAR file

catenacyber:
Regular write, without evasions, SMBv1

Changes #174 with checksums disabled, correct EICAR rule, and matching on count=1


jasonish:
Merged. Thanks!

PR #176, opened 2020-02-13
Adds draft test for ssh banner eol split

catenacyber:
Draft not to be merged yet : test.yaml needs to be improved to find the right condition for testing

Should need Suricata PR https://github.com/OISF/suricata/pull/4546 to work

The pcap file triggers a bug in SSH parser, having an incorrect offset leading to a much too big `pkt_len` here : https://github.com/OISF/suricata/blob/master/src/app-layer-ssh.c#L173

This results in `header->flags |= SSH_FLAG_PARSER_DONE;` never being applied for the TOSERVER side.
This results in `AppLayerParserStateSetFlag(pstate, APP_LAYER_PARSER_NO_INSPECTION);` not being applied and `SSHGetAlstateProgress` not returning the right value.

Can we test for these with suricata-verify ? How do I check this in test.yaml ? Or am I bound to do a C unit test for this like https://github.com/OISF/suricata/blob/master/src/app-layer-ssh.c#L2079 ?


catenacyber:
Suricata PR got merged 

Any thoughts on this ?
Otherwise, I will close it

PR #178, opened 2020-02-18
Document "shell" checks - v1

jasonish:


README.md
```diff
@@ -118,4 +118,15 @@ checks:
 		
 		# Check that a field does not exist:
 		not-has-key: flow
+
+  - shell:
+      # A simple shell check. If command exits non 0 the check
```
victorjulien: suggest: If the command exists with a non-0 exit code, the check...

jlucovsky: s/exists/exits

jasonish:
Merged with suggested fixups.

PR #179, opened 2020-02-19
pcre: PCRE modifier tests

jlucovsky:
This commit ensures that the PCRE modifiers are properly handled and that results from shell executions are properly checked for errors.

This PR is a companion of [#4570](https://github.com/OISF/suricata/pull/4570)


victorjulien:
File list contains some (vim?) temp .swp files. Think these are not supposed to be checked in.

jlucovsky:
Continued in #180

PR #184, opened 2020-02-24
Removed unified2 test.

jasonish:
Related Suricata pull-request:
https://github.com/OISF/suricata/pull/4596


victorjulien:
Shouldn't we keep the test until u2 is removed from all supported Suricata versions?

jasonish:
> Shouldn't we keep the test until u2 is removed from all supported Suricata versions?

Yeah, probably.  Its already designed to be skipped if not supported (must have been some forward thinking there!)

PR #185, opened 2020-02-25
add basic RFB checks

satta:
This PR adds simple checks for the upcoming RFB parser in Suricata. It verifies that RFB traffic for protocol versions 3.3, 3.7 and 3.8 can be parsed, keywords trigger alerts and the correct metadata are logged.
The pcaps are taken from https://github.com/kholia/my-pcaps/tree/master/VNC.


satta:
New PR: #198

PR #183, opened 2020-02-24
tests: pcrexform tests

jlucovsky:
Test cases for the `pcrexform` transform.

Companion [Suricata PR](https://github.com/OISF/suricata/pull/4595)

PR #186, opened 2020-02-27
tests: pcrexform tests

jlucovsky:
Continuation of #183 

Test cases for the `pcrexform` transform.

Companion [Suricata PR](https://github.com/OISF/suricata/pull/4595)

Updates
- Additional test case to verify mandatory quotes requirement.


victorjulien:
Could you also add a test where a single rule uses the xform twice, on different buffers? This would test the non-mpm path as well.

e.g. `alert http ... http.request_line; pcrexform... ; content... ; http.uri; pcrexform:...; content...`

jlucovsky:
Continued in #196

PR #172, opened 2020-01-26
tests: add ipv6 evasion tests

norg:
This is a complete but WIP batch of suricata-verify tests for the ipv6 evasion issue https://redmine.openinfosecfoundation.org/issues/3323.

The ones without rules are skipped.
Some of the ones that trigger, trigger just the IPv6 checksum rule which is not detecting the actual attack.

PR #187, opened 2020-02-28
Adds test for IPv6 evasion atomic fragment

catenacyber:
This is one of the pcaps from #172 with a modified rule to make it pass (with suricata PR) cc @norg


norg:
LGTM

victorjulien:
Merged, thanks!

PR #190, opened 2020-03-03
Adds test for SMB EICAR file nbss-more-ffsmb

BasTNCY:
Regular write, two NetBIOS messages in one segment, SMBv1


BasTNCY:
cc @catenacyber

catenacyber:
> two NetBIOS messages in one segment

Do you mean two NetBIOS messages in one TCP packet ?

PR #191, opened 2020-03-03
Adds test for SMB EICAR file segmentation random v1

BasTNCY:
Regular write, segmentation in random size, SMBv1


BasTNCY:
CC @catenacyber

catenacyber:
This should only have one commit and not the one from #190

PR #195, opened 2020-03-05
tests: ERSPAN Type I packet decode if config

jlucovsky:
This commit adds tests for ERSPAN Type I decoding based on configuration
settings.

Requires [Suricata PR #4627](https://github.com/OISF/suricata/pull/4627)


victorjulien:
This fails for master because there we decided not to make the decoder optional. I don't think we can specify a `max-version` @jasonish @inashivb ?

inashivb:
> This fails for master because there we decided not to make the decoder optional. I don't think we can specify a `max-version` @jasonish @inashivb ?

No, we cannot. Do we need that feature?
One way with the current options in s-v could be writing multiple tests with `version:XXX`? Move all the tests in a folder `decode-erspan`. Would that help?

jasonish:
> No, we cannot. Do we need that feature?

I think a max-version will have its uses as we deprecate things as well.

tests/decode-erspan-typeI-03/test.yaml
```diff
@@ -0,0 +1,17 @@
...
+
+checks:
+
+    - filter:
+        count: 0
+        match:
+            event_type: flow
+
+    - stats:
+        decoder.erspan: 0
```
catenacyber, line 17: This test does not pass even if the suricata PR was merged.
`Sub test #2: FAIL : stats.decoder.erspan: expected 0; got 42`
What is wrong ? The test or suricata ?

jlucovsky:
Continued in #286

PR #196, opened 2020-03-07
tests: pcrexform tests

jlucovsky:
This PR adds tests for the `pcrexform` including tests for
- basic functionality: Simple PCRE
- extended functionality: Multiple PCREs
- Negative: PCRE that do not match anything
- Negative: Missing option values

[Suricata PR #4651](https://github.com/OISF/suricata/pull/4651)


tests/detect-pcrexform-01/test.yaml
```diff
@@ -0,0 +1,8 @@
+requires:
+
```
catenacyber, line 2: Should we set a min-version of 6.0 ?

jlucovsky: Thanks for pointing this out. I'll update the `requires` section

jlucovsky:
Continued in #285

PR #200, opened 2020-03-20
framework: warns on different type between yaml and json

catenacyber:
This is a draft for discussion

I lost some time debugging this issue using `2.0` instead of `"2.0"` in yaml for `ssh.client.proto_version`

### DIFF
--- a/run.py
+++ b/run.py
@@ -338,10 +338,10 @@ class FilterCheck:
                     return False
             else:
                 val = find_value(key, event)
-                if type(val) != type(expected):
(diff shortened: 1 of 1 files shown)


jasonish:
I have no issues with doing this.

jasonish:
But I get lots of stuff like this:
```
Different type <type 'unicode'> alert <type 'str'> alert
Different type <type 'NoneType'> None <type 'int'> 1
```

catenacyber:
Hmm... Then what do you think about latest commit ?

jasonish:
Merged. Thanks!

PR #206, opened 2020-03-27
tests: add http file blocking on sha256 test

victorjulien:
Test based on https://forum.suricata.io/t/suricata-5-0-1-in-ips-mode/94/14

Note that it fails on 4.1 (for this reason I've set the min-version to 5.0.0 for now). @inashivb can you create a redmine ticket for this, and try to see what fix we are missing in 4.1?


victorjulien:
Replaced by #207

PR #207, opened 2020-04-01
tests: add http file blocking on sha256 tests

victorjulien:
Test based on https://forum.suricata.io/t/suricata-5-0-1-in-ips-mode/94

Note that it fails on 4.1 (for this reason I've set the min-version to 5.0.0 for now). @inashivb can you create a redmine ticket for this, and try to see what fix we are missing in 4.1?


victorjulien:
I've done some investigating and it seems it never worked, at least not after 3.2 (didn't go back further). Identified the missing commits.

victorjulien:
Merged with small improvements in #213

PR #208, opened 2020-04-01
tests: mark 2 smb file tests as requiring rust

jasonish:
- filestore-container-smb
- smb-eicar-file


victorjulien:
Thanks Jason.

PR #209, opened 2020-04-03
framework: use python 3 by default - v1

jasonish:
Also:
- Move pcapng check to GitHub CI
- Use checkout@v2


jasonish:
> What is the status on this @jasonish ?

Thanks for the ping. Merged what was relevant. The Python 3 update had already been done.

PR #212, opened 2020-04-07
Adds test for SMB keyword smb_named_pipe with ascii

BasTNCY:
Test for SMB "smb_named_pipe" keyword in the rules


BasTNCY:
cc @catenacyber

catenacyber:
Thanks @BasTNCY 

Can we have two tests ?
A new one named smb-named-pipe-unicode and this one renamed smb-named-pipe-ascii ?

PR #215, opened 2020-04-14
Adds 2 tests for SMB keyword smb_named_pipe with ascii and unicode

BasTNCY:
2 Tests for SMB "smb_named_pipe" keyword in the rules


catenacyber:
Looks good to me, thanks @BasTNCY

victorjulien:
Merged, thanks! For future SMB tests please add the `RUST` feature, as Suricata 4.1 only supports SMB when Rust is built-in.

PR #216, opened 2020-04-16
tests/dns: add test for logging SOA fields

simdugas:
Test config and pcap were copied from an existing test dns-udp-nxdomain-soa.

See [pull request](https://github.com/OISF/suricata/pull/4955)


simdugas:
Rebased master and opened #309.

PR #219, opened 2020-04-21
test/ftp: Verify FTP memory usage

jlucovsky:
This commit adds a test case for verifying that FTP memory use and
accounting are correct.

Companion [Suricata PR](https://github.com/OISF/suricata/pull/4844)


victorjulien:
Merged with a small adjustment: I removed the dependency on the eve.json output so that the 4.1.x branch is also tested.

Thanks Jeff!

PR #223, opened 2020-04-26
tests: check for tls events to trigger once per flow

norg:
Test for issue https://redmine.openinfosecfoundation.org/issues/1983


jasonish:
Appears to pass now: https://github.com/jasonish/suricata-verify/actions/runs/4963447610

jasonish:
Reabsed here to show that this now passes.

catenacyber:
Which certificate is supposed to have an invalid length in this big pcap with many flows ?

catenacyber:
I would close this PR as it does not get any anomalies (instead of twice as much as wanted)

catenacyber:
Closing this PR, work to be continued in the redmine ticket...

PR #224, opened 2020-04-26
tests: add regression tests for smb close response messages

norg:
tests for issue https://redmine.openinfosecfoundation.org/issues/2656


tests/smb-nt-create-andx-request-02/test.yaml
```diff
@@ -0,0 +1,17 @@
+# skip as it fails
+skip: true
```
victorjulien, line 2: I see 4 alerts on this instead of 3. Is that the failure?

norg: that's a difference between 4.1.x and 5.0.x although I had 3 on 5.0.1 as well for example. Will try to narrow it down to a commit/version.

norg: So the first narrow down shows that 4.1.7 and 5.0.1 behave the same. With 5.0.2 the difference appears with the first test, while the others are still the same. What's a bit strange is this https://github.com/norg/suricata-verify/runs/620367010?check_suite_focus=true also shows the same issue but with the major versions the other way round.

So this made me testing more and it seems to be another issue. It can happen with each of those version (4.1.7, 5.0.1, 5.0.2). I did run 100 tests with `run.py smb-nt-create-andx-request-01` on each of those and it happened more often on 5.0.2 and less often on 4.1.7.

I did test the other ones in that commit as well but each had the same amount of events in those runs. So it's only a strange behavior in `smb-nt-create-andx-request-01`.

catenacyber:
smb-powershell_activity-03 is simply missing `--set stream.midstream=true` to pass (no TCP handshake)

victorjulien:
@catenacyber why is this closed? Is there no use for the test when it is fixed up?

catenacyber:
> Is there no use for the test when it is fixed up?

This is just about stream.midstream=true, not about some fancy SMB stuff...
I do not think it is worth having 3 times the same pcap.
But I can rebase this with the stream.midstream=true fix if you wish

PR #230, opened 2020-05-07
tests: Add test for file_data w/transform

jlucovsky:
Companion [Suricata PR](https://github.com/OISF/suricata/pull/4927)

Add a test for to apply transforms to `file_data` sticky buffers.


victorjulien:
Can you expand this test to have multiple combinations of file_data and transforms? I fear we may have an issue in that case.

So rule file with rules with:
- just file_data
- file_data + strip_whitespace
- file_data + compress_whitespace
- file_data + urldecode

jlucovsky:
> Can you expand this test to have multiple combinations of file_data and transforms? I fear we may have an issue in that case.
> 
> So rule file with rules with:
> 
> * just file_data
> * file_data + strip_whitespace
> * file_data + compress_whitespace
> * file_data + urldecode

Is the issue that the other transforms should validate?

PR #231, opened 2020-05-15
framework: new requires: lt-version (less than version)

jasonish:
Implement a less than version requirement. Can be used prevent a test running on versions 6 and newer like required in https://github.com/OISF/suricata-verify/pull/195.

Other changes:
- Fix the self unit tests to run.
- Add a command line option to run the self unit tests.
- Run the self unit tests in GitHub CI on push
- Set a lt-version for filestore v1 tests.


inashivb:
Seems like this can be closed?

jasonish:
Merged.

PR #239, opened 2020-06-02
dhcp test: fixup for unicode in eve output

jasonish:
With JsonBuilder, NUL bytes in a string are now encoded as
unicode (\u0000) instead of our previous convention of \\x00.


victorjulien:
Breaks 4.1 and 5.0.

jasonish:
Should be fixed here: https://github.com/OISF/suricata-verify/pull/241

PR #243, opened 2020-06-03
xor: test match on xor decoded data

simdugas:
Adds a test case to match on xor decoded data using the xor and xor_data
keywords.

Link to PR: [#5015](https://github.com/OISF/suricata/pull/5015)


catenacyber:
Closing, waiting for the transform implementation

PR #247, opened 2020-06-09
tests/hassh Add test cases for hassh

MalakhatkoVadym:
This commit adds test cases for the ssh.hassh, ssh.hassh.string, ssh.hassh.server, ssh.hassh.server.string keywords.


tests/ssh-hassh/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+  min-version: 5.0.0
+features:
+    - RUST
+
+args:
+ - -k none
+
+checks:
+  # Check that we have the following events in eve.json
+  # 2 - alert events
```
catenacyber, line 11: The comment says 2 and the code says 4...

tests/ssh-hassh/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+  # 1 - tls event
+  - filter:
+      count: 4
```
catenacyber, line 14: Should we check that there is one alert per signature (ie per sid) ?

tests/ssh-hassh/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+      match:
+        event_type: alert
+  - filter:
+      count: 1
+      match:
+        event_type: ssh
```
catenacyber, line 20: Should we add a deeper filter to check the value of ssh.hassh ?

tests/ssh-hassh/test.yaml
MalakhatkoVadym: Yes, will add now.

tests/ssh-hassh/test.yaml
MalakhatkoVadym: it's checked already in rules (rule will match only for specific hassh values).

tests/ssh-hassh/test.yaml
MalakhatkoVadym: Sorry, my bad.

tests/ssh-hassh/test.yaml
catenacyber: Ok, so should we remove the whole filter of one `event_type == ssh` ?
Or is there more to it ?

MalakhatkoVadym: yes, I think, it can be removed.

MalakhatkoVadym:
https://github.com/OISF/suricata-verify/pull/248

PR #250, opened 2020-06-14
tests/ftp-data Test case for event type ftp_data

jlucovsky:
This commit adds a test case to insure that `ftp_data` events are being
generated.

Redmine issue [3770]((https://redmine.openinfosecfoundation.org/issues/3770)


victorjulien:
Merged as part of #275, thanks!

PR #252, opened 2020-06-17
tests: add test for http raw header

amirabell:
Add test to check that bug fix https://redmine.openinfosecfoundation.org/issues/3504 has been properly addressed.


victorjulien:
Merged with a small adjustment: the min-version set to 4.1.0 so we test there as well. Thanks!

PR #256, opened 2020-06-22
add ikev1 tests

fhonza:
This PR adds checks for a new IKEV1 parser for suricata, which is currently in review. 


tests/ikev1-rules/test.rules
```diff
@@ -0,0 +1,18 @@
+alert ikev1 any any -> any any (msg:"ikev1 initiator"; ikev1.spi_initiator; pcre:"/.*/"; sid:1;)
```
catenacyber, line 1: This is a strange signature matching for any initiator. Can we be more specific ?

catenacyber:
Is this replaced by https://github.com/OISF/suricata-verify/pull/301 ?

PR #258, opened 2020-06-26
tests/filestore Filestore Stream-depth test cases

jlucovsky:
Continuation of #255 
This commit adds the following test cases for filestore related
stream-depth configuration settings

1.`file-stream.stream-depth` exceeds `stream.reassembly.depth`.
2.`file-stream.stream-depth` is less than `stream.reassembly.depth`.
3.`file-stream.stream-depth` set to 0.

Redmine issue: [2506](redmine)
*Requires* Suricata PR [#5104](https://github.com/OISF/suricata/pull/5104)


victorjulien:
I tried this and it didn't pass, even through it should now, right? Can you have a look?

tests/filestore-v2.8-stream-depth/suricata.yaml
```diff
@@ -0,0 +1,17 @@
...
+outputs:
+  - eve-log:
+      enabled: yes
+      types:
+        - files
+  - file-store:
+      version: 2
+      enabled: yes
+      force-filestore: yes
+      stream-depth: 10000
```
catenacyber, line 13: I get a failure with
`Sub test #1: FAIL : grep "SC_WARN_FILESTORE_CONFIG(330)] - file-store.stream-depth value 10000 has no effect since it's less than stream.reassembly.depth value" stdout | wc -l | xargs`

jlucovsky:
Continued in #284

PR #259, opened 2020-06-26
tests/hassh Add test cases for hassh

MalakhatkoVadym:
This commit adds test cases for the ssh.hassh, ssh.hassh.string, ssh.hassh.server, ssh.hassh.server.string keywords.

Modifies https://github.com/OISF/suricata-verify/pull/257 by
1. Added testing for ssh banners only session.
2. Added testing for ssh session without reaching NewKeys message.

PR #262, opened 2020-06-30
tests/hassh Add test cases for hassh

MalakhatkoVadym:
This commit adds test cases for the ssh.hassh, ssh.hassh.string, ssh.hassh.server, ssh.hassh.server.string keywords.

Modifies https://github.com/OISF/suricata-verify/pull/259 by
1. Fixing comments in ssh-banner-only test.


victorjulien:
Can you also add checks to one of the tests that looks for the various hassh fields and confirms they are correct?

tests/ssh-banner-only/test.yaml
```diff
@@ -0,0 +1,18 @@
+requires:
+  min-version: 5.0.0
```
victorjulien, line 2: these should be 6.0.0

catenacyber:
> Can you also add checks to one of the tests that looks for the various hassh fields and confirms they are correct?

I am not sure I understand what you mean @victorjulien 

Do you mean using alerts as checks against the values is not good ? cf
https://github.com/OISF/suricata-verify/pull/262/files#diff-9dabb85edfb1a42602d72e3013624b99R3

MalakhatkoVadym:
https://github.com/OISF/suricata-verify/pull/263

PR #268, opened 2020-07-09
tests: conditional logging tests for http and dns

victorjulien:


victorjulien:
Merged as part of #271

PR #269, opened 2020-07-10
dnp3: new test with more detail

jasonish:
Also rename to make it clear its an dnp3/eve test.


victorjulien:
Merged in #271, thanks Jason

PR #272, opened 2020-07-13
Http gap 3359 v2

catenacyber:
Needs https://github.com/OISF/suricata/pull/4951 and https://github.com/OISF/libhtp/pull/296

1. Fixes http sticky server

* gap within the body ✅ 
* etc.


catenacyber:
Follows #234 by adding more tests

tests/http-gap-beyond-body/test.yaml
```diff
@@ -0,0 +1,39 @@
...
+        event_type: http
+        http.url: "/2"
+# response is not parsed in gap
+#       http.status: 200
+  - filter:
+      count: 1
+      match:
+        event_type: http
+        http.url: "/3"
+# response is no longer parsed after gap
```
catenacyber, line 31: For now, we only recover from gaps with known size.

In the future, we could try to recover the 3rd response even if we missed the content-length of the second one

tests/http-gap-whole-body/test.yaml
```diff
@@ -0,0 +1,51 @@
...
+      match:
+        event_type: fileinfo
+        fileinfo.size: 14
+        fileinfo.state: "CLOSED"
+        fileinfo.gaps: false
+  - filter:
+      count: 1
+      match:
+        event_type: fileinfo
+        fileinfo.size: 0
```
catenacyber, line 42: Maybe this is not the best behavior and we would rather like `fileinfo.size: 14`
Thoughts ?

victorjulien:
Can you split things so the GAP'ed flow is in its own test? This way we can limit that to 6.0.0 and keep the other test from failing on 5 and 4.1

catenacyber:
Done something in #277 
Is it good for you ?

PR #273, opened 2020-07-13
ssh: adds regression test for hassh kex parsing

catenacyber:
Cf https://github.com/OISF/suricata/pull/5161#issuecomment-657450916


catenacyber:
Unlike the other ssh.hassh tests, I used `--set app-layer.protocols.ssh.hassh=yes` instead of signatures to enable hassh

victorjulien:
Merged in #291, thanks!

PR #276, opened 2020-07-15
Dcerpc logging/v3

inashivb:


victorjulien:
Why is this failing on 4.1 and 5.0? Do we need any updates there?

inashivb:
> Why is this failing on 4.1 and 5.0? Do we need any updates there?

Yeah I think dcerpc never had logging before. I should add version check in test.yaml. Thanks.

catenacyber:
Does this depend on a Suricata PR ?

inashivb:
Replaced with #296

PR #278, opened 2020-07-22

catenacyber:

Needs https://github.com/OISF/suricata/pull/5176

Replaces #146 by setting min version to 6


catenacyber:
From what I gathered, transforms are not well handled by Suricata for HTTP

PR #279, opened 2020-07-22
Http gap 3359 v4

catenacyber:
Needs https://github.com/OISF/suricata/pull/4951 and https://github.com/OISF/libhtp/pull/296

1. Fixes http sticky server
The fix is to remove the gaped flow from the pcap

* gap within the body ✅ 
* etc.

Modifies https://github.com/OISF/suricata-verify/pull/277 by changing the pcap with removing only one packet


victorjulien:
Still getting ` ===> http-sticky-server: Sub test #1: FAIL : expected 28 matches; got 29 for filter {'count': 28, 'match': {'event_type': 'alert', 'alert.signature_id': 2}}`

catenacyber:
Indeed, empty commit...

PR #282, opened 2020-07-28
Http gap 3359 v5

catenacyber:
Needs https://github.com/OISF/suricata/pull/5172 and https://github.com/OISF/libhtp/pull/297

1. Fixes http sticky server
The fix is to remove the gaped flow from the pcap

* gap within the body ✅ 
* etc.

Modifies https://github.com/OISF/suricata-verify/pull/279 truly by changing the pcap with removing only one packet


catenacyber:
Replaced by #317

PR #283, opened 2020-07-31
Http2 basic v2

catenacyber:
cf https://github.com/OISF/suricata/pull/5179 to be modified

Modifies #274 by adding a test with a pcap with HTTP1 connection upgrade


catenacyber:
Now https://github.com/OISF/suricata/pull/5249

PR #286, opened 2020-07-31
tests: ERSPAN Type I packet decode if config

jlucovsky:
Continuation of #195 

This commit adds tests for ERSPAN Type I decoding based on configuration
settings.

This update restricts test `03` to version 5.


victorjulien:
Merged in #289, thanks!

PR #287, opened 2020-07-31
Http2 basic v3

catenacyber:
cf https://github.com/OISF/suricata/pull/5251

Modifies #283 with only rebase


catenacyber:
A part of the tests should succeed
But it still requires a fix for issue https://redmine.openinfosecfoundation.org/issues/3866

PR #294, opened 2020-08-04
snmp rdp alert metadata/v1

jasonish:
- snmp-v2c-get: test snmp metadata in alert
- rdp-protocol: test rdp metadata in alert

Requires Suricata PR.

### DIFF
--- a/tests/rdp-protocol/suricata.yaml
+++ b/tests/rdp-protocol/suricata.yaml
@@ -7,6 +7,7 @@ outputs:
       filetype: regular
       filename: eve.json
       types:
+        - alert
         - rdp
         - flow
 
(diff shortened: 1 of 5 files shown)


victorjulien:
Is the snmp one supposed to include more than just a rule?

jasonish:
> Is the snmp one supposed to include more than just a rule?

Ah yeah. Its supposed to check that there is an snmp object in the alert.

jasonish:
> Ah yeah. Its supposed to check that there is an snmp object in the alert.

Resolved within this PR.

victorjulien:
Had to add a fix for suricata 5 too.

Merged in #298

PR #299, opened 2020-08-12
tests: Remove obsoleted standalone output test

jlucovsky:
This commit removes a test obsoleted with removal of standalone output
logging.


victorjulien:
Since we'll continue to support this in suri 4 and 5, wouldn't it be better to keep the test but limit it to those versions?

jlucovsky:
Continued in #316

PR #300, opened 2020-08-15
Give test count as output not check count

inashivb:
With the current setup, after running suricata-verify, the output would
give stats about the passed, failed or skipped checks. However, it is
cleaner to get the total stats about tests only.

The priority order for the output is:
FAILED
SKIPPED
PASSED

i.e. if a check or sub-test has failed, the entire test will be counted
as failed in the final output. If all the checks have been skipped,
the test will be counted as skipped in the final output else passed.

Closes redmine ticket https://redmine.openinfosecfoundation.org/issues/3172.

### DIFF
--- a/run.py
+++ b/run.py
@@ -620,8 +620,10 @@ class TestRunner:
             if check_value["check_sh"]:
                 return check_value
 
-        if not check_value["failure"]:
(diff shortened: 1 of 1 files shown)


victorjulien:
Merged in #315, thanks!

PR #301, opened 2020-08-17
added ikev1 test

fhonza:

Updated IKEv2 checks, this parser is now combined with IKEv1.


victorjulien:
Any idea why this fails on 4.1 and 5?
`===> ikev2-weak-dh: FAILED: got exit code 1, expected 0`

fhonza:
> Any idea why this fails on 4.1 and 5?
> `===> ikev2-weak-dh: FAILED: got exit code 1, expected 0`

I guess this comes from renaming output from ikev2 to common ike, according to the corresponding suricata PR merging both protocol versions.
When you are testing older version as well i suppose once implemented output doesn't change normally?
I'll split the output in the suricata-PR and update this PR, then it should work here again....

PR #302, opened 2020-08-18
nsh: Add tests for NSH headers types 1 and 2

carlgsmith:
To be used for once NSH header support is accepted into Suricata


catenacyber:
https://github.com/OISF/suricata/pull/5309

tests/decode-nsh-type1/test.yaml
```diff
@@ -0,0 +1,19 @@
+# *** Add configuration here ***
```
catenacyber, line 1: You can remove this header line

catenacyber (commented):
This looks good to me.
You can add a README.md file to each test, so as to explain its purpose

tests/decode-nsh-type1/test.yaml
carlgsmith: > You can remove this header line

@catenacyber Do I need something this in test.yaml as well?

`requires:
    min-version: 6
`

catenacyber: Yes indeed, as NSH is not yet in version 5

carlgsmith:
Updated by https://github.com/OISF/suricata-verify/pull/307

PR #304, opened 2020-08-25
dnp3: test of dnp3 flow direction fix

ilya-bakhtin:


catenacyber:
Should this not modify `dnp3-midstream-confirm` ?

ilya-bakhtin:
From my point of view, dnp3-midstream-confirm does not make any sense. It based on a manually and wrongly patched file.
Well, this patch does not alter dnp3-midstream-confirm. I do not know whether its update or complete removal is better.

catenacyber:
>  I do not know whether its update or complete removal is better.

Update looks better to me

> From my point of view, dnp3-midstream-confirm does not make any sense

To have some background information on this, see https://github.com/OISF/suricata/pull/5093 and the previous and following Pull Requests.
If I remember correctly, the point was to create a pcap that triggered this buffer overflow behavior fixed by https://github.com/OISF/suricata/pull/5093/commits/9673d2ad03d7fb220d1c7bd9891abfe786eb6a74
This buffer overflow led us to the other problem of the probing parser not finding the right direction.

> It's based on a manually and wrongly patched file.

Does this pcap not induce the buffer overflow fixed by the mentioned commit ?


Then, the question is if the test dnp3-midstream-confirm can test both the buffer overflow fix and the probing parser finding the right direction

ilya-bakhtin:
well, i finally understood the point.
however to proof buffer overflow fix, it's necessary  to patch DIR flag in the packet instead of ports swapping

catenacyber:
> it's necessary to patch DIR flag in the packet instead of ports swapping

Maybe indeed.
Can you make the patch for this ? and replace this PR ?

ilya-bakhtin:
Sure i will
Immediately i do not have time; i hope i'll update the PR tomorrow

ilya-bakhtin:
I've made the required modification (forced update)

tests/dnp3-toclient-start/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+args:
+- -k none --set stream.midstream=true --set app-layer.protocols.dnp3.enabled=yes
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        app_proto: dnp3
+        flow.pkts_toserver: 2
```
catenacyber, line 16: Would it be possible to have only 3 packets in the pcap ? So as to have different values for `flow.pkts_toserver` and `flow.pkts_toclient `

catenacyber (changes requested):
Thanks.
This looks good, except for one small change request.

Did you try to trigger the buffer overflow by reverting the commit fixing it in Suricata ?

For Pull requests, could you create a new Pull Request when you make changes rather than force-push ? (and name your branch `ilya-bakhtin:dnp3-toclient-start-v2`)

One last thing : I think we will have a bit of trouble as we will need to back port the fix to version 5 in the same time.

@jlucovsky thoughts ?
Ideally, we should merge at the same time 
- the suricata-verify PR fixing a test
- the suricata PR fixing the behavior
- the suricata back port PR so that CI does not get broken

ilya-bakhtin:
Finally, i realized that single packet pcap can not be used to verify buffer overrun issues.
At the same time, my example definitely results in the mentioned buffer overrun at packet #4 on a version before buffer overrun fix.
Although i believe the dest_port verification is quite enough, i added 5th packet to show some asymmetry.

So i removed dnp3-midstream-confirm, since this new test covers both cases

Please see https://github.com/OISF/suricata-verify/pull/310

catenacyber:
Closing in favor of #310

PR #311, opened 2020-08-27
stream/tcp: Test verifies the behavior when direction of TCP flow is … - v2

ilya-bakhtin:
…changed by the probing parser.

pcap file contains 2 http transactions. The request is missing for the first one.
The second transaction is fully complete. So eve.json must contain one and only anomaly event.


tests/tcp-stream-after-swap/test.yaml
```diff
@@ -0,0 +1,15 @@
...
+# disables checksum verification, and uses midstream
+args:
+- -k none --set stream.midstream=true
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: anomaly
+        anomaly.event: "UNABLE_TO_MATCH_RESPONSE_TO_REQUEST"
```
catenacyber, line 15: Should not we test for the bytes and packets to client and to server ?

victorjulien:
Replaced by #313

ilya-bakhtin:
Well, we may add common flow details verification. Bytes number instead of packets must be compared since the change fixes reassembly behavior
https://github.com/OISF/suricata-verify/pull/313

PR #318, opened 2020-09-01
use python 3; assign copyright - v1

jasonish:


jasonish:
@victorjulien Copyright stuff look OK?

PR #316, opened 2020-09-01
tests: Skip obsoleted standalone output test on 6+

jlucovsky:
Continuation of #299 

This commit restricts the test case to versions less than 6. It's been deprecated in 6.0+

PR #319, opened 2020-09-03
tests: Skip obsoleted standalone output test on 6+

victorjulien:
This commit restricts the test case to versions less than 6. It's been
deprecated in 6.0+


victorjulien:
#316 rebased

PR #322, opened 2020-09-07
datasets: add test for loading 1m StringSets

inashivb:


victorjulien:
Merged as part of #450, thanks!

PR #324, opened 2020-09-07
ikev1: added ikev1 test - v3

fhonza:


catenacyber (changes requested):
Thanks for this work.

CI is not happy :-(

Could you review your commits so that each one compiles ?

And have them formatted ok with `git rebase master -x "git clang-format HEAD^"` ?

tests/ikev2-weak-dh/test.yaml
```diff
@@ -3,6 +3,7 @@ requires:
     - HAVE_LIBJANSSON
     - RUST
   min-version: 4.1.0
+  lt-version: 6
```
catenacyber, line 6: @jasonish this test is getting deprecated and replaced by `ike-weak-dh`
Is this the way to go ?

@fhonza could you modify ikev2-weak-dh` README (or add a comment in this yaml) to explain this ?

fhonza: with the unification of ikev1 and ikev2 into ike both parsers now use output-json-ike.c, which just adds the event_type "ike" (if i've debugged correctly), which is breaking the old test. Perhaps i can try to use either ikev1 or ikev2 depending on the protocol version.
In this function (JsonIKELogger(...)) I have the state as a parameter, perhaps it's possible to check there.

In this case the existing test could be left untouched, would be better?

catenacyber: I think event_type "ike" is good.
@jasonish what do you think ?

jasonish: Just "ike" would be idea I think.

catenacyber: Does the framework enable to have some checks with some version and some others checks with other versions ?
cc @inashivb 

That is what we want here I think.
The pcap is the same.
Previously, we checked `ikev2` field in eve.json, and now we will want it to be simply `ike`

catenacyber:
@fhonza what are the changes with your force-push ?

fhonza:
> @fhonza what are the changes with your force-push ?

just pushed the wrong branch, was preparing v4 PR, just added a comment on the deprecated test...

but discussion seems to go on, i'll wait a bit with it :)

tests/ikev2-weak-dh/test.yaml
catenacyber: Thanks Shivani.
Could you do this @fhonza ?

fhonza: great! created a new PR...

PR #325, opened 2020-09-08
Http2 basic v6

catenacyber:
Adds http2 connection upgrade from http1 test

Modifies #320 with translating HTTP connect unit tests to suricata-verify


catenacyber:
By the way, do we want to merge dummy code to generate these pcaps ?
```
import sys
import binascii
from threading import Thread
import time
import socket

# Create a pcap from a htp test file
# Launches a server on port PCAP_TCP_PORT
# Launches a client in another thread that connects to it
# Both client and server read the input file
# And they send and receive data as described (without analysing it)
# So, you have one tcp packet per line

PCAP_TCP_PORT = 8001

class ServerThread(Thread):

    def __init__(self, filename):
        Thread.__init__(self)
        self.filename = filename

    def run(self):
        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        s.bind(("127.0.0.1", PCAP_TCP_PORT))
        s.listen(1)
        conn, addr = s.accept()
        f = open(self.filename)
        sending = ""
        receiving = ""

        for l in f.readlines():
            data = l[4:-1].decode('string_escape')
            if l[:4] == "s2c ":
                conn.send(data)
                print "server sent", len(data)
            elif l[:4] == "c2s ":
                data = conn.recv(len(data))
                print "server recvd", len(data)

        conn.close()
        s.close()
        f.close()


class ClientThread(Thread):

    def __init__(self, filename):
        Thread.__init__(self)
        self.filename = filename

    def run(self):
        time.sleep(1)
        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
        s.connect(("127.0.0.1", PCAP_TCP_PORT))
        f = open(self.filename)
        sending = ""
        receiving = ""

        for l in f.readlines():
            data = l[4:-1].decode('string_escape')
            if l[:4] != "s2c ":
                s.send(data)
                print "client sent", len(data)
            elif l[:4] != "c2s ":
                data = s.recv(len(data))
                print "client recvd", len(data)

        s.close()
        f.close()

t1 = ServerThread(sys.argv[1])
t2 = ClientThread(sys.argv[1])

# Launch threads
t1.start()
t2.start()

# Wait for threads to finish
t1.join()
t2.join()
```

PR #327, opened 2020-09-08
run.py: use an unversioned python in #!

jasonish:
This should be more compatible across more distributions.


jasonish:
Rebased here https://github.com/OISF/suricata-verify/pull/355.

PR #328, opened 2020-09-11

catenacyber:
Modifies #51 : min version is now 6.0.0

cf https://github.com/OISF/libhtp/pull/309


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/691

PR #330, opened 2020-09-11
tests: Add test for file_data w/transform

jlucovsky:
Continuation of #230 

This update adds test cases for the `compress-whitespace` transform that validates pre-checks on the buffer are performed.


victorjulien:
Merged, thanks.

PR #334, opened 2020-09-23
ikev1: added ikev1 test - v4

fhonza:


tests/ikev2-weak-dh/test.yaml
```diff
@@ -31,3 +32,38 @@ checks:
         ikev2.payload[1]: KeyExchange
         ikev2.payload[2]: SecurityAssociation
         ikev2.payload[3]: NoNextPayload
+
+  - filter:
+      count: 1
+      version: 5
```
catenacyber, line 38: @inashivb can we do only one filter for versions 4 and 5 ?

tests/ikev2-weak-dh/test.yaml
```diff
@@ -31,3 +32,38 @@ checks:
...
+        ikev2.payload[0]: Nonce
+        ikev2.payload[1]: KeyExchange
+        ikev2.payload[2]: SecurityAssociation
+        ikev2.payload[3]: NoNextPayload
+
+  - filter:
+      count: 1
+      min-version: 6
+      match:
+        event_type: ike
```
catenacyber, line 57: @fhonza Can you add a simple comment to say that the only change is renaming the event from ikev2 to generic ike ?

catenacyber:
CI is broken for now for versions 4 and 5 cf https://github.com/OISF/suricata/pull/5440

catenacyber:
Follows #324

tests/ikev2-weak-dh/test.yaml
fhonza: yes, will add one...

tests/ikev2-weak-dh/test.yaml
inashivb: No, we could check for any version.

PR #335, opened 2020-09-24
Failing test for dcerpc last response not logged

inashivb:
This test should ideally pass as there exists a request response pair for call ID of 20.


inashivb:
Fixed by https://github.com/OISF/suricata/commit/9f9c29a14a91f86c1bda92285e876e026dc9e3a4

PR #336, opened 2020-09-25
dns: add test for NULL record

cccs-sadugas:
Add test case to show that NULL record types are being logged.

**Suricata PR**
- [TODO: insert link here]


cccs-sadugas:
Reopened as #337 .

PR #337, opened 2020-09-25
dns: add test for NULL record

simdugas:
Add test case to show that NULL record types are being logged.

**Suricata PR**
- [#5446](https://github.com/OISF/suricata/pull/5446)


victorjulien:
Part of #366

PR #338, opened 2020-09-30
ikev1: added ikev1 test - v5

fhonza:


tests/ikev2-weak-dh/test.yaml
```diff
@@ -31,3 +32,39 @@ checks:
         ikev2.payload[1]: KeyExchange
         ikev2.payload[2]: SecurityAssociation
         ikev2.payload[3]: NoNextPayload
+
+  - filter:
+      count: 1
+      version: 5
```
catenacyber, line 38: Can you not do a max-version 5 ?

fhonza: don't think so, in the run.py is only ```version``` and ```min-version``` defined

catenacyber:
replaced by #351

PR #345, opened 2020-10-20
tests: Add exact match argument

jlucovsky:
This commit extends the CLI with a flag to treat the "pattern"
argument(s) as an exact match instead of a substring match.


jasonish (approved):
Looks and tests OK to me.

PR #351, opened 2020-10-26
ikev1: added ikev1 test - v7

fhonza:


catenacyber:
Could we also have a test with
- a pcap containing IKEv1 and IKEv2
- one signature for IKEv1 only
- one signature for IKEv2 only
- one signature for IKE whatever the version
- and the checks that these signatures trigger as expected, see example with HTTP1 or 2 https://github.com/OISF/suricata-verify/pull/377

satta:
Next PR:   https://github.com/OISF/suricata-verify/pull/465

fhonza:
> Could we also have a test with
> 
>     * a pcap containing IKEv1 and IKEv2
> 
>     * one signature for IKEv1 only
> 
>     * one signature for IKEv2 only
> 
>     * one signature for IKE whatever the version
> 
>     * and the checks that these signatures trigger as expected, see example with HTTP1 or 2 #377

There is currently no keyword to test for a specific version, signatures will always alert independently of the version.
Is that required?

PR #355, opened 2020-11-03
run.py: use an unversioned python in #! - v2

jasonish:
This should be more compatible across more distributions.

Previous PR: https://github.com/OISF/suricata-verify/pull/327
Changes: Rebased.

### DIFF
--- a/run.py
+++ b/run.py
@@ -1,4 +1,4 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python
 #
 # Copyright (C) 2017-2020 Open Information Security Foundation
 #


victorjulien:
My Fedora 31 builder is the only one to fail with this
```
$ ../suricata-verify/run.py
/usr/bin/env: 'python': No such file or directory
```
I can of course call it with an explicit python version. Should I? Or is something else going on?

jasonish:
> I can of course call it with an explicit python version. Should I? Or is something else going on?

I think you should.  I clean install of minimal Fedora in a CI environment is unlikely to have it. An installation setup for actual developer (ie: Workstation or Server) will have `python-unversioned-command` installed which provides the unversioned name. I don't think there is some distro standard around this.

abungay:
I like this PR since createst.py also uses an unversioned python shebang line, and this change reduces confusion about the required python version for suricata-verify.

Recently, on my new development setup I had looked at the shebang in run.py and thought that I needed to install python2 instead of python3 in my suricata-verify python virtual environment, but then later trying to run createst.py it failed since createst.py requires python3. I then had to delete the new python virtual environment and re-create one with python3 to get running both python files in suricata-verify to work.

PR #356, opened 2020-11-03
dns: add test for SRV EVE-JSON output

satta:
This complements https://github.com/OISF/suricata/pull/5529 by providing a test pcap with SRV query/answer DNS traffic.


victorjulien:
Part of #366

PR #361, opened 2020-11-10
feature #4092

claudius96:
making --debugfailed to continue it's work even if we suffix it with --fail


inashivb:
Hi @claudius96 !
Would you be interested in/have time to look into this? I'm cleaning up old PRs from the repo. Thank you for the work you put into this feature. :)

PR #362, opened 2020-11-10

catenacyber:

Needs https://github.com/OISF/suricata/pull/5554

Replaces #278 with rebase


catenacyber:
ping @jasonish 
Ready to merge as tests now pass with merging of https://github.com/OISF/suricata/pull/5573

victorjulien:
Merged in #370, thanks!

PR #367, opened 2020-11-17
Add tests for DCERPC/UDP

inashivb:
Suricata PR has already been merged.


victorjulien:
Merged in #370, thanks!

I updated the smb checks to remove the `smb.request_done` and `smb.response_done` fields. These are only logged if Suricata is built with `--enable-debug`.

PR #357, opened 2020-11-05
tests: ICMPv4 header tests

jlucovsky:
Additional test cases for `icmpv4.hdr` keyword.

PR #348, opened 2020-10-23
bug-3616: New test with mixed file_data/transform rules

jlucovsky:
This PR adds a test case to demonstrate the changes in issue 3616

- file_data
- file_data + strip_whitespace (http, smtp)
- file_data  + urldecode (http)
- file_data + fast_pattern

PR #369, opened 2020-11-20
Next/20201120/v4

victorjulien:
#357 #362 #367 #348 

Removed the asan leak suppression, so any leak should now be an error if asan is enabled.


victorjulien:
Replaced by #370

PR #373, opened 2020-12-04
run: convert cmp tests to python

TharushiJay:
Some tests use a check.sh script to run "cmp" on 2 files to check that they are the same. This could be converted to an internal check and become part of test.yaml.

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/3973

Example test: `output-tcp-data`

Possible YAML syntax for test.yaml:
```
checks:
  - file-compare:
      filename: tcp-data.log
      expected: expected/tcp-data.log
```

### DIFF
--- a/run.py
+++ b/run.py
@@ -37,6 +37,7 @@ import glob
 import re
 import json
 import unittest
+import filecmp
 from collections import namedtuple
 
 import yaml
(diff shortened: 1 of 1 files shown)


jasonish:
Sorry I left this so long.  Can you:
- add an example to the readme
- convert at least one test from using cmp to this check. `tests/datasets-01` is a good one as it already has a test.yaml, and a check.sh that only does the file compare.

jasonish (changes requested):
Changes requested.

TharushiJay:
Changes continued in #447

PR #375, opened 2020-12-06
Http2 files v3

catenacyber:
Right test condition for http2 files
Adds a check that there is no anomaly in http2 basic test
check for file decompression
adding a check for brotli decompressed files


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/397

PR #376, opened 2020-12-07
Bug 78/v3

inashivb:
Closes redmine ticket: https://redmine.openinfosecfoundation.org/issues/3055


catenacyber:
Could you add README files to explain the purpose of these tests ? (point to the redmine 78 ticket if it is the case)

inashivb:
Replaced with #381

PR #379, opened 2020-12-12
bug-4199: Tests for shared http request body

jlucovsky:
Add test case for shared rules using http client body and client body and transforms.


victorjulien:
Another test case: https://forum.suricata.io/t/missing-alert-http-request-body-alone-and-http-request-body-url-decode-in-a-single-rule/942/2

jlucovsky:
Continued in #420

PR #380, opened 2020-12-14
run: parallel tests capability

TharushiJay:
SU keeps getting tests added and after some point the volume would become quite big maybe thousand(s)+. In that case maybe running tests in multithreading/parallel test runs would help speed up the checks in terms of time.

### DIFF
--- a/run.py
+++ b/run.py
@@ -37,6 +37,7 @@ import glob
 import re
 import json
 import unittest
+import multiprocessing as mp
 from collections import namedtuple
 
 import yaml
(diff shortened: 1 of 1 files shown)


inashivb:
Hi @TharushiJay !
Thanks. :) This is good work. While elapsed time has certainly come down by a huge factor, I think something might be up somewhere with the async handling. I tried s-v
`master`:
```
PASSED:  786
FAILED:  0
SKIPPED: 44

real	1m30.655s
user	0m51.947s
sys	0m31.670s
```

On your branch:
Output 1:
```
PASSED:  781
FAILED:  0
SKIPPED: 44

real	0m18.947s
user	1m10.025s
sys	0m41.855s
```
Output 2:
```
PASSED:  774
FAILED:  0
SKIPPED: 42

real	0m19.405s
user	1m11.873s
sys	0m42.901s
```
and this keeps on changing.. Notice the number of passed and skipped tests change on every run.

jasonish:
I get the same test count issue as well. But on the whole this looks like a great start.

Currently we print the test name without a new line, then when the test is done we output the result and the new line. Thats what leads to the bad display output.  This will have to change to write the test name and the result at the same time which should clean up the output.

TharushiJay:
Thanks for the feedback @inashivb @jasonish! I will look into the test count issue and fix the display output 👍

TharushiJay:
Continued in #385

PR #381, opened 2020-12-16
Bug 78/v4

inashivb:


tests/bug-78-http-uri/test.yaml
```diff
@@ -0,0 +1,73 @@
...
+ - -k none
+
+checks:
+- filter:
+    count: 1
+    match:
+      alert:
+        action: allowed
+        category: ''
+        gid: 1
```
catenacyber, line 15: Maybe some fields are not important... cf creates...

catenacyber:
Looks overall good to me, event if it can be improved by removing some parts of the checks (like undefined gid of the signature where default value gets used)

inashivb:
I was just redoing this and found out that suricata-verify would fail in case any of the inner levels of yaml tree do not have **all** the fields available. I don't know if it was always like this or was introduced recently. I'll check it out. Created ticket: https://redmine.openinfosecfoundation.org/issues/4276

inashivb:
Replaced with #429

PR #377, opened 2020-12-08
Adds checks with http.uri keyword and HTTP2 traffic

catenacyber:

PR #382, opened 2020-12-16
Adds checks with http.uri keyword and HTTP2 traffic

catenacyber:
Replaces #377 with comments taken into account (ie more checks)

needs  OISF/suricata#5660


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/383

PR #383, opened 2020-12-17
Http2 keywords v3

catenacyber:
Replaces #382 with added check for `http.user_agent` keyword

needs next version of OISF/suricata#5660

PR #384, opened 2020-12-17
Http2 keywords v4

catenacyber:
Replaces #383 with added check for `http.stat_msg` keyword

needs next version of OISF/suricata#5668


victorjulien:
Can you do a new version of this that works on all branches now that https://github.com/OISF/suricata/pull/5875 is merged? It seem the `alert http1` rule file additions cause the failures.

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/452

I think `alert http1` should indeed fail for versions 6... let's check

PR #389, opened 2020-12-24
run: parallel tests capability/v3

TharushiJay:
Running tests in parallel test runs to help speed up the checks in terms of time.

Changes from last PR:

- Added a command line parameter to provide the number of jobs to run (eg: `-j 2`)
- Fixed the output in `--debug-failed` command line option

### DIFF
--- a/run.py
+++ b/run.py
@@ -37,6 +37,7 @@ import glob
 import re
 import json
 import unittest
+import multiprocessing as mp
 from collections import namedtuple
 
 import yaml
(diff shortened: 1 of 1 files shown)


jasonish:
This looks good.  I'd like to see a little more detail in the commit message though..  Including something like "used the Python multiprocessing module to add paralellel stuff"... "added a -j parameter to control the number of jobs"... "defaulting to the number of cores found by the mp module"..

The only code change would be to maybe log the number of concurrent jobs will be run before starting the tests.. I think it could be a useful thing to log, especially when running in CI environments where we may not know the exact details before hand.

Thanks!

jasonish:
Oops.. Found another issue. Modify some test to fail.  In my case I modified `dns-eve-type-filtering` so it would fail. I then run:
```
python ../../suricata-verify/run.py -j2 dns
```
all is good. You see the failure.  But then add the `--fail` command line option on, for example:
```
python ../../suricata-verify/run.py -j2 --fail dns
```
and then tests run but then `run.py` appears to get stuck and never returns.

TharushiJay:
> Oops.. Found another issue. Modify some test to fail. In my case I modified `dns-eve-type-filtering` so it would fail. I then run:
> 
> ```
> python ../../suricata-verify/run.py -j2 dns
> ```
> 
> all is good. You see the failure. But then add the `--fail` command line option on, for example:
> 
> ```
> python ../../suricata-verify/run.py -j2 --fail dns
> ```
> 
> and then tests run but then `run.py` appears to get stuck and never returns.

Sorry about that. I will look into the changes thoroughly and make sure the command line options are working.

TharushiJay:
Changes continued in #395

PR #391, opened 2020-12-30
smtp-md5: test md5 logging of smtp subject and body - v1

jasonish:
New test to test the MD5 logging of email body and subjects. Requires Suricata PR to fix https://redmine.openinfosecfoundation.org/issues/4245.


tests/smtp-md5/test.yaml
```diff
@@ -0,0 +1,27 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+    - HAVE_NSS
+
+args:
+  - --set app-layer.protocols.smtp.mime.body-md5=yes
```
victorjulien, line 7: is there a reason to set it here instead of in the yaml?

jasonish: Ah, no..  Was hoping to get by without adding any app-layer section at first.  But turns out our smtp tests don't actually pass with 0 config suggesting our default suricata.yaml doesn't match the smtp app-layer defaults.  But thats a problem for another day.  Will move it into the suricata.yaml.

jasonish:
Replaced by https://github.com/OISF/suricata-verify/pull/393.

PR #396, opened 2021-01-08
Feat 3357/v5

TharushiJay:
Implements parallel tests capability using the python multiprocessing module.

Changes from last PR:

- Set default maximum job count to 8
- Set default python version to python3

### DIFF
--- a/run.py
+++ b/run.py
@@ -37,6 +37,7 @@ import glob
 import re
 import json
 import unittest
+import multiprocessing as mp
 from collections import namedtuple
 
 import yaml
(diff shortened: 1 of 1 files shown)


jasonish:
@victorjulien Can you give this another try?

jasonish:
Run into something I wasn't expecting. This doesn't work on Windows and Mac.  Works great on Linux though.  My initial hunch was that it had something to do with the `mp` module being initialized in global module space inside of `main` leading me to try this:

https://github.com/jasonish/suricata-verify/commit/19d7638b1c95f7c535b26090fe1c1b35a911c951

but then the tests aren't run at all.

It might have something to do with the discussion under "Explicitly pass resources to child processes" over at https://docs.python.org/3/library/multiprocessing.html#multiprocessing-programming if you could have a look.

Another option is to fallback to the old way on Windows and Mac.  Even specifying -j1 failed on Windows/Mac at the moment.

TharushiJay:
Changes continued in #400

PR #397, opened 2021-01-08
Http2 files v4

catenacyber:
Right test condition for http2 files
Adds a check that there is no anomaly in http2 basic test
check for file decompression
adding a check for brotli decompressed files

Modifies #375 with adding per-filter features check
This changes the framework

PR #398, opened 2021-01-11
Http2 files v5

catenacyber:
Right test condition for http2 files
Adds a check that there is no anomaly in http2 basic test
check for file decompression
adding a check for brotli decompressed files

Modifies #397 with extending the README


catenacyber:
https://redmine.openinfosecfoundation.org/issues/4185

PR #401, opened 2021-01-15
Smb andx 3475 v1

catenacyber:
SMB evasions tests including AndX


victorjulien:
Merged in #459, but limited to suri 7.

PR #410, opened 2021-01-19
tests: add dnp3 tests

victorjulien:
Based on pcaps from:
https://github.com/bro/bro/tree/master/testing/btest/Traces/dnp3


catenacyber:
Why did you add these ? Any related tickets about DNP3 ?

victorjulien:
Wanted to improve coverage for dnp3 parser and logging code. Even with these tests added its among the lowest we have:
https://codecov.io/gh/OISF/suricata/src/master/src/app-layer-dnp3-objects.c
https://codecov.io/gh/OISF/suricata/src/master/src/output-json-dnp3-objects.c

catenacyber:
As part of the ideal fuzzing, we should run targets on corpus as part of CI
cf https://redmine.openinfosecfoundation.org/issues/4125

That should increase part

PR #99, opened 2019-07-17
Adds test about DNS probing with junk request

catenacyber:
Test case for https://github.com/OISF/suricata/pull/4044

The point is to recognize DNS when first UDP request is junk

PR #415, opened 2021-01-21
Adds test about DNS probing with junk request

catenacyber:

Replaces #99 with rebase


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/417

PR #416, opened 2021-01-21
Adds test for http post data decompression

catenacyber:
Modifies #329 with rebase


catenacyber:
needs https://github.com/OISF/suricata/pull/5757

victorjulien:
In future tests please review how it works against all branches. This one fails for all while the referenced PR only fixes it for master.

Will merge after fixing that up in #421

catenacyber:
> In future tests please review how it works against all branches. This one fails for all while the referenced PR only fixes it for master.

Indeed, I initially thought there would be backports...

PR #417, opened 2021-01-21
Protocol detection 2757 v3

catenacyber:

Replaces #415 with a commit fixing the test for the new behavior.
If we have only one side of the communication, and failed protocol detection, we only log this partial failure
That means `app_proto_ts: failed` instead of `app_proto: failed`


catenacyber:
Needs https://github.com/OISF/suricata/pull/5759

catenacyber:
Force-pushed with rebase for new suricata PR

catenacyber:
Maybe we will want min-version: 7 instead of back ports...

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/813

PR #420, opened 2021-01-24
bug-4199: Tests for shared http request body 

jlucovsky:
Continuation of #379

This update adds a test case from the forum post https://forum.suricata.io/t/missing-alert-http-request-body-alone-and-http-request-body-url-decode-in-a-single-rule/942/2


jlucovsky:
Merged in https://github.com/OISF/suricata-verify/pull/463

PR #421, opened 2021-01-25
Adds test for http post data decompression

victorjulien:
#416 rebased


victorjulien:
@catenacyber this won't work for 5 and 6 w/o changes there, right?

catenacyber:
> this won't work for 5 and 6 w/o changes there, right?

Right :-) there is no yaml option to do it
Thanks for fixing the testcase

PR #426, opened 2021-01-31
tests: add test for app proto bittorrent-dht v1

abungay:
Adds a test case for feature 3086.

Companion PR https://github.com/OISF/suricata/pull/5814


abungay:
Opened PR v2 - https://github.com/OISF/suricata-verify/pull/466

PR #427, opened 2021-02-01
Validate eve.json output /v1

TharushiJay:
Validates eve.json output using a JSON schema.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,72 @@
+{
+    "type" : "object",
+     "properties": {
+        "timestamp": {
+            "type": "string",
+	        "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+-\\d+$",
+            "optional" : false
+	                },
+        "flow_id" : {
+            "type": "integer",
+            "optional" : true
+                    },
+        "pcap_cnt" : {
+            "type": "integer",
+            "optional" : true
+                    },
+        "event_type" : {
+            "type": "string",
+            "optional" : false
+                    },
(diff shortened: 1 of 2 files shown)


schema.json
```diff
@@ -0,0 +1,72 @@
+{
+    "type" : "object",
+     "properties": {
+        "timestamp": {
+            "type": "string",
+	        "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+-\\d+$",
```
jasonish, line 6: Need to handle a "+" in the timezone offset as well for timestamps like:
```2020-07-30T14:53:17.457474+0000```
this should probably do it:
```
            "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+[+\\-]\\d+$",
```

File: schema.json [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+            "optional" : false
+	                },
```
jasonish, line 8: minor: indenting seems off here.  vscode shift+ctrl+i seems to clean it up nicely.

jasonish:
Nice work, this will be very useful.  I'd like to see some modifications to the output:
- Don't output anything except errors.  And when there is an error, output the test name, etc.
- If `-v` is provided, then print the "===> test_name: OK/FAIL" type thing.
- Include a summary at the end of how many eve.json files were checked and pass fail count.

jasonish (changes requested):

TharushiJay:
Changes continued in #428

PR #428, opened 2021-02-03
Validate eve.json output /v2

TharushiJay:

Changes from last PR : 
- Changed the format of the output message
- Added a -v command line argument
- Included the (checked, passed, failed) eve.json file count summary at the end

### DIFF
--- a/check-eve.py
+++ b/check-eve.py
@@ -1,0 +1,66 @@
+#! /usr/bin/env python3
+
+import sys
+import os
+import os.path
+import argparse
+import json
+from jsonschema import validate
+from jsonschema.exceptions import ValidationError
+
+def validate_json(args, dirpath, schema):
+    json_filename = os.path.join(dirpath, 'eve.json')
+    testname = os.path.basename(os.path.dirname(dirpath))
+
+    status = "OK"
+    errors = []
+
+    with open(json_filename) as f:
+        for line in f:
+            obj = json.loads(line)
+            try:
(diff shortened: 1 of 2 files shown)


check-eve.py
```diff
@@ -0,0 +1,66 @@
+#! /usr/bin/env python3
+
```
jasonish, line 2: This needs a license and copyright. You can copy it from `run.py`, but use a date of just 2021.

jasonish (changes requested):

jasonish:
I have a few more changes I'd like to see.. Sorry, kind of defining the use-cases as we go.

If an argument is provided, for example: `check-eve /path/to/eve.json`, this script should just that single file.  But if that argument is a directory, then it should use that directory as the topdir.  If no provided, the current behaviour is good. This will allow us to use this script in other contexts outside of the suricata-verify repo.

TharushiJay:
> I have a few more changes I'd like to see.. Sorry, kind of defining the use-cases as we go.
> 
> If an argument is provided, for example: `check-eve /path/to/eve.json`, this script should just that single file. But if that argument is a directory, then it should use that directory as the topdir. If no provided, the current behaviour is good. This will allow us to use this script in other contexts outside of the suricata-verify repo.

Thanks for the feedback on the changes. It's easier for me this way, since we're increasing complexity as we go. 👍

TharushiJay:
Changes continued in #431

PR #431, opened 2021-02-09
Validate eve.json output /v3

TharushiJay:

Changes from last PR :

- Included license and copyright
- Added the use case where an argument may be provided as a single file or a directory

### DIFF
--- a/check-eve.py
+++ b/check-eve.py
@@ -1,0 +1,113 @@
+#! /usr/bin/env python3
+#
+# Copyright (C) 2017-2021 Open Information Security Foundation
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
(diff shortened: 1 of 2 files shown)


check-eve.py
```diff
@@ -0,0 +1,113 @@
+#! /usr/bin/env python3
+#
+# Copyright (C) 2017-2021 Open Information Security Foundation
```
jasonish, line 3: This should just be 2021 as its a new file this year.

jasonish:
This looks a little odd when I run on a file outside the suricata-verify tree:
```
./check-eve.py  -v /var/log/suricata
```
In this case it would be better to see the whole filename.

I wonder, if we should just use the whole filename in the output.  Ideally if its running in the suricata-verify directory, we'd chop off the common prefix to be less verbose?

jasonish:
Can you also add in a test for an array..  There are eve.json records with a vlan in them.. Looks like:
```
{
  "timestamp": "2019-07-11T14:13:06.980924+0000",
  "flow_id": 1319518172542908,
  "pcap_cnt": 1,
  "event_type": "alert",
  "vlan": [
    6
  ],
```
This should always be an array.

jasonish:
What I'd like to look at next is adding this directly to run.py.  After each test in run.py, execute this script with the tests output directory as the argument.  That way we can fail the test with an `invalid json schema` error.

jasonish (changes requested):

TharushiJay:
Changes continued in #440]

PR #432, opened 2021-02-09
Update tests for initial tcp flags capture

rtkjweeks:


catenacyber:
Requires https://github.com/OISF/suricata/pull/5843

tests/bug-130/test.yaml
```diff
@@ -165,6 +165,8 @@ checks:
         psh: true
         state: closed
         syn: true
-        tcp_flags: 1b
-        tcp_flags_tc: 1b
-        tcp_flags_ts: 1b
+        tcp_flags: '1b'
```
catenacyber, line 168: Do we need to add simple quotes ?

rtkjweeks: technically not for the 1b, but for consistency I added it.
The 02 needed quotes, otherwise the test would fail because, I believe, it was being taken as a number and removing the initial 0, making it no longer match the actual output.

catenacyber: ok thanks yaml

PR #435, opened 2021-02-13
remove test as pcap is not good

inashivb:


victorjulien:
This commit message isn't clear, it should explain why.

inashivb:
Replaced with #445

PR #436, opened 2021-02-13
tests: Add test case for unreachable content

jlucovsky:
This PR
- Adds test cases for checking whether `distance` causes unreachable content
- Updates existing test cases to use the modified message.

This PR depends on suricata pr# (tbd)


jlucovsky:
Continued in #446

PR #434, opened 2021-02-11
parallel tests: windows fixup - v1

jasonish:
Just a small fixup over https://github.com/OISF/suricata-verify/pull/423 I did while updating my Windows build.

PR #439, opened 2021-02-16
parallel runs: initialize failedLogs on non-Linux

TharushiJay:
Fixes bug in #434 where failedLogs is not defined in non-Linux


jasonish:
Merged. Thanks.

I think a command line option `--force-single` would be good where it ran it through the non-mp runner even on Linux.  So it would follow the non-Linux code path if provided, instead of the Linux codepath.

PR #440, opened 2021-02-16
Validate eve.json output /v4

TharushiJay:

Changes from last PR : #431 

- If file runs outside the suricata-verify tree, print entire filename 
`./check-eve.py  -v /var/log/suricata`
- If file runs in the suricata-verify directory, chop off the common prefix
`./check-eve.py  -v`
- Add a check for an array in jsonschema
- After each test in run.py, execute check-eve.py with the tests output directory as the argument (error in the path to jsonschema)

### DIFF
--- a/check-eve.py
+++ b/check-eve.py
@@ -1,0 +1,115 @@
+#! /usr/bin/env python3
+#
+# Copyright (C) 2021 Open Information Security Foundation
+#
+# Permission is hereby granted, free of charge, to any person
+# obtaining a copy of this software and associated documentation files
+# (the "Software"), to deal in the Software without restriction,
+# including without limitation the rights to use, copy, modify, merge,
+# publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be
+# included in all copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+# EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+# MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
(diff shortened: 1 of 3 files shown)


run.py
```diff
@@ -872,6 +872,7 @@ def run_test(dirpath, args, cwd, suricata_config):
         elif results["success"] > 0:
             with lock:
                 count_dict["passed"] += 1
+        subprocess.call(["../suricata-verify/check-eve.py",outdir])
```
TharushiJay, line 875: Since we run the script after each test, it prints the pass fail count after each count.

check-eve.py
```diff
@@ -0,0 +1,115 @@
...
+
+    parser = argparse.ArgumentParser(description="Validation schema")
+    parser.add_argument("-v", dest="verbose", action="store_true")
+    parser.add_argument("file", nargs="?", default=[])
+    args = parser.parse_args()
+    TOPDIR = os.path.abspath(os.path.dirname(sys.argv[0]))
+    tdir = os.path.join(TOPDIR, "tests")
+
+    # ../suricata-verify/schema.json 
+    schema = json.load(open("schema.json"))
```
TharushiJay, line 75: To run this with the output dir from suricata directory, the path to json file needs to be in the format "../suricata-verify/schema.json". But then it won't work if you run the script from the inside of suricata directory.

run.py
jasonish: This doesn't seem to be working for me.  I updated the schema so all checks would fail, but it will passes.

A few ideas:
- We probably have to check the return value of this function.  If it fails, we'd throw a TestError with a reason which would be caught in the except block.
- We probably want to prevent any output when calling `check-eve.py` here.  A common command like option is `-q` and `--quiet` to suppress all output.  Then `check-eve` would exit with an error code..  It should exit with an error code if any checked json files fail anyways.

check-eve.py
jasonish: Should be able to use TOPDIR to resolve this I think.

jasonish (changes requested):
Some final comments here.. What I think we have here should be broken into 3 commits:
- The first one could be the commit of the schema.
- The second would be check-eve.py.
- And the third would be the integration of check-eve.py into run.py.

Thanks!

TharushiJay:
Changed continued in #461

PR #441, opened 2021-02-17
http: increase time limit for sticky start test

catenacyber:
So that Fedora CI does not fail for every PR


catenacyber:
Needs https://github.com/OISF/suricata/pull/5871

catenacyber:
force-pushed the rebased version

PR #447, opened 2021-02-19
Opt 3973 /v2

TharushiJay:
Some tests use a check.sh script to run "cmp" on 2 files to check that they are the same. This could be converted to an internal check and become part of test.yaml.

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/3973

Example test: output-tcp-data

Possible YAML syntax for test.yaml:
```
checks:
  - file-compare:
      filename: tcp-data.log
      expected: expected/tcp-data.log
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,11 @@ checks:
       # A shell check that also tests the output of the command.
       args: cat fast.log | wc -l | xargs
       expect: 1
+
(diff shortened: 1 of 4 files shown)


jasonish:
Merged. Thanks Tharushi!

PR #452, opened 2021-02-22
Http2 keywords v5

catenacyber:
Replaces #384 with rebase

CI should fail for version 6...
I will inspect the failure and do a new version

PR #453, opened 2021-02-23
Http2 keywords v6

catenacyber:
Replaces #452 with moving checks to other tests with same pcap, but different min-version
as the signatures cannot be conditioned on min-version by suricata-verify


jasonish:
Merged via: https://github.com/OISF/suricata-verify/pull/474

PR #455, opened 2021-02-23
Adds test about decoder against too many layers

catenacyber:
Needs Suricata PR to come


jasonish:
Note: GitHub says the pcap is an empty file.  However, on checkout it is not.

catenacyber:
Good you can get it

catenacyber:
Merged in commit d14fbbd3372732132a6e62a859b8e014b0f947da

PR #461, opened 2021-02-24
Validate eve.json output /v5

TharushiJay:

Changes from last PR : #440

- Added `--quiet `to suppress all output
- Resolved path errors
- Threw error in `run.py` if `check-eve.py` exited with error
- Broke commit into 3

### DIFF
--- a/run.py
+++ b/run.py
@@ -40,7 +40,7 @@ import unittest
 import multiprocessing as mp
 from collections import namedtuple
 import threading
-
+import subprocess
 import yaml
 
 WIN32 = sys.platform == "win32"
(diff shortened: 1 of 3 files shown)


run.py
```diff
@@ -872,6 +872,9 @@ def run_test(dirpath, args, cwd, suricata_config):
         elif results["success"] > 0:
             with lock:
                 count_dict["passed"] += 1
+        check_output = subprocess.call(["../suricata-verify/check-eve.py", outdir, "-q"])
```
jasonish, line 875: We need to use TOPDIR here to get the path correct.  I can suricata-verify like `../../suricata-verify.run.py` from my Suricata source directory, and the path is not found.

run.py
```diff
@@ -872,6 +872,9 @@ def run_test(dirpath, args, cwd, suricata_config):
...
+        if check_output == 0:
+            raise TestError("Invalid JSON schema")
```
jasonish, line 877: This should be `!= 0`.

jasonish (changes requested):
The output is still a little odd when the schema is updated to fail:
```
===> alert-distance-within-1: OK
===> alert-distance-within-1: FAILED: Invalid JSON schema
===> alert-no-3whs-established: OK
===> alert-no-3whs-established: FAILED: Invalid JSON schema
===> alert-testmyids: OK
===> alert-testmyids: FAILED: Invalid JSON schema
```
but lets focus on the other issues first.

TharushiJay:
Changes continued in #464

PR #469, opened 2021-03-01
ish - validate eve.json - v7.1

jasonish:
Based on https://github.com/OISF/suricata-verify/pull/467.

Add jsonschema go GitHub workflow so tests pass.

Rebased.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,103 @@
+{
+    "type": "object",
+    "properties": {
+        "timestamp": {
+            "type": "string",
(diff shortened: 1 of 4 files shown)


schema.json
```diff
@@ -0,0 +1,103 @@
+{
```
catenacyber, line 1: So, the point is to extend this schema, right ?

inashivb: I think so, yes. With all the remaining fields/types, I suppose.

PR #471, opened 2021-03-04
Pcap conditional

regit:
Fix a test and add some tests for the pcap conditional logging.


victorjulien:
Should only fail for master, the other branches won't get this feature.

tests/output-pcap-log-conditional-alert/check.sh
```diff
@@ -0,0 +1,3 @@
+#! /bin/sh
+
+exec cmp ${TEST_DIR}/expected/log.pcap.1444144603 log.pcap.0
```
victorjulien, line 3: this type of check is not cross platform. For example a big endian pcap with have a different header. So we'll need a bit smarter comparison logic here.

regit: Copied from https://github.com/OISF/suricata-verify/blob/master/tests/output-pcap-log/check.sh so we may need 2 fixes.

victorjulien: Yeah. Maybe we can compare `capinfos` output, although this would intro a new dependency. I guess another way would be to run the pcap through suri and inspect eve.json

regit: Running through suri may not work as we start capture in the middle of a app layer tx in most cases.

victorjulien: Guess it depends on the output you'd want. You'd be able to validate the flow record wrt 5 tuple and pkt counts? Also payloads in some way maybe, like actually using alerts for this... Open to other ideas

PR #465, opened 2021-02-28
ikev1: add ikev1 test - v8

satta:
Previous PR: https://github.com/OISF/suricata-verify/pull/351

* Only rebased so far.

PR #472, opened 2021-03-05
ikev1: add ikev1 tests - v9

satta:
Previous PR: https://github.com/OISF/suricata-verify/pull/465

Changes to previous version:

* Adjust version requirements for IKE changes. These were still annotated to be in version 6, yet will only be in version 7. This caused some of the tests with version 6 to fail.
* Add tests for transforms.

Tests for `ikev2-weak-dh` on the 7.x codebase (master) will succeed once IKEv1 support from the respective PR is merged.


victorjulien:
Merged in 2 steps: I accidentally merged #465, so I rebased #472 on top of it and merged the result as well.

satta:
Great, thanks anyway!

PR #313, opened 2020-08-31
stream/tcp: Test verifies the behavior when direction of TCP flow is … - v3

ilya-bakhtin:
…changed by the probing parser.

pcap file contains 2 http transactions. The request is missing for the first one.
The second transaction is fully complete. So eve.json must contain one and only anomaly event.

Also common flow details are verified.
It must be http, to port 80 with specified number of bytes_toclient and bytes_toserver

PR #473, opened 2021-03-10
stream/tcp: Test verifies the behavior when direction of TCP flow is …

catenacyber:
…changed by the probing parser.

pcap file contains 2 http transactions. The request is missing for the first one.
The second transaction is fully complete. So eve.json must contain one and only anomaly event.

Also common flow details are verified.
It must be http, to port 80 with specified number of bytes_toclient and bytes_toserver

Replacing https://github.com/OISF/suricata-verify/pull/313


jasonish:
Merged via https://github.com/OISF/suricata-verify/pull/474.

PR #374, opened 2020-12-05
threshold-config: New test for issue in 2190

jlucovsky:
This commit adds a test case for large IP address handling in the
threshold.config file.

PR #485, opened 2021-04-08
Next/20210408/v2

victorjulien:
#374 #484


catenacyber:
#374 needs back ports in 6

PR #486, opened 2021-04-15
modbus: add test cases

cccs-sadugas:
Add tests for modbus logging and alerting.


tests/modbus/test.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - filter:
+      count: 2
+      match:
+        event_type: modbus
+        modbus.request.function_code: RdCoils
+
+  - filter:
+      count: 2
+      match:
+        event_type: alert
```
catenacyber, line 24: Would it be possible to test also the `modbus` part of the alert ?

tests/modbus/suricata.yaml
```diff
@@ -0,0 +1,9 @@
+%YAML 1.1
+---
+
+outputs:
+  - eve-log:
+      enabled: true
+      types:
+        - modbus
```
catenacyber, line 8: Why do not we need to set `app-layer.protocols.modbus.enabled` ?

catenacyber (changes requested):
Looks quite good.
I would like to see the `modbus` part of one alert tested

tests/modbus/test.yaml
cccs-sadugas: Sure, we can look into this.

tests/modbus/suricata.yaml
cccs-sadugas: I think it's because it's enabled by default in [modbus.rs](https://github.com/OISF/suricata/pull/6090/files#diff-57a6cf267e4fb39aa835b56cc817f0c7dccbafc37de20f4c90eaf8eb216a5958R543):

```rust
// enable_default: 1
AppLayerRegisterProtocolDetection(&parser, 1);
```

cccs-sadugas:
Replaced by #499

PR #487, opened 2021-04-18
http-sha256-drop-02: don't use flow_id in test

regit:
flow_id has no reason to be fix from a user perspective so we
should not use it in test but use information from the flow
itself.


jasonish:
Merged without the bug-78 commit that did not pass.

PR #489, opened 2021-04-24
tests/vntag: Basic test case for VNTAG decode

jlucovsky:
This PR adds a basic test for VNTAG decoding.


catenacyber:
cf https://github.com/OISF/suricata/pull/6085

jlucovsky:
Continued in #500

PR #139, opened 2019-10-14
tests: add dcerpc test per #3109

travisbgreen:

PR #494, opened 2021-04-29

catenacyber:
Rebase of #139 + adds `midstream` in configuration to get TCP rule working
https://redmine.openinfosecfoundation.org/issues/3109


tests/dcerpc/dcerpc-3109/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+# disables checksum verification
+args:
+  - -k none --set stream.midstream=true
+
+checks:
+
+  - filter:
+      count: 1
+      match:
+        event_type: dcerpc
```
inashivb, line 11: It's smb w dcerpc so the event would never be dcerpc, dcerpc will be nested inside smb event

catenacyber: Will you do the changes ? Or should I ? (and the suricata fixes if any)

inashivb: I will. Thank you for this :)
I'm chasing the bug as it still persists

catenacyber: What should I do with this PR ? Is there a redmine ticket reference ?

inashivb: I think it's I who needs to do things.. Will check the tickets.. Got a report on similar lines from Brandon and possibly Peter. Will review them and give the update/create the tickets. 
Sorry. this fell off the radar completely!

PR #480, opened 2021-03-16
pcrexform: test becomes obsolete in version 7

catenacyber:
converted to a unit test failing to load the signature

PR #502, opened 2021-05-25
Pcre2 3194 v2

catenacyber:
Modifies #480 with adapting the threshold-config-validate-02 test to work for both PCRE1 and PCRE2

cc @jlucovsky for the threshold.config test change

see https://github.com/OISF/suricata/pull/6138

### DIFF
--- a/tests/detect-pcrexform-03/test.yaml
+++ b/tests/detect-pcrexform-03/test.yaml
@@ -1,4 +1,5 @@
 requires:
+  lt-version: 7
 
   files:
     - src/detect-transform-pcrexform.c
--- a/tests/threshold-config-validate-02/test.yaml
+++ b/tests/threshold-config-validate-02/test.yaml
@@ -7,7 +7,7 @@ command: |
 checks:
 
     - shell:
-        args: grep -e "pcre_exec parse error, ret -1, string this is not correct" suricata.log | wc -l | xargs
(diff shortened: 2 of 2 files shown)


victorjulien:
I don't understand why we disable the pcrexform3 test?

catenacyber:
> I don't understand why we disable the pcrexform3 test?

It has been moved to unit tests as `DetectTransformPcrexformParseTest03` 

It is just about testing the validity of one signature

catenacyber:
The real trick is that now we check that

> pcd->regex has exactly one capture expression

This is not the case in the signature in pcrexform3, so suricata fails on loading it

victorjulien:
I think the original test was not about testing sig parsing, but about validating runtime execution of the sig. It probably fails to parse on pcre2 because of the non-escaped `-`

catenacyber:
> I think the original test was not about testing sig parsing, but about validating runtime execution of the sig. 

I asked @jlucovsky about it on March 16th in Mattermost direct messages, and he confirmed me that this test is meant to test the signature's failure, and it it is better to fail early at parsing stage.

> It probably fails to parse on pcre2 because of the non-escaped `-`

No, it does not fail to parse on pcre2, ie pcre2 successfully parses it.
It fails on Suricata because we added a check (for the specific pcrexfrom keyword) to see if the regex has one capture expression.
Otherwise, `pcrexform` keyword does not make any sense because it has nothing to transform.

victorjulien:
Merged in #549

PR #503, opened 2021-05-26
Make HTTP2 test work on version 6

catenacyber:
PR to come

PR #506, opened 2021-05-31
Make HTTP2 test work on version 6

catenacyber:
Replaces #503 with using `app-layer.protocols.http2.http1-rules` for config


jasonish:
Merged. Thanks!

PR #509, opened 2021-06-04
Adds test for SMTP with RFC2231

catenacyber:
ie attachment filename split over multiple sections

see https://redmine.openinfosecfoundation.org/issues/4386


inashivb:
Thanks! Merged as a part of #596

catenacyber:
Oh I guess there should have been a min-version: 7 for this one

PR #511, opened 2021-06-26
mqtt: add suricata.yaml enabling MQTT for testing

satta:
This PR ensures that the MQTT app-layer parser is always enabled in the MQTT tests, See https://github.com/OISF/suricata/pull/6229#issuecomment-867926602

### DIFF
--- a/tests/mqtt-events-invalid-qos/suricata.yaml
+++ b/tests/mqtt-events-invalid-qos/suricata.yaml
@@ -1,0 +1,17 @@
+%YAML 1.1
+---
+
+outputs:
+  - eve-log:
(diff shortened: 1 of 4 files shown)


catenacyber:
You can also just add `--set app-layer.protocols.mqtt.enabled=yes` in test.yaml
What do you prefer @satta ?

satta:
I think I would prefer having separate `suricata.yaml` config files just for consistency with the other MQTT test cases.

catenacyber (approved):
Good for me

jasonish:
Merged. Thanks!

PR #513, opened 2021-06-28
Ensure the midstream flag is checked in midstream tests

rtkjweeks:


catenacyber:
cf https://github.com/OISF/suricata/pull/5930

PR #516, opened 2021-07-05
add tests for SCFlowTuple & SCPacketTuple

jufajardini:
Related to task: https://redmine.openinfosecfoundation.org/issues/4182 use lua_pushinteger for pushing integer types as integers instead of floats for SCPacketTuple and SCFlowTuple.

Add tests (separated, for now) for SCFlowTuple and SCPacketTuple. The test compares generated and expected output. PCAPs reused from test lua-output-http.

PR with patch: https://github.com/OISF/suricata/pull/6254


jasonish:
@jufajardini Can you tell me which Lua you used here? Mine logs `4.0` where your expected files log `4`.  I'm using Lua 5.4.

This might just show the importance of this fix in Suricata.

jasonish:
Something you can do is `ldd <your/suricata/binary>` and it will tell you which lua library Suricata is linked against.

We don't want the version in the log as we want to hopefully achieve the same behaviour on all Lua versions we can build against which these days is luajit, lua 5.1, lua 5.3+.

jasonish:
I'm also wondering if the change to Suricata is what makes it `4.0` and now its `4`?  Thats an observation worth noting.

jufajardini:
> I'm also wondering if the change to Suricata is what makes it `4.0` and now its `4`? Thats an observation worth noting.

That's interesting. Here, with liblua5.1 (my Suri will complain if I don't have exactly that one, it seems), no matter if I run the tests with master or with my changes, it always shows 4...

jufajardini:
> Something you can do is `ldd <your/suricata/binary>` and it will tell you which lua library Suricata is linked against.
> 
> We don't want the version in the log as we want to hopefully achieve the same behaviour on all Lua versions we can build against which these days is luajit, lua 5.1, lua 5.3+.

Thanks, I didn't know about that. Ok, removed the version from the log.

jufajardini:
> I'm also wondering if the change to Suricata is what makes it `4.0` and now its `4`? Thats an observation worth noting.

So, I re-run the test without my changes, now, using lua5.3, and it fails, because apparently with lua5.3 Suri will output 4.0 and not 4. If I use lua5.3 with my changes, the tests pass, though, so my understanding is that things are working as they should? I mean, the purpose of these tests is to ensure that we'll have integers for those values, right?

With lua5.1, my output is always `4`, though. With or without `lua_pushinteger`.

jasonish:
> So, I re-run the test without my patch, now, using lua5.3, and it fails, because apparently with lua5.3 Suri will output 4.0 and not 4. If I use lua5.3 with my changes, the tests pass, though, so my understanding is that things are working as they should? I mean, the purpose of these tests is to ensure that we'll have integers for those values, right?

Yes, I think the test is good after looking more closely.

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/522

PR #443, opened 2021-02-18
Http range file v2

catenacyber:
Test http file extraction over multiple transactions with unordered range headers.

PR #517, opened 2021-07-14
Http range file v3

catenacyber:
Modifies #443 with improving of last pcap to test more code paths in the PR, that is a range out of order which will have overlaps with previous ranges


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/531

PR #518, opened 2021-07-19
Adds test about icmpv6 type and code detection

catenacyber:

Base test case From #172 

From the paper, the first failing test by Suricata is `denial6-5`
cf `Table 4: IDSv6 Benchmark Results.`

I think the authors expected a signature present by default to trigger (ie some app-layer event)
But we can easily create the signature to detect as is shown in this Suricata-Verify test.

Also note that 51 Authentication Header is a known extension header even if it is not supported by the target (ie `2001:db8:1::1`)
and that attacker is in `2001:db8:1::2/128` rather than in `2001:db8:2::0/64`


catenacyber:
Added another commit in the same branch

catenacyber:
Maybe for the latest smurf, we can reduce the pcaps from the paper : getting 591234 alerts on ipv6-rsmurf

jasonish:
Should we worry about this? 

remote: warning: File tests/ipv6-evasion/ipv6-rsmurf/rsmurf6.pcap is 53.00 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB

catenacyber:
> Should we worry about this?
> 
> remote: warning: File tests/ipv6-evasion/ipv6-rsmurf/rsmurf6.pcap is 53.00 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB

Yes, we surely can take a smaller part of the pcap. Should I ?

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/534

PR #522, opened 2021-07-26
add tests for lua functions that use integer types

jufajardini:
Related to task: https://redmine.openinfosecfoundation.org/issues/4182 use lua_pushinteger for pushing integer types as integers instead of floats.

Change comparted to previous PR: 
- Add SV tests for SCFlowStats and SCRuleIds. The tests compare generated and expected output. PCAPs reused from test lua-output-http and alert-testmyids

Previous PR: https://github.com/OISF/suricata-verify/pull/516

Patch PR: https://github.com/OISF/suricata/pull/6279


tests/lua-scflowstats/check.sh
```diff
@@ -0,0 +1,3 @@
+#! /bin/sh
+
+exec cmp ${TEST_DIR}/output/lua-scflowstats.log ${TEST_DIR}/expected/lua-scflowstats.log
```
jasonish, line 3: Can you remove the `check.sh` scripts? We now have a `file-compare` that can be put in `test.yaml`, see README.md for an example.

jufajardini:
Sorry and thanks I hadn't realised this approach. Incorporated that in PR https://github.com/OISF/suricata-verify/pull/523

PR #528, opened 2021-09-03
tests: add test for bug 4663

victorjulien:


tests/bug-4663/blacklist
```diff
@@ -0,0 +1,5 @@
+cac5f74a1e146d6de26845e3e03732c01167e4bbb4333e76889b538c9bd30362
```
catenacyber, line 1: What is this blacklist file ?

PR #534, opened 2021-09-10
Ipv6 evasion 3323 v2

catenacyber:

Base test case From #172 

Note that 51 Authentication Header is a known extension header even if it is not supported by the target (ie `2001:db8:1::1`)
and that attacker is in `2001:db8:1::2/128` rather than in `2001:db8:2::0/64`

Modifies #518 with smaller pcap


jasonish:
Merged via https://github.com/OISF/suricata-verify/pull/547. Thanks!

PR #541, opened 2021-09-17
Adds test about HTTP2 disabled

catenacyber:
I do not know if we will want this for version 6


tests/http2-disabled/test.yaml
```diff
@@ -0,0 +1,28 @@
...
+args:
+  - -k none --set stream.midstream=true
+
+checks:
+
+  # Check that there is one file event with content range.
+  - filter:
+      count: 1
+      match:
+        event_type: http
```
victorjulien, line 16: can we add something here to show the alert happens on time (so at the moment of the non-switch)? I guess pcap_cnt or flow bytes/pkts might be available.

tests/http2-disabled/test.yaml
```diff
@@ -0,0 +1,28 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+  min-version: 7.0.0
```
inashivb, line 4: The Suricata fix's backport has been merged into master-6.0.x, should the min-version here be updated?

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/552

PR #543, opened 2021-09-20
Remove check.sh based tests.

jasonish:
Updates all tests using check.sh to use test.yaml instead. And removes support for check.sh.

Check.sh failure reporting was broken, and has been for a while:
https://github.com/OISF/suricata/pull/6391/checks?check_run_id=3640665242

I figure it was just better to kill of check.sh once and for all.


jasonish:
Merged via https://github.com/OISF/suricata-verify/pull/547.

PR #565, opened 2021-10-18
tls-fingerprint: test sticky buffer variant

jasonish:
Add another rule and filter to test the sticky buffer variant
of a tls fingerprint match. Also shows that the 2 are more or
less equivalent.


jasonish:
Merged.

PR #566, opened 2021-10-21
lua: adds test about ScFlowVarGet

catenacyber:
cf redmine 2094
https://redmine.openinfosecfoundation.org/issues/2094

cc @jufajardini


catenacyber:
@jufajardini what do you think of this PR ?

tests/lua-scflowvarget/test.yaml
```diff
@@ -0,0 +1,13 @@
...
+    - HAVE_LUA
+
+args:
+ - -k none
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: alert
```
jufajardini, line 13: I was thinking if we could add TestVar to the check here:
```suggestion
        event_type: alert
        metadata.flowvars[0].TestVar: "/zib100/zib100.json?origin=orf.at HTTP/1.1"
```

To make it more explicit/illustrative

jufajardini (commented):
Sorry for not reviewing it before, didn't realize I should. :x

I like it, I think it also does a good job as an example on how to approach such cases.

Did just one suggestion, with that in mind.

tests/lua-scflowvarget/test.yaml
catenacyber: Thanks

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/580

PR #575, opened 2021-10-22
Pcrexform submatch 4771 v4

catenacyber:
https://github.com/OISF/suricata/pull/6504#issuecomment-948945991

`cat expected/uri-param-seen.csv | while read i; do echo -n $i | base64 -D && echo; done`
gives
```
param2
param3
param
```

with the PR
and without
```
/test?param2
/test?param3
/test?param
```

(we get the whole match, not just the captured substring)

Replaces #573 with min-version 6 which introduced `pcrexform`


jasonish:
Merged. Thanks!

PR #576, opened 2021-10-26
eve: schema validation - v2

jasonish:
Previous PR: https://github.com/OISF/suricata-verify/pull/469

Changes from previous PR:
- Only run if `jsonschema` module is available.
- Some output tweaks.
- Add to github-ci.

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -1,0 +1,103 @@
+{
+    "type": "object",
+    "properties": {
+        "timestamp": {
+            "type": "string",
+            "pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}\\.\\d+[+\\-]\\d+$",
+            "optional": false
+        },
+        "flow_id": {
+            "type": "integer",
+            "optional": true
+        },
+        "pcap_cnt": {
+            "type": "integer",
+            "optional": true
+        },
+        "event_type": {
+            "type": "string",
+            "optional": false
+        },
+        "vlan": {
+            "type": "array",
+            "items": {
+                "type": "number"
+            }
+        },
+        "src_ip": {
+            "type": "string",
+            "optional": true
+        },
+        "src_port": {
+            "type": "integer",
+            "optional": true
+        },
+        "dest_ip": {
+            "type": "string",
+            "optional": true
+        },
+        "dest_port": {
+            "type": "integer",
+            "optional": true
+        },
+        "proto": {
+            "type": "string",
+            "optional": true
+        },
+        "http": {
+            "type": "object",
+            "optional": true,
+            "properties": {
+                "hostname": {
+                    "type": "string"
+                },
+                "url": {
+                    "type": "string"
+                },
+                "http_user_agent": {
+                    "type": "string"
+                },
+                "http_content_type": {
+                    "type": "string"
+                },
+                "http_method": {
+                    "type": "string"
+                },
+                "protocol": {
+                    "type": "string"
+                },
+                "status": {
+                    "type": "integer"
+                },
+                "length": {
+                    "type": "integer"
+                }
+            }
+        },
+        "app_proto": {
+            "type": "string",
+            "optional": true
+        },
+        "fileinfo": {
+            "type": "object",
+            "optional": true,
+            "properties": {
+                "filename": {
+                    "type": "string"
+                },
+                "state": {
+                    "type": "string"
+                },
+                "stored": {
+                    "type": "boolean"
+                },
+                "size": {
+                    "type": "integer"
+                },
+                "tx_id": {
+                    "type": "integer"
+                }
+            }
+        }
+    }
+}
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -83,6 +83,8 @@ jobs:
       - name: Running suricata-verify
         working-directory: suricata
         run: python3 ../run.py --quiet
+      - name: Running check-eve
+        run: python3 ./check-eve.py
 
   almalinux:
     name: AlmaLinux 8
@@ -129,6 +131,7 @@ jobs:
                 pcre2-devel \
                 pkgconfig \
                 python3-devel \
+                python3-jsonschema \
                 python3-sphinx \
                 python3-yaml \
                 rust-toolset \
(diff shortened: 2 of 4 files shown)


schema.json
```diff
@@ -0,0 +1,103 @@
...
+                },
+                "protocol": {
+                    "type": "string"
+                },
+                "status": {
+                    "type": "integer"
+                },
+                "length": {
+                    "type": "integer"
+                }
```
catenacyber, line 74: There is no test failure if there is one http log with a property which is not in this list ?
(like `range`)

jasonish: In JSONSchema unknown properties are allowed by default.  Consider this schema more as a starter that needs to be completed.  Once complete we could make unknown properties fail, which might not be a bad idea.  Forces schema updates for every new field logged.

catenacyber: How do we make JSONSchema to fail on unknown properties ?
Then, is there an easy way to complete the schema with these failures ?

jasonish: > How do we make JSONSchema to fail on unknown properties ? 

There is an `additionalProperties` rule we can set to false. http://json-schema.org/understanding-json-schema/reference/object.html#additional-properties

> Then, is there an easy way to complete the schema with these failures ?

Manually.

catenacyber: >> Then, is there an easy way to complete the schema with these failures ?
> Manually.

Challenge accepted ;-)

catenacyber: See https://github.com/catenacyber/suricata-verify/pull/new/github-ci-jsonschema/v3 for work in progress

json schema kinds of takes forever to process ipv6-dos-with-ext-headers-1 test case.
Its eve.json is 20 Mb
@jasonish should we exclude some test cases for this schema validation ?

catenacyber: Why did you pick `jsonschema` by the way ?
If some people at Suricon want to check their eve.json performance may be an issue, right ?
Are there alternatives ? Protobuf ?

jasonish: > @jasonish should we exclude some test cases for this schema validation ?

Could make sense.  Could be a flag set in `test.yaml` per test.. Default to enabled, but selectively could disable schema validation.

jasonish: > Why did you pick `jsonschema` by the way ? If some people at Suricon want to check their eve.json performance may be an issue, right ? Are there alternatives ? Protobuf ?

How would protobuf help validate correct generation of JSON?

This is about catching unexpected changes to existing JSON logs (when the schema is complete) or catching badly formatted JSON.  Maybe a flag to disable schema validation since its now always enabled if the dependencies are there.

catenacyber: > How would protobuf help validate correct generation of JSON?

I think protobuf will not do everything that jsonschema can do.
For instance, it will not to the regexp validation on the timestamp

But it could check the correctness of fields and their types by simply decoding the json.
And it would come with the feature to make it easy to reuse the logs as structures in many programming languages...
But I think jsonschema may have some useful features I do not know of...

Do you want me to try a POC ?

jasonish: I'm still a little confused.  Are you talking about logging with protobuf instead of JSON?

catenacyber: > I'm still a little confused. Are you talking about logging with protobuf instead of JSON?

Protobuf is a language to define a data structure.
It is then used to generate code in different languages (C++, Rust, python...).
The generated code comes with encoder/decoders to different formats including json (even if the most common format is the binary one I guess).

(It is a bit like ASN1 and DER)

Is this clearer ?

jasonish: Yeah, I'm familiar with Protobuf from years past, at least the binary format.  So I'm still not sure how it would help.  You'd create a protobuf schema, then it can validate the JSON using its tooling? Or would custom tooling be involved then.

catenacyber: > You'd create a protobuf schema, then it can validate the JSON using its tooling? Or would custom tooling be involved then.

You will need to have a 10 line program which tries to load the json into a structure and returns error if the loading failed...

You can extend this 10-line program if you want to check the timestamp is consistent with a regexp (as jsonschema does) as I do not think this feature is build in protobuf, or if you want other checks...

jasonish: Python program? Do we know it would be faster?

catenacyber: We can have a Rust program.
I think Rust will be faster than python ;-)

If we want to keep just python, I do not know if it will be faster indeed... (except by trying)

jasonish: Probably be better to try this first https://github.com/Stranger6667/jsonschema-rs, using protobuf as a JSON validator seemsl ike a hack when jsonschema is specifically built for this.

catenacyber: Looks a great idea, I will try it ;-)

jasonish:
So this is actually merged already, forgot to close this.

schema.json
catenacyber: Takes 8 seconds to validate ipv6-dos-with-ext-headers-1
When python is still working after 2 minutes...

catenacyber:
@jasonish to get jsonschema-rs to work, I needed to patch it so that it works with an iterator over the file (because the whole eve.json is not one single valid json)

That is instead of using `serde_json::from_str` on the whole eve.json, I use `serde_json::Deserializer::from_reader` and `deserializer.into_iter::<serde_json::Value>()`

So, do we 
- want to keep the python and exclude big eve.json ?
- propose a patch to jsonschema-rs to use their binary
- create out own binary using the jsonschema-rs library ?

jasonish:
Might want to ask before providing a patch.  It could be useful, but I could also see it being considered out of scope.

catenacyber:
> Might want to ask before providing a patch. It could be useful, but I could also see it being considered out of scope.

So, do you mean that "create out own binary using the jsonschema-rs library" is the right solution ?

jasonish:
> So, do you mean that "create out own binary using the jsonschema-rs library" is the right solution ?

Yeah, probably.  We know all the Rust build tools will be available as Suricata was already built.  We can then hide the building of the tool behind run.py so it would be pretty transparent.  Thats all assuming its faster.  Otherwise may as well stick with Python and exclude schema validation based on some factors.  But I'm sure it will be considerably faster.

catenacyber:
So, I did https://github.com/catenacyber/jsonschema-rs-iter.
Now I will change python scripts to use it and to a PR ;-)

catenacyber:
Continues in https://github.com/OISF/suricata-verify/pull/585

PR #585, opened 2021-11-22
GitHub ci jsonschema/v3

catenacyber:
cc @jasonish 

- extends the schema

I hosted myself the rust utility...
What do you think ?
Should it rather belong here in a subdirectory ?

check-eve has now an option to run either with rust or python (which is more verbose about errors, which is good for debugging)
What do you think ?

The schema should then be extended with turning `"additionalProperties": true` to false in some places

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -69,6 +69,11 @@ jobs:
                 zlib1g \
                 zlib1g-dev
       - run: cargo install --force --debug cbindgen
+      - run: git clone --depth 1 https://github.com/catenacyber/jsonschema-rs-iter.git
+      - name: Build jsonschema-rs-iter
+        working-directory: jsonschema-rs-iter
+        run: |
+          cargo install --path .
       - run: echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v2
       - run: python3 ./run.py --self-test
@@ -139,6 +144,11 @@ jobs:
                 which \
                 zlib-devel
       - run: cargo install --force --debug cbindgen
+      - run: git clone --depth 1 https://github.com/catenacyber/jsonschema-rs-iter.git
+      - name: Build jsonschema-rs-iter
+        working-directory: jsonschema-rs-iter
+        run: |
+          cargo install --path .
       - run: echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v2
       - run: python3 ./run.py --self-test
--- a/check-eve.py
+++ b/check-eve.py
@@ -27,10 +27,11 @@ import os
 import os.path
 import argparse
 import json
+import subprocess
 from jsonschema import validate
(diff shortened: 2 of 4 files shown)


catenacyber:
Should I have used 2018 rust ? instead of 2021 ?

jasonish:
> Should I have used 2018 rust ? instead of 2021 ?

Yes, and we should test that it compiles with Rust 1.48 as well.

.github/workflows/builds.yml
```diff
@@ -69,6 +69,11 @@ jobs:
                 zlib1g \
                 zlib1g-dev
       - run: cargo install --force --debug cbindgen
+      - run: git clone --depth 1 https://github.com/catenacyber/jsonschema-rs-iter.git
+      - name: Build jsonschema-rs-iter
+        working-directory: jsonschema-rs-iter
+        run: |
+          cargo install --path .
```
jasonish, line 76: You could `cargo install --git https://github.com/catenacyber/jsonschema-rs-iter`.

check-eve.py
```diff
@@ -63,6 +70,7 @@ def main():
 
     parser = argparse.ArgumentParser(description="Validation schema")
     parser.add_argument("-v", dest="verbose", action="store_true")
+    parser.add_argument("-r", dest="rust", action="store_true")
```
jasonish, line 73: Add a `help="Use Rust validator"` or something..

jasonish:
This is much faster, thanks.  Some thoughts on how this should look.
- I think the Rust validator should exist in this repo. One less hop if it needs to be modified.
- The default test runner should use the Rust mode, no need to expose the Python way to `run.py` as an option. This looks like the way it is now.
- I think `check-eve.py` should default to the Rust runner as well.  Select Python for the more verbose errors if needed.
- I also think we could run.py build/rebuild the Rust validator as needed on startup. It builds reasonably quick, and we could play with caching options in CI.

.github/workflows/builds.yml
catenacyber: Ok, but moved it to S-V repo

check-eve.py
catenacyber: ok

catenacyber:
> This is much faster, thanks. Some thoughts on how this should look.
> 

Indeed much faster

> * I think the Rust validator should exist in this repo. One less hop if it needs to be modified.

Done

> * The default test runner should use the Rust mode, no need to expose the Python way to `run.py` as an option. This looks like the way it is now.

`run.py` only uses the rust mode

> * I think `check-eve.py` should default to the Rust runner as well.  Select Python for the more verbose errors if needed.

Done

> * I also think we could run.py build/rebuild the Rust validator as needed on startup. It builds reasonably quick, and we could play with caching options in CI.

Why not using a prebuilt version ?

jasonish:
> Why not using a prebuilt version ?

Just one less thing to worry about, one less command to run to get the tests bootstrapped.

PR #586, opened 2021-11-23
GitHub ci jsonschema/v4

catenacyber:
cc @jasonish 

- extends the schema

The schema should then be extended with turning `"additionalProperties": true` to false in only one remaining place (the base object)

Replaces #585

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -59,6 +59,7 @@ jobs:
                 libjansson-dev \
                 libpython2.7 \
                 libnss3-dev \
+                libssl-dev \
                 make \
                 parallel \
                 python3-distutils \
@@ -71,6 +72,10 @@ jobs:
       - run: cargo install --force --debug cbindgen
       - run: echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v2
+      - name: Build jsonschema-rs-iter
+        working-directory: jsonschema-rs-iter
+        run: |
+          cargo install --path .
       - run: python3 ./run.py --self-test
       - run: git clone https://github.com/OISF/suricata -b ${{ matrix.branch }}
       - run: git clone https://github.com/OISF/libhtp suricata/libhtp
@@ -123,6 +128,7 @@ jobs:
                 libevent-devel \
                 libmaxminddb-devel \
                 libpcap-devel \
+                openssl-devel \
                 libtool \
                 lz4-devel \
                 make \
@@ -141,6 +147,10 @@ jobs:
       - run: cargo install --force --debug cbindgen
       - run: echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v2
+      - name: Build jsonschema-rs-iter
+        working-directory: jsonschema-rs-iter
+        run: |
+          cargo install --path .
       - run: python3 ./run.py --self-test
       - run: git clone https://github.com/OISF/suricata -b ${{ matrix.branch }}
       - run: git clone https://github.com/OISF/libhtp suricata/libhtp
--- a/check-eve.py
+++ b/check-eve.py
@@ -27,26 +27,29 @@ import os
 import os.path
 import argparse
 import json
+import subprocess
 from jsonschema import validate
 from jsonschema.exceptions import ValidationError
 
-def validate_json(args, dirpath, schema, isDirectory):
-    json_filename = dirpath
-    if isDirectory:
-        json_filename = os.path.join(dirpath, 'eve.json')
-        
+def validate_json(args, json_filename, schema):
     status = "OK"
     errors = []
 
-    with open(json_filename) as f:
-        for line in f:
-            obj = json.loads(line)
-            try:
-                validate(instance = obj, schema=schema)
-            except ValidationError as err:
-                status = "FAIL"
-                errors.append(err.message)
-    
+    if not args.python_validator:
+        cp = subprocess.run(["jsonschema-rs-iter", schema, "-i", json_filename])
+        if cp.returncode != 0:
+            status = "FAIL"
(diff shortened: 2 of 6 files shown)


catenacyber:
Should schema.json be part of Suricata rather than Suricata-verify in the end ?
When we modify Suricata to extend the logging, we will want to modify schema.json, right ?

catenacyber:
But then when the suricata PR gets merged, S-V will fail until its PR gets merged.
If schema.json is in suricata, S-V should stay green

catenacyber:
@jasonish do you know openssl name for Alma ? It is not `libssl-dev`

jasonish:
> @jasonish do you know openssl name for Alma ? It is not `libssl-dev`

Its `openssl-devel` as it should be ;)

jasonish:
> But then when the suricata PR gets merged, S-V will fail until its PR gets merged. If schema.json is in suricata, S-V should stay green

Good point.  I had never really thought of the schema being that strict.  I'm more interested in it failing when something that exists changes, rather than new stuff showing up.  Not that new stuff isn't as important, but something changing is usually a red flag.  It could probably go in `qa/` or `etc/`, any thoughts @victorjulien ?

catenacyber:
@satta doing this schema, I realize that for Mqtt, some common fields like `qos` are logged into message-specific (like `publish`) fields.
Could we make this generic ?

catenacyber:
> Its openssl-devel as it should be ;)

It looks to fail, or am I missing something @jasonish ?

catenacyber:
Stil todo : MQTT and base to be completed

catenacyber:
> Stil todo : MQTT and base to be completed

```
jsonschema-rs-iter schema.json -i tests/*/output/eve.json | grep -v VALID | sort | uniq -c | sort -n
   2 1. Additional properties are not allowed ('dcerpc' was unexpected)
   4 1. Additional properties are not allowed ('krb5' was unexpected)
   6 1. Additional properties are not allowed ('rfb' was unexpected)
   8 1. Additional properties are not allowed ('rdp' was unexpected)
  23 1. Additional properties are not allowed ('email' was unexpected)
  25 1. Additional properties are not allowed ('dhcp' was unexpected)
  43 1. Additional properties are not allowed ('http2' was unexpected)
  60 1. Additional properties are not allowed ('modbus' was unexpected)
  86 1. Additional properties are not allowed ('dnp3' was unexpected)
  89 1. Additional properties are not allowed ('ike' was unexpected)
```

+ lua

jasonish:
> > Its openssl-devel as it should be ;)
> 
> It looks to fail, or am I missing something @jasonish ?

Odd, when I copy and paste those commands into an almalinux docker container it works fine.  When I add it as the only change it also works: https://github.com/jasonish/suricata-verify/runs/4305413036?check_suite_focus=true#step:3:132

And I don't see any specials chars in there either.

satta:
> @satta doing this schema, I realize that for Mqtt, some common fields like `qos` are logged into message-specific (like `publish`) fields. Could we make this generic ?

In theory, yes, these could be moved outside the message-specific subobject, as these are indeed part of the fixed header that is common to all MQTT messages. However, in the current protocol versions they are only ever really used for `PUBLISH` (see https://docs.oasis-open.org/mqtt/mqtt/v5.0/os/mqtt-v5.0-os.html#_Toc3901023). I think my idea there was to not even log them for non-`PUBLISH` messages, but that must have been overlooked in the final code.

catenacyber:
> And I don't see any specials chars in there either.

But there was a special char for the space after the package's name, seen with my hex editor  ;-)
Copy/paste are bad for this :-/
Thanks for the tip

catenacyber:
@satta thanks for the explanation.
I think one benefit of this schema.json is to make us think about what we want in the output logs ;-)

catenacyber:
@jasonish this should pass now with all S-V tests.

How should we order this schema ?
I guess for each object, its properties should be order by type size : bool, integer, string, array, object, and then by name...
What is left to do here ?

Then, we can look into logs improvements such as the one with MQTT.
(for instance in DNP3, some fields are often logged in the root, but sometimes in `request` or `response`)

jasonish:
> How should we order this schema ?

How about just alphabetically?  The only thing I'd do is group all top-level non-object types together at the top, alphabetically, then do the object types (`dns`, `stats`, `http`, etc) alphabetically after the top level types.

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/588

Here is my dummy script to order the schema.json :
```
import json
from collections import OrderedDict

f = open("schema.json")
schema = json.load(f)

def reorder(s):
    if s["type"] != "object":
        return
    d1 = dict()
    d2 = dict()
    d3 = dict()
    for k in s["properties"]:
        if s["properties"][k]["type"] == "object":
            d1[k] = s["properties"][k]
        elif s["properties"][k]["type"] == "array":
            d2[k] = s["properties"][k]
        else:
            d3[k] = s["properties"][k]

    df = OrderedDict()
    for k, v in sorted(d3.items()):
        df[k] = v
    for k, v in sorted(d2.items()):
        df[k] = v
    for k, v in sorted(d1.items()):
        df[k] = v
    s["properties"] = df

def recurs(s):
    try:
        for k in s["properties"]:
            if s["properties"][k]["type"] == "array":
                recurs(s["properties"][k]["items"])
                reorder(s["properties"][k]["items"])
            elif s["properties"][k]["type"] == "object":
                recurs(s["properties"][k])
                reorder(s["properties"][k])
    except:
        pass

recurs(schema)
print(json.dumps(schema, indent=4))
```

victorjulien:
Is this for https://redmine.openinfosecfoundation.org/issues/1369 ?

catenacyber:
> Is this for https://redmine.openinfosecfoundation.org/issues/1369 ?

Indeed

PR #588, opened 2021-11-24
GitHub ci jsonschema/v6

catenacyber:
cc @jasonish 

- extends the schema to get all S-V tests passing 

Replaces #586

To me, it does not seem to make sense to run on suricata version < 7

Looks ready to me, what do you think ?

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -89,6 +89,7 @@ jobs:
         working-directory: suricata
         run: python3 ../run.py --quiet
       - name: Running check-eve
+        if: github.ref == 'refs/heads/master'
         run: python3 ./check-eve.py
 
   almalinux:
@@ -164,4 +165,5 @@ jobs:
         working-directory: suricata
         run: python3 ../run.py --quiet
       - name: Running check-eve
+        if: github.ref == 'refs/heads/master'
         run: python3 ./check-eve.py
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -59,6 +59,7 @@ jobs:
                 libjansson-dev \
                 libpython2.7 \
                 libnss3-dev \
+                libssl-dev \
                 make \
                 parallel \
                 python3-distutils \
@@ -71,6 +72,10 @@ jobs:
       - run: cargo install --force --debug cbindgen
       - run: echo "$HOME/.cargo/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v2
+      - name: Build jsonschema-rs-iter
+        working-directory: jsonschema-rs-iter
+        run: |
+          cargo install --path .
       - run: python3 ./run.py --self-test
       - run: git clone https://github.com/OISF/suricata -b ${{ matrix.branch }}
       - run: git clone https://github.com/OISF/libhtp suricata/libhtp
@@ -123,6 +128,7 @@ jobs:
                 libevent-devel \
                 libmaxminddb-devel \
                 libpcap-devel \
+                openssl-devel \
                 libtool \
                 lz4-devel \
                 make \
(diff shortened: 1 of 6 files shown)


catenacyber:
By the way, this schema looks like a good tool to improve :
For instance, do we all understand the base fields ?

For instance, I did not know about the `command`
Then looking ad `jq 'select(.command)' tests/*/output/eve.json`, it looks like it should belong to a `ftp-data` field instead of being directly at the root...
`filename` is the same

I think the suricata fix would be 
```
diff --git a/src/output-json-alert.c b/src/output-json-alert.c
index 0ebe6fe0c..99a16633f 100644
--- a/src/output-json-alert.c
+++ b/src/output-json-alert.c
@@ -519,7 +519,13 @@ static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb,
             }
             break;
         case ALPROTO_FTPDATA:
-            EveFTPDataAddMetadata(p->flow, jb);
+            jb_get_mark(jb, &mark);
+            jb_open_object(jb, "ftp-data");
+            if (EveFTPDataAddMetadata(p->flow, tx_id, jb)) {
+                jb_close(jb);
+            } else {
+                jb_restore_mark(jb, &mark);
+            }
             break;
         case ALPROTO_DNP3:
             AlertJsonDnp3(p->flow, tx_id, jb);
```

jasonish:
> By the way, this schema looks like a good tool to improve : For instance, do we all understand the base fields ?
> 
> For instance, I did not know about the `command` Then looking ad `jq 'select(.command)' tests/*/output/eve.json`, it looks like it should belong to a `ftp-data` field instead of being directly at the root... `filename` is the same
> 
> I think the suricata fix would be
> 
> ```
> diff --git a/src/output-json-alert.c b/src/output-json-alert.c
> index 0ebe6fe0c..99a16633f 100644
> --- a/src/output-json-alert.c
> +++ b/src/output-json-alert.c
> @@ -519,7 +519,13 @@ static void AlertAddAppLayer(const Packet *p, JsonBuilder *jb,
>              }
>              break;
>          case ALPROTO_FTPDATA:
> -            EveFTPDataAddMetadata(p->flow, jb);
> +            jb_get_mark(jb, &mark);
> +            jb_open_object(jb, "ftp-data");
> +            if (EveFTPDataAddMetadata(p->flow, tx_id, jb)) {
> +                jb_close(jb);
> +            } else {
> +                jb_restore_mark(jb, &mark);
> +            }
>              break;
>          case ALPROTO_DNP3:
>              AlertJsonDnp3(p->flow, tx_id, jb);
> ```

I'd be in favour of fixing some of these while they're being highlighted instead of adding them to the schema.  Root level fields that stand out to me are:
- command
- packet
- payload
- payload_printable
- spi (maybe)
- stream?

Or if it does make sense, but seems odd it looks like there is a `$comment` field we can use to add comments to a declaration.

catenacyber:
> I'd be in favour of fixing some of these while they're being highlighted instead of adding them to the schema.

Ok, I will take a look, make a suricata PR, and an update to this schema

catenacyber:
> I'd be in favour of fixing some of these while they're being highlighted instead of adding them to the schema. Root level fields that stand out to me are:
>

Agreed
 
> * command

Removing it to `ftp-data` subobject

> * packet
> * payload
> * payload_printable

These all make sense to me cf https://redmine.openinfosecfoundation.org/issues/3480
Where would you put these ?

> * stream?

stream indicates if the previous fields such as `packet` come from a stream (or real packet I guess)
It should be a boolean

They only happen on `event_type == alert`

So, would you have a subobject for them ?

> * spi (maybe)

It looks like kind of a port, but for ESP instead of TCP/UDP.
There is also the ICMP stuff which is in the root object.
So, I guess this one makes sense

> 
> Or if it does make sense, but seems odd it looks like there is a `$comment` field we can use to add comments to a declaration.

There is also `xff` which seems odd to me as it should belong to `http` (where it also is already)

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/590

catenacyber:
> > I'd be in favour of fixing some of these while they're being highlighted instead of adding them to the schema. Root level fields that stand out to me are:
> 
> Agreed
> 
> > * command
> 
> Removing it to `ftp-data` subobject
> 
> > * packet
> > * payload
> > * payload_printable
> 
> These all make sense to me cf https://redmine.openinfosecfoundation.org/issues/3480 Where would you put these ?
> 
> > * stream?
> 
> stream indicates if the previous fields such as `packet` come from a stream (or real packet I guess) It should be a boolean
> 

By the way, there is a `packet_info` object which is not used so much

catenacyber:
There is also `rpc` which should maybe be part of `nfs` ?

catenacyber:
After that, we can review objects which are not the root object

For instance, do we want `alert.source.ip` when we could have `alert.src_ip` (or even directly `.src_ip`) ?

PR #589, opened 2021-11-29
decode/event: Test Applayer engine events

jlucovsky:
Ticket: #4482

Suricata verify test that insures events created with
DecodeEngineSetEvent create anomaly records.

This test uses a SWF file manually modified with an invalid version to
trigger an event. The test verifies that the anomaly record was created.

This PR is referred to by [Suricata PR #6643](https://github.com/OISF/suricata/pull/6643)


tests/http-swf-engine-event/test.rules
```diff
@@ -0,0 +1 @@
+alert http any any -> any any (msg:"SWF Decode Engine Event Set"; file_data; content:"Any Conent"; sid:1;)
```
victorjulien, line 1: we don't test the matching of a decoder event here, which is the one of the goals of the PRs right?

victorjulien:
Do we need the full suricata.yaml? Can we either have a minimal one or a `--set ...` option to change the things that need to be modified?

tests/http-swf-engine-event/test.rules
jlucovsky: the event is checked in `test.yaml`

this rule causes the swf file to be inspected which then leads to the anomaly event (`test.yaml`).

victorjulien: The event logging is checked, but event rules are not tested.

catenacyber:
Closing as suricata PR was closed cf https://github.com/OISF/suricata/pull/6663

PR #590, opened 2021-11-30
GitHub ci jsonschema/v7

catenacyber:
cc @jasonish 

- extends the schema to get all S-V tests passing
- then removes some fields from schema.json root, so CI will fail without the upcoming suricata PR

Replaces #588

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -89,6 +89,7 @@ jobs:
         working-directory: suricata
         run: python3 ../run.py --quiet
       - name: Running check-eve
+        if: ${{ matrix.branch == 'master' }}
         run: python3 ./check-eve.py
 
   almalinux:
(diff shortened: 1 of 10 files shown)


.github/workflows/builds.yml
```diff
@@ -84,6 +89,7 @@ jobs:
         working-directory: suricata
         run: python3 ../run.py --quiet
       - name: Running check-eve
+        if: github.ref == 'refs/heads/master'
```
jasonish, line 92: Is this needed?

catenacyber: Yes because master-6.0.x has some additional fields (for dns if I remember correctly).
You can check the previous CI runs...
Or run it after running S-V tests with suricata 6.0.4

jasonish: But isn't this checking the ref of suricata-verify itself, not the ref of Suricata.

catenacyber: Indeed...
Force-pushing another version to test CI

catenacyber:
cf https://github.com/OISF/suricata/pull/6660

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/698

PR #593, opened 2021-12-02
Fixes mqtt unassgined test

catenacyber:
There is only one occurence of `mqtt.msgtype == 0`
So there should be only one alert

cc @satta 

needs https://github.com/OISF/suricata/pull/6662


satta:
Looks good.

catenacyber:
Should now pass on master and fail on master6 waiting for the backport

inashivb:
Thanks! Merged as a part of #596

PR #595, opened 2021-12-06
dataset-06-state-long: add test on learning dataset

regit:
Test that suricata can learn a set of 1200 DNS rrname and store
correctly the result on disk.


inashivb:
Does this go with any Suricata PR?

jasonish:
> Does this go with any Suricata PR?

No, its just increasing our dataset coverage.

jasonish:
This test is failing on GitHub CI, but passes for me locally.  I think its something about the environment and a shell check, and not really the fault of this test.  I'll do some checking myself as to why this type of check is failing.

jasonish:
Merged with change to deal with `sort` on different systems producing different results.

PR #602, opened 2021-12-09
Add test about enip keywords

catenacyber:


catenacyber:
The point of this is to increase fuzzing coverage cf https://storage.googleapis.com/oss-fuzz-coverage/suricata/reports/20211208/linux/src/suricata/src/detect-engine-enip.c.html

Fuzzing does not cover any line in detect-engine-enip.c

Fuzzing covers parsing of these enip keywords, and parsing of network traffic
But it did not find yet that it should combine them...
Not sure why.
Maybe we should have fuzz_sigpcap_aware targets for each app-layer protocol (with all the other app-layer protocols disabled)

victorjulien:
Tests pass is set but we have failures on 5 and 6?

catenacyber:
Failure on 6 is mqtt and smtp, so unrelated to this MR...
We should run CI again once fixes are merged

victorjulien:
Merged in #607

PR #605, opened 2021-12-12
This should cause a runtime exception.

jasonish:
But that exception does not show up as an error.

For demontration of issue https://redmine.openinfosecfoundation.org/issues/4899.


jasonish:
A fix for this case has now been merged.

PR #610, opened 2021-12-15
tests: test for security ticket 4710

victorjulien:


tests/security-4710-01/test.yaml
```diff
@@ -0,0 +1,23 @@
+args:
+- -k none
+
+checks:
+  - filter:
+      count: 2
+      match:
+        event_type: alert
```
catenacyber, line 8: This check seems a bit re

PR #611, opened 2021-12-16
Ssh banner test 4903 v1

TheKharleeci:


inashivb:
Hi @TheKharleeci !
I'd suggest reading the README of this project. It'll give you a better idea of what a minimal test must comprise of and how can you create tests.
In general, if we are testing something, we are usually matching it against an expected output. Like, if you're running A then B must be the output, if B is in fact the output then the test should pass else fail.
We cannot do anything with just the rules. We must check what would the rules be applied to and what is our expected output.

TheKharleeci:
> Hi @TheKharleeci ! I'd suggest reading the README of this project. It'll give you a better idea of what a minimal test must comprise of and how can you create tests. In general, if we are testing something, we are usually matching it against an expected output. Like, if you're running A then B must be the output, if B is in fact the output then the test should pass else fail. We cannot do anything with just the rules. We must check what would the rules be applied to and what is our expected output.

Ok. you mean adding the filters in the test.yaml to validate the output.

inashivb:
> > Hi @TheKharleeci ! I'd suggest reading the README of this project. It'll give you a better idea of what a minimal test must comprise of and how can you create tests. In general, if we are testing something, we are usually matching it against an expected output. Like, if you're running A then B must be the output, if B is in fact the output then the test should pass else fail. We cannot do anything with just the rules. We must check what would the rules be applied to and what is our expected output.
> 
> Ok. you mean adding the filters in the test.yaml to validate the output.

Yes, please. README shows a script called `createst.py` that will do everything for you.

inashivb:
Replaced w #618

PR #578, opened 2021-11-04
dcerpc-iface-01: update alert count to also match response

zer1t0:
Update to the dcerpc-iface-01 test as required by https://github.com/OISF/suricata/pull/6511 since this fix makes dcerpc responses to also match with dce_iface, which is the intended behaviour.

PR #571, opened 2021-10-22
dcerpc: dce_iface test to match many request/responses

zer1t0:
The dce_iface keyword should match for all those dcerpc requests and responses sent in the context of the given interface.

The following pcap includes a dcerpc connection that uses the netlogon interface (uuid: 12345678-1234-abcd-ef00-01234567cffb). After connecting to the interface, 6 requests/responses are made. The given rule should match 12 times, once for each request/response. 

This test is related to the ticket https://redmine.openinfosecfoundation.org/issues/4769 and PR https://github.com/OISF/suricata/pull/6511

PR #569, opened 2021-10-22
add smb dce_iface rule example to test it

zer1t0:
Added a test to match dce_iface keyword over SMB.

The dce_iface keyword should match for all those dcerpc requests and responses sent in the context of the given interface.

The pcap  contains a connection to the Windows printer service (uuid 12345678-1234-abcd-ef00-0123456789ab) with 2 requests/responses of the EnumPrinterDrivers method,  and 1 request/response of AddPrinterDriverEx. Therefore the rule should match 6 times in total, for all the requests and responses.

This test is related to the ticket https://redmine.openinfosecfoundation.org/issues/4767 and PR https://github.com/OISF/suricata/pull/6509

PR #568, opened 2021-10-22
add smb dce_opnum rules examples to test them

zer1t0:
Added a test to match dce_opnum keyword over SMB.

The dce_opnum keyword should match dcerpc requests and responses for a given opnum.

The pcap  contains a connection to the Windows printer service with 2 requests/responses of the EnumPrinterDrivers method (opnum 10), so first rule should match 4 times, and 1 request/response of AddPrinterDriverEx (opnum 89), so second rule should match 2 times.

This test is related to the ticket https://redmine.openinfosecfoundation.org/issues/4767 and PR https://github.com/OISF/suricata/pull/6510

PR #613, opened 2021-12-18
Dce fixes eloy/v1

inashivb:
#578 
#571 
#569 
#568


tests/dcerpc/dcerpc-dce-iface-many/test.rules
```diff
@@ -0,0 +1 @@
+alert dcerpc any any -> any any (msg: "DCE Netlogon"; dce_iface: 12345678-1234-abcd-ef00-01234567cffb; sid: 1;)
```
pevma, line 1: Not sure if it matters for the test but is it worth doing it with `dcerpc.iface` and `dcerpc.opnum` keywords (respectively below) instead as those are the new/sticky buffer stanza ?

inashivb: indeed. Thanks, Peter!

victorjulien:
This fails on all branches. Should it?

zer1t0:
Yes, it should only work after apply the fixes in https://github.com/OISF/suricata/pull/6725

inashivb:
> This fails on all branches. Should it?

I'll check the behavior of these parsers pre-rust, 6.0.x should require a fix.

zer1t0:
Yes, the branch 6.0.x required also the fixes. Here they were:
https://github.com/OISF/suricata/pull/6499
https://github.com/OISF/suricata/pull/6497

inashivb:
> Yes, the branch 6.0.x required also the fixes. The following PR introduces the fixes (new ones will be required without the C unittests): [OISF/suricata#6499](https://github.com/OISF/suricata/pull/6499) [OISF/suricata#6497](https://github.com/OISF/suricata/pull/6497)

Thank you very much, @zer1t0 ! I had overlooked these PRs, will use these to prepare a PR for 6.0.x.

inashivb:
Replaced w #620 

When https://github.com/OISF/suricata/pull/6725 gets merged, it will be backported to the `6.0.x` branch and these tests be updated.

PR #614, opened 2021-12-20
ssh banner: Add tests for ssh banner

TheKharleeci:
bug: 4903


tests/bug-4903-01/test.yaml
```diff
@@ -0,0 +1,66 @@
+# *** Add configuration here ***
+
```
jufajardini, line 2: The fix added by Shivani to this bug isn't yet backported to 5.0.x, that's why we are seeing the CI failures. In order for the test to be run for 6.0.x onwards, can you please add here the option:
```yaml
requires:
  min-version: 6.0
```
Same thing for the other test.yaml file :)

TheKharleeci: Ok. Thank you. I'll do that now.

jufajardini: Perfect, thanks! While you advance with your project and add more s-v tests,  when such failures happen, could you please reach out to us to confirm if this is a case for adding this configuration option or if's something that must be addressed?

TheKharleeci:
Followed by: #618

PR #616, opened 2021-12-21
runner: wrap file compare exceptions in TestError

jasonish:
If filecmp.cmp failed with an exception due to a missing file,
its exception was being lost due to another issue in the runner.

For this case it makes sense to catch this exception and wrap
it in a TestError exception as its most likely a missing file.


jasonish:
Will provide better output for https://github.com/OISF/suricata/pull/6723#issuecomment-998529461.

PR #619, opened 2021-12-23
Uricontent 4911 v1

TheKharleeci:


tests/uricontent-01/test.yaml
```diff
@@ -0,0 +1,14 @@
+# *** Add configuration here ***
+
+args:
+- -k none
```
victorjulien, line 4: since the pcap is just a single packet w/o the TCP 3 way handshake (SYN, SYN/ACK, ACK), Suricata will not parse this as HTTP unless you add:
`--set stream.midstream=true`
Then you should be able to add `checks` for an `alert` based on the rules in `test.rules` and a `http` event as well.

TheKharleeci:
Followed by: #621

PR #621, opened 2021-12-25
uricontent: add test for DetectUriSigTest02

TheKharleeci:
Previous PR: #619


tests/uricontent-01/test.yaml
```diff
@@ -0,0 +1,22 @@
+# *** Add configuration here ***
```
inashivb, line 1: nit: Please remove this line and the newline following this,

tests/uricontent-01/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+    match:
+      event_type: flow
+- filter:
+    count: 1
+    match:
+      event_type: stats
+- filter:
+    count: 1
+    match:
+      event_type: alert
```
inashivb, line 18: Please add the signature ID which should alert among the three rules provided.

tests/uricontent-01/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: flow
+- filter:
+    count: 1
+    match:
+      event_type: stats
```
inashivb, line 14: We probably don't need the `stats` event but it doesn't hurt without the checks for other params, I think.

inashivb (changes requested):
All in all, LGTM. :)
Some nits are requested from me.
Next steps:
1. Please check if any of the other tests in `detect-uricontent.c` do something differently. If yes, add tests corresponding to them too.
2. Make a folder for all these uricontent tests in the `tests` directory here.
3. Once all the convertible C tests have been converted to s-v, delete the corresponding tests from the C file and create a PR on https://github.com/OISF/suricata for that.

That should complete the process for this one file.

TheKharleeci:
Followed By: #624

PR #622, opened 2021-12-27
run.py: raise silently ignored exceptions

inashivb:
[Ticket 4899](https://redmine.openinfosecfoundation.org/issues/4899)

Demo in PR #605


inashivb:
Replaced w #623

PR #623, opened 2021-12-27
run.py: raise silently ignored exceptions

inashivb:
[Ticket 4899](https://redmine.openinfosecfoundation.org/issues/4899)

Demo in PR #605


jasonish:
This appers to disable the parallelism.  While poking around to see why (I think its the res.get()) I found that this seems to be a solution to this issue:

https://github.com/OISF/suricata-verify/commit/f94551cf129e208df8fc6827cf0ba32effac407d

jasonish:
Thanks Shivani.. I've gone with an alternate fix for now.

PR #624, opened 2021-12-30
Uricontent 4911 v3

TheKharleeci:
Previous PR: #621


tests/uricontent-05/test.rules
```diff
@@ -0,0 +1,3 @@
+alert tcp any any -> any any (msg:"Test uricontent"; content:"foo"; http_uri; content:"bar"; sid:1;)
+alert tcp any any -> any any (msg:"Test uricontent"; content:"one"; http_uri; offset:1; depth:10; content:"one"; offset:1; depth:10; content:"two"; http_uri; distance:3; within: 4; content:"two"; distance:1; within: 4; content:"three"; http_uri; distance:1; within: 6; content:"/three"; distance:0; within: 7; sid:2;)
+alert tcp any any -> any any (msg:"Test uricontent"; content:"one"; http_uri; offset:1; depth:10; content:"two"; http_uri; distance:1; within: 4; content:"six"; http_uri; distance:1; within: 6; sid:3;)
```
inashivb, line 3: Can we change the rule or packet content here somehow to make it alert?

tests/uricontent-06/test.rules
```diff
@@ -0,0 +1,2 @@
+alert udp any any -> any any (msg:"test"; content:"|af|af|af|"; http_uri; sid:238012;)
+alert tcp any any -> any any (msg:"test"; content:"boo"; http_uri; sid:1;)
```
inashivb, line 2: Can we change the packet content or the rule here to make it alert?

tests/uricontent-06/test.yaml
```diff
@@ -0,0 +1,18 @@
+# *** Add configuration here ***
+
```
inashivb, line 2: nit: Please remove this.

inashivb (commented):
Hi @TheKharleeci !
Mostly, LGTM. :) I have a few questions inline that can probably help us make the test better. Please follow the same for any tests from C that may not have yet been covered here.

jufajardini (commented):
Good job! On top of Shivani's comments:
- can you please squash the two commits?
- I think all these tests could be put into a directory called `uricontent`, so the path to each test would be smth like `test/uricontent/uricontent-0x`
- for the next tests, I think it could also be useful to add to the PR description which S-V test covers which c-tests, so it's easier to keep track of changes and of what has already been covered or not. :)

TheKharleeci:
Followed By: #625

PR #636, opened 2022-01-13
detect-modbus/v1: add tests to modbus

TheKharleeci:
Tests included: 
DetectModbusTestAccess
DetectModbusTestFunction

### DIFF
--- a/tests/modbus/test.rules
+++ b/tests/modbus/test.rules
@@ -1,3 +1,5 @@
+alert modbus any any -> any any (msg:"Testing modbus code function"; modbus: access write; sid:6;)
+alert modbus any any -> any any (msg:"Testing modbus code function"; modbus: function 8; sid:7;)
 alert modbus any any -> any any (msg:"Modbus function number test"; modbus: function 1; sid:1; rev:1;)
 alert modbus any any -> any any (msg:"Modbus function word test"; modbus: function assigned; sid:2; rev:1;)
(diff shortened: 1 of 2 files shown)


tests/modbus/test.rules
```diff
@@ -1,3 +1,5 @@
+alert modbus any any -> any any (msg:"Testing modbus code function"; modbus: access write; sid:6;)
```
jufajardini, line 1: This is maybe nit, but according to the documentation, this rule message should have a message like: `Testing modbus access` or something like that, I think...

cf https://suricata.readthedocs.io/en/latest/rules/modbus-keyword.html?highlight=modbus#modbus-keyword

tests/modbus/test.rules
```diff
@@ -1,3 +1,5 @@
...
+alert modbus any any -> any any (msg:"Testing modbus code function"; modbus: function 8; sid:7;)
```
jufajardini, line 2: Another nit: can we have the new signatures be put _after_ the ones that were already there and follow `sid` order? :)

jufajardini (commented):
Looking good!

Two small notes in the comments.

I'm wondering if it wouldn't be good to use matches like we have for `sid` 4, with ` modbus.request.function_code:` as s-v tests also serve as use-cases for other folks learning about Suri. Let's see if the others think this would be a valid addition :)

tests/modbus/test.rules
TheKharleeci: Ok I'll make that change

TheKharleeci:
Followed By: #648

PR #649, opened 2022-01-15
detect-proto/v1: add tests for detect-engine-proto

TheKharleeci:
Adds test for DetectProtoTestSig01


tests/detect-engine-proto/test.rules
```diff
@@ -0,0 +1,3 @@
+alert udp any any -> any any (msg:"Not tcp"; flow:to_server; sid:1;)
```
jufajardini, line 1: I think that it's important to add a filter to ensure we get zero matches for this rule, otherwise it seems lost in this test...

jufajardini (changes requested):
Looks almost ready to me!

One change requested in the comments.

Got wondering if we should add a rule to match on http as well, since we have that traffic in the pcap?

tests/detect-engine-proto/test.rules
inashivb: Agreed. Let's try to target all `sid`s that we use.

inashivb (changes requested):

tests/detect-engine-proto/test.rules
TheKharleeci: Ok. Should I go-ahead to add the rule for HTTP?

inashivb: I'm not sure I understand the need for http rule, could you please explain?
Here in the above comments, we are suggesting to add a filter for sid:1 making sure alert count stays 0.

jufajardini: I think the comment is because of this "Got wondering if we should add a rule to match on http as well, since we have that traffic in the pcap?", @inashivb. :P

inashivb: oh! missed that. Sure. :)

TheKharleeci:
Followed By:  #670

PR #650, opened 2022-01-16
detect-bypass/v2: add tests for detect-bypass

TheKharleeci:
Adds test for DetectBypassTestSig01

Previous PR: #645


TheKharleeci:
I am not sure if this is valid. From the Suricata test, it seems the first packet should generate 0 alerts for the first rule.

tests/detect-bypass/writepcap.py
```diff
@@ -0,0 +1,13 @@
...
+from scapy.all import *
+
+pkts = []
+
+pkts += Ether(dst='ff:ff:ff:ff:ff:ff', src='00:01:02:03:04:05')/ \
+    Dot1Q(vlan=6)/ \
+    IP(dst='1.2.3.4', src='5.6.7.8')/TCP(sport=6666, dport=80, flags='P''A')/"GET /index.html HTTP/1.0\r\nHost: This is dummy message body\r\nUser-Agent: www.openinfosecfoundation.org\r\nContent-Type: text/html\r\n\r\n"
+pkts += Ether(dst='ff:ff:ff:ff:ff:ff', src='00:01:02:03:04:05')/ \
+    Dot1Q(vlan=6)/ \
+    IP(dst='1.2.3.4', src='5.6.7.8')/TCP(sport=6060, dport=53, flags='P''A')/"HTTP/1.0 200 ok\r\nContent-Type: text/html\r\nContent-Length: 7\r\n\r\nmessage"
```
jufajardini, line 11: From the unittest, I think this one should be a packet to the client, not to the server again.

tests/detect-bypass/writepcap.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python
+from scapy.all import *
+
+pkts = []
+
+pkts += Ether(dst='ff:ff:ff:ff:ff:ff', src='00:01:02:03:04:05')/ \
+    Dot1Q(vlan=6)/ \
+    IP(dst='1.2.3.4', src='5.6.7.8')/TCP(sport=6666, dport=80, flags='P''A')/"GET /index.html HTTP/1.0\r\nHost: This is dummy message body\r\nUser-Agent: www.openinfosecfoundation.org\r\nContent-Type: text/html\r\n\r\n"
```
jufajardini, line 8: very nit, but I think we're missing an `a`: This is _a_ dummy message body

tests/detect-bypass/test.rules
```diff
@@ -0,0 +1,3 @@
+alert tcp any any -> any any (bypass; content:"GET "; sid:1;)
+alert http any any -> any any (bypass; content:"message"; http_server_body; sid:2;)
+alert http any any -> any any (bypass; content:"message"; http_host; sid:3;)
```
jufajardini, line 3: Since this is a new test, should we update the rule to use the new keywords `http.response_body` and `http.host`?

tests/detect-bypass/test.yaml
```diff
@@ -0,0 +1,22 @@
+args:
+- --set stream.midstream=true
+
+checks:
+- filter:
+    count: 1
```
jufajardini, line 6: Answering your comment:
indeed, my understanding from the test is also that sid 1 should not alert.
And that we want sid 3 to alert.

tests/detect-bypass/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+      alert.signature_id: 1
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 2
+- filter:
+    count: 2
+    match:
+      event_type: flow
```
jufajardini, line 18: With the change in direction for packet 2 (mentioned in another comment), this should also change, as this is supposed to be a single flow.

jufajardini (changes requested):
Answered your comment and added a few others :)

tests/detect-bypass/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+- --set stream.midstream=true
+
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 1
+- filter:
+    count: 1
```
jufajardini, line 11: The CI failures indicate that we are not getting this match here, so do check what's going on, please! 🤔

tests/detect-bypass/test.rules
inashivb: indeed. @TheKharleeci we changed these keywords to dot notation a while ago, you can find the table of legacy content modifiers/sticky buffers and their corresponding new keyword here: https://suricata.readthedocs.io/en/latest/rules/http-keywords.html#http-keywords

tests/detect-bypass/writepcap.py
inashivb, line 11: In case you didn't know., Scapy has inbuilt support for creating HTTP requests and response in a much legible way: `HTTPResponse()` Read https://scapy.readthedocs.io/en/latest/layers/http.html and set fields that you want to set in there. ;)
What you're doing i fine too.

inashivb: indeed

tests/detect-bypass/test.yaml
inashivb: I think this is also due the flow directions.. should be fine once that is fixed.

tests/detect-bypass/test.yaml
inashivb: indeed

tests/detect-bypass/test.rules
```diff
@@ -0,0 +1,3 @@
+alert tcp any any -> any any (bypass; content:"GET "; sid:1;)
```
inashivb, line 1: hmm not sure what I'm missing but seems like a valid match to me, don't know the reason why it fails in C..

TheKharleeci:
Followed by: #682

PR #651, opened 2022-01-16
detect-flowbits/v1: add tests for detect-flowbits

TheKharleeci:
Adds test for FlowBitsTestSig9-11


tests/detect-flowbits/test.rules
```diff
@@ -0,0 +1,7 @@
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,fb1; sid:1;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,fb2; sid:2;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,fb3; sid:3;)
+alert ip any any -> any any (msg:"Flowbit isset ored flowbits"; flowbits:isset,fb3|fb4; sid:4;)
+alert ip any any -> any any (msg:"Flowbit isnotset ored flowbits"; flowbits:isnotset, fb1 | fb2 ; sid:5;)
+alert ip any any -> any any (msg:"Flowbit unset"; flowbits:unset,myflow2; sid:6;)
+alert ip any any -> any any (msg:"Flowbit unset"; flowbits:toggle,myflow2; sid:7;)
```
jufajardini, line 7: The result of this rule varies based on what was the previous state of the flowbits variable, so I think we'll have to decide first on what we want here, so we can settle on the test checks and on the rule message...
(acc to https://suricata.readthedocs.io/en/latest/rules/flow-keywords.html?#flowbits)

tests/detect-flowbits/test.yaml
```diff
@@ -0,0 +1,16 @@
+args:
+- --set stream.midstream=true
+
+checks:
+- filter:
+    count: 13
+    match:
+      event_type: alert
```
jufajardini, line 8: I _think_ we need more details here, as we have many rules and in the end it's not very clear which ones are matching, nor how. Some input (based on the unittests) to help deciding on that, hopefully...

Test06, Test07 and Test08:
not sure how to go about these... Maybe if we used also `noalert` to the first `flowbists:set`, then added a `content` keyword for the rule with `flowbits:toggle` and made sure the test expected that one to match for an alert? 
Test09:
expects matches from 1 and 2, but no match from 4
Test11:
expects matches from 1, 2, 3 but no match for 5
Test10:
expects matches from 1, 2, 3 and 4

jufajardini (changes requested):
This one needs input from the other mentors, as it seems that some of the unittests require more detailed checks or more detailed rules for them to be covered by this s-v test. I've left a few thoughts in the comments, to explain why do I think that.

jasonish:
I agree with @jufajardini, more detail needs to be in the match tests.  For example, to test `FlowBitsTestSig09`, I'd want to verify that sid 1 and 2 triggered, but that sid 4 (sid 3 from `FlowBitsTestSig09`) did not.

Then for `FlowBitsTestSig09` I'd want to verify that sid 1, 2, 3 and 4 fired.

As these both share the same pcap, I think they can all live in the test S-V test, but you may have to create more rules.

3 rules, sids 1-3 using a common prefix for the flow bits to test `Sig09`...

4 rules, sids 4-7, using a different flowbit prefix to test `Sig09` and so on.

Keeping in mind you can only test about 15 alerts per packet, but that should give you enough to replicate unit tests 9-11 here.

TheKharleeci:
Followed By: #696

PR #660, opened 2022-01-17
detect-hostbits/v1: convert unittests

TheKharleeci:
Adds test for:
- Detect-hostbits-01: HostBitsTestSig05
- Detect-hostbits-01: HostBitsTestSig07
- Detect-hostbits-01: HostBitsTestSig08


tests/detect-hostbits/detect-hostbits-01/test.yaml
```diff
@@ -0,0 +1,17 @@
+args:
+- -k none
+
+checks:
```
victorjulien, line 4: we should probably also add checks for sid 1 and sid 2 not generating alerts, so `count: 0`

tests/detect-hostbits/detect-hostbits-02/test.yaml
```diff
@@ -0,0 +1,32 @@
+args:
+- -k none
+
+checks:
```
victorjulien, line 4: can you add a filter for sid 1 here as well?

tests/detect-hostbits/detect-hostbits-03/test.yaml
```diff
@@ -0,0 +1,27 @@
+args:
+- -k none
+
+checks:
+- filter:
```
victorjulien, line 5: sid 4 should get a check

victorjulien (changes requested):
see comments

TheKharleeci:
Followed by: #666

PR #662, opened 2022-01-18
ssh-banner-only: update test.yaml checks - v1

jufajardini:
While studying this test in order to review a PR, I've noticed that there were two rules, but only one (sid 1) being checked for in the `test.yaml`.  

Describe changes:
- Added filter for rule sid 3 which had done
- Uncommented rule sid 2 (will see if this triggers failures, and might rollback to leaving it commented out)

### DIFF
--- a/tests/ssh-banner-only/test.rules
+++ b/tests/ssh-banner-only/test.rules
@@ -1,4 +1,4 @@
 alert ssh any any -> any any (ssh.software; content:"OpenSSH"; sid:1;)
 # broken?
-#alert ssh any any -> any any (ssh.softwareversion:OpenSSH_7.4; sid:2;)
+alert ssh any any -> any any (ssh.softwareversion:OpenSSH_7.4; sid:2;)
 alert ssh any any -> any any (ssh.proto; content:"2"; sid:3;)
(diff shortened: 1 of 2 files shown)


tests/ssh-banner-only/test.rules
```diff
@@ -1,4 +1,4 @@
 alert ssh any any -> any any (ssh.software; content:"OpenSSH"; sid:1;)
 # broken?
```
inashivb, line 2: what was supposed to be broken here @catenacyber ?

jufajardini: oops, I forgot to remove that line :x

catenacyber: I did not do that test...
I guess we can be happy with `ssh.software`
There is one old keyword, and a new one for the sticky buffer thing if I remember correctly...

jufajardini: Now I'm wondering if this rule was left there to check if the deprecated keyword still worked...

victorjulien (changes requested):
Needs minor cleanup: see inline comments

jufajardini:
Cleaned up with https://github.com/OISF/suricata-verify/pull/709

PR #641, opened 2022-01-13
Frames/v20

victorjulien:
replaces #631

PR #664, opened 2022-01-19
Test frames - v21

jasonish:
Takes https://github.com/OISF/suricata-verify/pull/641 and splits the tests so tests pass on Suricata 5 and 6 as well.

Includes runner change to only set `ASAN_OPTIONS` if not already set in the environment. This allows the user to override this variable if needed.


inashivb:
Merged in #671, thanks!

PR #674, opened 2022-01-22
nfs3: add test for readdirplus records v2

ghostinthehive:
Previous PR: #657 
S-V for unittests at https://github.com/OISF/suricata/pull/6779/commits/4e2edd44aa08f08286001701ec0860c6638625d5

PR #675, opened 2022-01-24
nfs3: add test for readdirplus records v3

ghostinthehive:
- Typo Fix :)

Previous PR: #674 
S-V for unittests at OISF/suricata@4e2edd4


tests/nfs3-readdirplus/test.yaml
```diff
@@ -0,0 +1,33 @@
...
+args:
+- -k none
+- --set stream.midstream=true
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: nfs
+        dest_ip: 192.168.1.6
```
inashivb, line 16: I see src/dst ip interchanged in 5.. I remember some work done about proper tx logging.. Maybe that's related. Perhaps we should add a min version check for 6? @victorjulien ?

tests/nfs3-readdirplus/suricata.yaml
```diff
@@ -0,0 +1,14 @@
...
+      enabled: yes
+      filetype: regular
+      filename: eve.json
+      types:
+        - nfs
+
+app-layer:
+  protocols:
+    nfs:
+      enabled: yes
```
inashivb, line 14: Was the `suricata.yaml` added to enable `nfs` across all versions? Just to let you know, you could also achieve this by adding `--set app-layer.protocols.nfs.enabled=yes` in `test.yaml` `args` ;)

ghostinthehive: I think that's a better option since there is really not much enabled for having a custom `suricata.yaml`
thanks for the suggestion!

tests/nfs3-readdirplus/test.yaml
victorjulien: yeah lets go with min version 6

victorjulien (changes requested):
See inline comments from Shivani and myself

ghostinthehive:
Followed by #683

PR #676, opened 2022-01-24
Adds test with dns/enip polyglot flow

catenacyber:
Not sure if we want to merge this

cf https://github.com/OISF/suricata/pull/6832


victorjulien:
@jasonish can you see if we can improve DNS detection

jasonish:
> @jasonish can you see if we can improve DNS detection

Is this a real world scenario to consider? Both ENIP and DNS parsers are pinned to different ports which should make this a non-issue I'd think.

jasonish:
So this test is showing something that is odd, and not what should be?

catenacyber:
> > @jasonish can you see if we can improve DNS detection
> 
> Is this a real world scenario to consider? Both ENIP and DNS parsers are pinned to different ports which should make this a non-issue I'd think.

ENIP uses source ports, so you can have source port ENIP and destination port DNS and both probing parsers are run 

Cf Suricata.yaml
```
    enip:
      enabled: no
      detection-ports:
        dp: 44818
        sp: 44818
```

> So this test is showing something that is odd, and not what should be?

I am not sure what should be indeed...

catenacyber:
Replaced by Gitlab MR by @jasonish

PR #629, opened 2022-01-07
test for max alert 15 issue

inashivb:
Ticket: https://redmine.openinfosecfoundation.org/issues/4942

PR #686, opened 2022-01-27
Add tests for packet_alert_max - v1

jufajardini:
Build upon [no-alert PR #629](https://github.com/OISF/suricata-verify/pull/629) by @inashivb, add test for when we do want to have more than 15 alerts to make sure that works.

Related to https://redmine.openinfosecfoundation.org/issues/4207
and https://redmine.openinfosecfoundation.org/issues/4942


jufajardini:
Considering the failures, should the test for packet-max-alert above 15 require min-version 7.0?

inashivb:
> Considering the failures, should the test for packet-max-alert above 15 require min-version 7.0?

Yes, since you're just changing this behavior.

tests/alert-max/alert-max-15/suricata.yaml
```diff
@@ -0,0 +1,15 @@
...
+  - eve-log:
+      enabled: yes
+      filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
+      filename: eve.json
+      types:
+        - alert:
+
+# Define maximun number of possible alerts that can be triggered for the same
+# packet. Default is 15
+packet-alert-max: 15
```
inashivb, line 14: Perhaps in order to check the default behavior, we should not read it from the yaml .. (whatever the default is supposed to be now ,15?)

tests/alert-max/alert-max-20/suricata.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+outputs:
+  - eve-log:
+      enabled: yes
+      filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
+      filename: eve.json
+      types:
+        - alert:
+
+# Define maximun number of possible alerts that can be triggered for the same
```
inashivb, line 12: nit: typo `maximun`

jufajardini: :D:

tests/alert-max/alert-max-15/suricata.yaml
jufajardini: I see your point! I thought that if no suricata.yaml was provided, s-v would use whatever the person has locally, and if they had changed the value in theirs, this test would fail. Is this assumption wrong?

inashivb: Indeed, good thinking. In that case, default test should probably provide a `suricata.yaml` w/o this entry.
I think we should not try to fix local setups but what we would expect Suri to do with latest `master` and other branches in a default setup. If `packet-alert-max` is going to be enabled by default and have a value set in `suricata.yaml`, this test for fallback to defaults should have the other way around (i.e. no such value in yaml).
Also: we should probably change the name of this test, `alert-max-default` in such a case or something?

Could use a senior's review on my suggestion here.

jufajardini: > Indeed, good thinking. In that case, default test should probably provide a `suricata.yaml` w/o this entry. I think we should not try to fix local setups but what we would expect Suri to do with latest `master` and other branches in a default setup. If `packet-alert-max` is going to be enabled by default and have a value set in `suricata.yaml`, this test for fallback to defaults should have the other way around (i.e. no such value in yaml). Also: we should probably change the name of this test, `alert-max-default` in such a case or something?
> 
> Could use a senior's review on my suggestion here.

I think you're totally right there. I think I'll submit a PR with said changes, and then if that's not wanted, changes would be requested :P

jufajardini:
Replaced by https://github.com/OISF/suricata-verify/pull/694 incorporating feedback & discussions.

PR #687, opened 2022-01-27
tests/classification: Tests to exercise classification.config validation

jlucovsky:
This PR 
- Adds 2 test cases for classification.config to ensure legacy behavior when not in test mode and that an invalid classification.config file raises an error when in test mode
- Adds a classification.config file for an extant test using test mode so the test can check/validate empty rules and not stop on an empty classification.config file.

[Suricata PR 6878](https://github.com/OISF/suricata/pull/6878)


jlucovsky:
Continued in #712

PR #688, opened 2022-01-27
detect/lua/v1: add tests for detect-lua

TheKharleeci:
Includes test for LuaMatchTest01


tests/detect-lua-01/test.yaml
```diff
@@ -0,0 +1,21 @@
...
+      event_type: flow
+- filter:
+    count: 1
+    match:
+      event_type: stats
+- filter:
+    count: 2
+    match:
+      event_type: alert
+      alert.signature_id: 6677001
```
jufajardini, line 21: Considering there are two other rules added to the test, can you add filters for those as well, or, if they're not needed for the test, considering the original ones in detect-lua, you could probably remove them.

jufajardini (changes requested):
Please check inline comment :)

The CI failure should be fixed by adding a suricata.yaml file, to prevent S-V to looking at the default location for the lua script, I believe.

tests/detect-lua-01/test.yaml
catenacyber: friendly ping @TheKharleeci

TheKharleeci: Thanks for the friendly ping @catenacyber. I think I'm confused on how to go about this particular test

catenacyber: How can we help ?
What about the comments from Juliana ?

TheKharleeci: Currently working on it. Apologies for the late reply

catenacyber:
Is this still live ?

TheKharleeci:
> Is this still live ?

yes please

TheKharleeci:
Followed By: #1086

PR #691, opened 2022-01-28

catenacyber:
Rebase of #328 
And now min-version is 7


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/837

PR #698, opened 2022-01-31
GitHub ci jsonschema/v8

catenacyber:
cc @jasonish 

- then removes some fields from schema.json root, so CI will fail without the upcoming suricata PR

Replaces #590 with rebasing and removing the need for alert metadata to get http.xff


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/756

PR #700, opened 2022-01-31
detect-dce-opnum/v1: add tests

TheKharleeci:


tests/detect-dce-opnum/test.yaml
```diff
@@ -0,0 +1,21 @@
+args:
+- -k none
+
+checks:
+- filter:
+    count: 2
+    match:
+      event_type: dcerpc
```
inashivb, line 8: Logging was not introduced till 6.0.x for dcerpc so there are no events for that. Please add a min-version check for 6 here.

inashivb (changes requested):

tests/detect-dce-opnum/README.md
```diff
@@ -0,0 +1 @@
+Tests the dce_opnum keyword
```
jufajardini, line 1: Can you also mention where does the pcap come from, please? ;)

jufajardini (changes requested):

TheKharleeci:
Followed By: #702

PR #702, opened 2022-02-01
detect-dce-opnum/v2: add test

TheKharleeci:
Adds test for DetectDceOpnumTestParse01 - 02

Previous PR: #700

Corresponding Suricata PR: OISF/suricata#6910


tests/detect-dce-opnum/test.rules
```diff
@@ -0,0 +1 @@
+alert tcp any any -> any any (msg:"DCERPC"; dce_opnum:4; sid:1;)
```
inashivb, line 1: The new notation for the keyword is `dce.opnum`. Sorry for not catching this earlier. Please use this. Most keywords with `_` in the name have been changed to this pattern. Latest docs will show you the correct latest usage.

tests/detect-dce-opnum/README.md
```diff
@@ -0,0 +1,7 @@
+# Description
+
+Tests the dce_opnum keyword
+
+# PCAP
+
+The pcap comes from dcerpc-dce-opnum Suricata-Verify folder.
```
inashivb, line 7: This makes me realize we probably do not need a new test then, could you please update the existing `dcerpc-dce-opnum` test with this rule and corresponding `test.yaml`?

tests/detect-dce-opnum/test.yaml
```diff
@@ -0,0 +1,24 @@
+requires:
+  min-version: 6
```
inashivb, line 2: Previous comment about this version check only applied to `dcerpc` logging so I meant this min-version check should go in the filter for `dcerpc` logs only. Here, it applies to the `alert` check too.

inashivb (changes requested):

TheKharleeci:
Followed By: #704

PR #697, opened 2022-01-30
detect-flowbits/v3: add tests

TheKharleeci:
Adds test for FlowBitsTestSig6-11

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4911

Previous PR: #696

PR #703, opened 2022-02-01
detect-flowbits/v4: add tests

TheKharleeci:
Adds test for FlowBitsTestSig6-11

Previous PR: #697 

Corresponding Suricata PR: OISF/suricata#6899


tests/detect-flowbits/test.yaml
```diff
@@ -0,0 +1,80 @@
...
+    match:
+      event_type: alert
+      alert.signature_id: 12
+- filter:
+    count: 0
+    match:
+      event_type: alert
+      alert.signature_id: 13
+- filter:
+    count: 0
```
inashivb, line 69: This does not seem right. We set flowbits `fb1` and `fb2` in `sid`s 1 and 2, so both are set and were never unset them in the following rules. This rule should alert.

tests/detect-flowbits/test.rules
```diff
@@ -0,0 +1,13 @@
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,fb1; sid:1;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,fb2; sid:2;)
+alert ip any any -> any any (msg:"Flowbit isset ored flowbits"; flowbits:isset,fb3|fb4; sid:3;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,fb5; sid:5;)
```
inashivb, line 4: We can skip this rule. It does not add anything new to the testing.

tests/detect-flowbits/test.rules
```diff
@@ -0,0 +1,13 @@
...
+alert ip any any -> any any (msg:"Flowbit isset ored flowbits"; flowbits:isset,fb5|fb6; sid:6;)
```
inashivb, line 5: and then check for ORing among `fb1` and `fb6` here.

tests/detect-flowbits/test.rules
```diff
@@ -0,0 +1,13 @@
...
+alert ip any any -> any any (msg:"Flowbit isnotset ored flowbits"; flowbits:isnotset, fb1 | fb2 ; sid:7;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,myflow; sid:8;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:set,myflow2; sid:9;)
```
inashivb, line 8: These rules are also not adding anything new..

tests/detect-flowbits/test.rules
```diff
@@ -0,0 +1,13 @@
...
+alert ip any any -> any any (msg:"Flowbit unset"; flowbits:unset,myflow2; sid:10;)
+alert ip any any -> any any (msg:"Flowbit set"; flowbits:noalert; flowbits:toggle,myflow2; content:"GET "; sid:11;)
```
inashivb, line 10: After a `toggle` rule, we should probably have to rule to check if the flowbit was indeed toggled.

inashivb (changes requested):
I think we still need some work on this.

tests/detect-flowbits/test.rules
inashivb: You would unset `fb1` here and then toggle it later on to make sure `sid:14` is still valid and toggle also works as expected..

jufajardini: I see, thanks! In that case, we could combine this one and the toggle, which is also missing a better way of testing, right now, I think.

inashivb: Yas! :D

tests/detect-flowbits/test.yaml
TheKharleeci: I'm not sure why it does not alert.

PR #709, opened 2022-02-02
ssh-banner-only: update test.yaml checks - v2

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/662

Changes:
- Added a comment explaining that sid:2 uses a deprecated keyword. But left it there, as I figured it would be a good way to know that it's still working even if deprecated.


victorjulien:
Merged in #761, thanks!

PR #715, opened 2022-02-05
nfs: add test for memleak w file_data

inashivb:


victorjulien:
Merged in #761, thanks!

PR #721, opened 2022-02-08
tests: require version 6 for dcerpc-dce-iface-many test

zer1t0:
Update the test so it also can be applied to suricata version 6 after the fix in the PR https://github.com/OISF/suricata/pull/6949 is applied.


zer1t0:
Moved to https://github.com/OISF/suricata-verify/pull/724

PR #724, opened 2022-02-08
Set minimum version to 6 in tests related with dcerpc iface,opnum fixes

zer1t0:
Set minimum version 6 in tests related with the PR https://github.com/OISF/suricata/pull/6955.


inashivb:
Thank you, @zer1t0 ! Could you please make the commit message like we have in Suricata? e.g. instead of `set minimum version to 6 in dcerpc-dce-iface-many test`, we could write `dcerpc-dce-iface-many: set min version to 6`

Rest, LGTM. Thanks! :)

zer1t0:
Sure, moved to https://github.com/OISF/suricata-verify/pull/728

PR #725, opened 2022-02-09
detect-uricontent/v7: add tests

TheKharleeci:
Previous PR: #659

Conversion:
Uricontent-01: DetectUriSigTest02
Uricontent-02: DetectUriSigTest03
Uricontent-03: DetectUriSigTest05-07
Uricontent-04: HttpUriTest01-04

Changes: 
-  update pcap for Uricontent-04
-  Added filter for sids without alerts

Corresponding Suricata PR: https://github.com/OISF/suricata/pull/6920


tests/uricontent/detect-uricontent-04/test.yaml
```diff
@@ -0,0 +1,26 @@
...
+    match:
+      event_type: alert
+      alert.signature_id: 1
+- filter:
+    count: 3
+    match:
+      event_type: alert
+      alert.signature_id: 2
+- filter:
+    count: 3
```
inashivb, line 24: I don't know really why there aren't 4 of these events.

tests/uricontent/detect-uricontent-04/test.yaml
```diff
@@ -0,0 +1,26 @@
...
+- filter:
+    count: 4
+    match:
+      event_type: flow
+- filter:
+    count: 1
+    match:
+      event_type: stats
+- filter:
+    count: 3
```
inashivb, line 14: Same here

tests/uricontent/detect-uricontent-04/test.yaml
```diff
@@ -0,0 +1,26 @@
...
+    match:
+      event_type: alert
+      alert.signature_id: 1
+- filter:
+    count: 3
```
inashivb, line 19: And here

inashivb (commented):
@jufajardini please check my comments. Are you able to spot anything?

jufajardini:
> @jufajardini please check my comments. Are you able to spot anything?

I think we've missed a couple of things in the `writepcap.py` script for this one, which may be leading to this:
- When comparing to the unittests, I notice that originally we should have four packets to server, while in the script I think we have half-half, or... 3:1 (due to the source and destination ports).
- currently, the stats output log for this test shows that we have one `failed_tcp` flow and three `http` tx, so _maybe_ that's what is leading to those alert counts...

From the unittests, I think that we want to see, for this test:
- we get the hostname right for all 4 packets
- we get the http1 proto for all 4 packets
- we correctly match on uri_path for all 4 packets
- we correctly parse the http_method on packets 1, 2 and 4 (third: method is unknown)

So... I _think_ we want to fix this test to make sure that:
- for sids 1 and 2 we get 4 alerts;
- we add a sid 3 that checks for HTTP method, matches on `GET`, and we expect to see 3 matches for that one
- rules should be `alert http`, not `alert tcp` 
- we may want to ensure we're detecting http1, not http2, but not sure. 

What do you think, @inashivb ?

inashivb:
>     * for sids 1 and 2 we get 4 alerts;

Sounds good.

> 
>     * we add a sid 3 that checks for HTTP method, matches on `GET`, and we expect to see 3 matches for that one

Good as well.
 
>     * rules should be `alert http`, not `alert tcp`

Q: Why?

>     * we may want to ensure we're detecting http1, not http2, but not sure.

I think this is already being done so we're good here too.

inashivb (changes requested):
@TheKharleeci please address the changes requested by @jufajardini

TheKharleeci:
> I just noticed that in the unittests there are checks to match on the HTTP protocol, so I thought we could do the same here, to try to transfer those checks.
Ok. I'll work on that. Thank you

TheKharleeci:
> * or sids 1 and 2 we get 4 alerts;

Do you think I should update the pcap? I tried matching on http and using http_method. I still get just 3 matches and stats still show one failed tcp

tests/uricontent/detect-uricontent-04/test.yaml
jufajardini: I don't know either, but I did notice that Wireshark didn't recognize the packets as `http` proto...

jufajardini:
> > * or sids 1 and 2 we get 4 alerts;
> 
> Do you think I should update the pcap? I tried matching on http and using http_method. I still get just 3 matches and stats still show one failed tcp

I think that for the http_method we will only get 3 alerts for `GET` since in the original unittests one of them (packet 3) should result in unknown method.

I think the pcap should/could be rewritten to ensure that we have 4 packets to the server. Maybe also set the proto as HTTP in the script?

catenacyber:
@jufajardini what should be done with this PR ?

jufajardini:
> @jufajardini what should be done with this PR ?

I saw that there was recent activity in another PR by Modupe, I'll talk to her that the team will be taking this one over since we need these merged soon.

jufajardini:
I've created a redmine ticket for us to keep track of this one, and assigned it to myself. Closing this one for now. Thanks for your contributions, @TheKharleeci  :)

PR #732, opened 2022-02-10
Add tests for nfs* frames v1

ghostinthehive:
Goes with https://github.com/OISF/suricata/pull/6976


inashivb (commented):
I think we need one README explaining the source of pcap and describing what it does per test, please. :)

ghostinthehive:
Followed by #739

PR #739, opened 2022-02-12
Add tests for nfs* frames v2

ghostinthehive:
Previous  PR #732
Changes from Previous PR:
- Add `README.md` for all tests
- Update `test.yaml` with `min-version: 7.0` for all tests

Goes with https://github.com/OISF/suricata/pull/6992


tests/nfs2-frames/test.rules
```diff
@@ -0,0 +1,3 @@
+alert nfs any any -> any any (flow:to_server; frame:rpc.creds; content:"|00 00 00 09|werrmsche|00 00 00|"; sid:1;)
```
victorjulien, line 1: with all these rules it would be good to consider anchoring them to the start and end of buffers with `startswith`, `endswith` and/or `depth`/`offset`. This way the buffers are checked more exact.

victorjulien (changes requested):
Looking quite good. Minor request.

ghostinthehive:
Followed by #749

PR #741, opened 2022-02-14
detect-engine-state/v1: add tests

TheKharleeci:

Conversion:
- detect-engine-state-01: DeStateSigTest03-DeStateSigTest06
- detect-engine-state-02: DeStateSigTest07
- detect-engine-state-03: DeStateSigTest08- DeStateSigTest10


tests/engine-state/detect-engine-state-01/test.yaml
```diff
@@ -0,0 +1,42 @@
...
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 5
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 6
```
jufajardini, line 42: I think that `DeStateSigTest03` expects that a file is stored. I wonder if we should somehow check that here. If we do, this may be a good use case to check: https://github.com/OISF/suricata-verify/tree/master/tests/filestore-v2.1-forced

tests/engine-state/detect-engine-state-03/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+      event_type: flow
+- filter:
+    count: 1
+    match:
+      event_type: stats
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 1
```
jufajardini, line 17: This test also seems to ask for some filestore check.

jufajardini (changes requested):
Looking good! I think we should add checks for filestore somehow, though, for `01` and `03`.

I noticed we're missing tests for `DeStateSigTest01-02`, in this one...

tests/engine-state/detect-engine-state-01/test.rules
```diff
@@ -0,0 +1,6 @@
+alert http any any -> any any (flow:to_server; content:"POST"; http_method; content:"upload.cgi"; http_uri; filestore; sid:1; rev:1;)
+alert http any any -> any any (content:"GET"; http_method; content:"upload.cgi"; http_uri; filestore; sid:2; rev:1;)
+alert http any any -> any any (content:"GET"; http_method; content:"upload.cgi"; http_uri; filename:"nomatch"; sid:3; rev:1;)
+alert http any any -> any any (content:"POST"; http_method; content:"upload.cgi"; http_uri; filename:"nomatch"; filestore; sid:4; rev:1;)
+alert http any any -> any any (content:"POST"; http_method; content:"upload.cgi"; http_uri; filename:"somepicture1"; filestore; sid:5; rev:1;)
+alert http any any -> any any (content:"POST"; http_method; content:"upload.cgi"; http_uri; fileext:"jpg"; filestore; sid:6; rev:1;)
```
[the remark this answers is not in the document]
TheKharleeci: Was testing the file Keywords. Should I remove them? I don't know that it adds value to the test

tests/engine-state/detect-engine-state-01/test.yaml
jufajardini: I was thinking more of this check in the test.yaml file: https://github.com/OISF/suricata-verify/blob/master/tests/filestore-v2.1-forced/test.yaml

jufajardini: But I think I need a backup here @inashivb would that check and enabling file-info be enough, or would an actual file be needed?

tests/engine-state/detect-engine-state-01/test.rules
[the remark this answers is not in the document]
jufajardini: I was really just curious, they seem like a good addition to me :)

tests/engine-state/detect-engine-state-01/test.yaml
inashivb: yeah I think filestore would require file to be in the traffic..

PR #744, opened 2022-02-15
test/http2-files: add comparison for fast.log output - v1

jufajardini:
The idea is to compare that, each time, the alerts generated are queued
in the same order. This test was selected because it has several txs,
some of which trigger alerts for same signatures in the same packet,
and it failed with a bug related to alert queuing optimization.

Related to work on Suricata PR: https://github.com/OISF/suricata/pull/7000


victorjulien (approved):

victorjulien:
Merged in #757, thanks!

PR #746, opened 2022-02-17
smb-file.name: new test for file.name in smb

zer1t0:
Test for the file.name sticky buffer in the SMB protocol.

Related ticket https://redmine.openinfosecfoundation.org/issues/5087.


zer1t0:
file.name doesn't use unicode to detect the filenames

PR #748, opened 2022-02-21
Update bug 1450 test v0.2

chifflier:
Supersedes #742 

- Add maximum version 7.0.0 for previous test for bug-1450 (test was not well-encoded in DER, so new parser will reject it with an according error message).
- Add 2 new tests for bug-1450, which were derived from the above one, but properly DER-encoded: one for certificate with incomplete (large) header, and one for incomplete (large) content


catenacyber:
ping @victorjulien 
https://github.com/OISF/suricata-verify/pull/695 introduced a false result to S-V
cf https://github.com/OISF/suricata/pull/7076#issuecomment-1054131241

catenacyber:
Merged in #852

PR #749, opened 2022-02-21
Add tests for nfs* frames v3

ghostinthehive:
Previous PR #739 
Changes from Previous PR:

- Add `content` modifier keywords for needed rules
 
Goes with https://github.com/OISF/suricata/pull/7043


tests/nfs2-frames/test.yaml
```diff
@@ -0,0 +1,56 @@
+requires:
+  min-version: 7.0
+  files:
+      - rust/src/nfs/nfs2.rs
```
jufajardini, line 4: I think we don't need to use that any longer, these days for Rust is not optional.

jufajardini (changes requested):
Looking good to me! Just a minor thing from my end, see inline comment :)

inashivb:
Another nit: I think we can drop the `S-V` in the commit message.

ghostinthehive:
Followed by #754

PR #756, opened 2022-02-22
GitHub ci jsonschema/v9

catenacyber:
cc @jasonish

Replaces https://github.com/OISF/suricata-verify/pull/698 with rebasing and renaming http.xff to http.xff_header

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -89,6 +89,7 @@ jobs:
         working-directory: suricata
         run: python3 ../run.py --quiet
       - name: Running check-eve
+        if: ${{ matrix.branch == 'master' }}
         run: python3 ./check-eve.py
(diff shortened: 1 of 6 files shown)


jasonish:
Can you provide samples of the output? As mentioned here, https://github.com/OISF/suricata/pull/6904#issuecomment-1065600747, I don't think I'm in favour of adding a new field.  I'd rather see `xff` gone at the root, and `xff` only logged in the `http` object like here: https://github.com/OISF/suricata/pull/6904#issuecomment-1030405503 regardless of the code path that caused it to be logged.

This lets us remove a `xff` at the root, which I think is a better breaking change that renaming `http.xff`.

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/785

PR #758, opened 2022-02-22
runner: warn user when no test matches given args - v3

jufajardini:
https://github.com/OISF/suricata-verify/pull/751 splitting warnings per unmatched pattern, so if we tried to match for tls-frame, smb-frame, postgres, we get this as output, instead of in just one line:
```shell
Warning: Couldn't find any tests to match the pattern "tls-frame"
Warning: Couldn't find any tests to match the pattern "smb-frame"
Warning: Couldn't find any tests to match the pattern "postgres"
```

Task #4944

### DIFF
--- a/run.py
+++ b/run.py
@@ -1011,6 +1011,17 @@ def main():
     # Sort alphabetically.
     tests.sort()
 
+    # If we don't find any tests, let's let the user know and exit
+    if not tests:
+        # Let's make some space between our warnings and other possible output
+        print()
+        if args.testdir:
+            print("Warning: Couldn't find any directory to match the pattern \"%s\"" % args.testdir)
+        else:
+            for pattern in args.patterns:
+                print("Warning: Couldn't find any tests to match the pattern \"%s\"" % pattern.rstrip(','))
+        return 1
+
     if LINUX:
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -1011,6 +1011,17 @@ def main():
...
 
+    # If we don't find any tests, let's let the user know and exit
+    if not tests:
+        # Let's make some space between our warnings and other possible output
+        print()
+        if args.testdir:
+            print("Warning: Couldn't find any directory to match the pattern \"%s\"" % args.testdir)
+        else:
+            for pattern in args.patterns:
+                print("Warning: Couldn't find any tests to match the pattern \"%s\"" % pattern.rstrip(','))
```
inashivb, line 1022: Looks great to me. Will wait for others to see if they like the output.
One q: why are we using `.rstrip(',')` here?

jufajardini: > One q: why are we using `.rstrip(',')` here?

that's probably because I added them to my list of arguments 😅 but now that I'm checking, I think that was just bad user behavior. I'm ALSO thinking that... if we have more than one pattern and one matches but the other doesn't, this code isn't enough to cover that corner case.

So I guess we have incoming v4 :x

jufajardini (commented):
Needs more work.

PR #760, opened 2022-02-22
detect-dnp3/v1: add tests

TheKharleeci:
Conversions:

dnp3-dnp3_data-alert: DetectDNP3DataTest01-02
dnp3-dnp3_obj-alert: DetectDNP3ObjSetupTest


TheKharleeci:
Corresponding Suricata-verify PR: #OISF/suricata/pull/7057

tests/dnp3-dnp3_data-alert/test.yaml
```diff
@@ -3,8 +3,39 @@ requires:
...
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 5
+        pcap_cnt: 17
+  - filter:
+      count: 0
+      match:
+        event_type: alert
+        alert.signature_id: 6
```
jufajardini, line 36: In `DetectDNP3DataTest02`, I see that we expect the signature that is close to our sid:6 here to match. So I think we should change that signature to have a content that would match, maybe, then add a check for the pcap_cnt, like you did for sid:7.

jufajardini (changes requested):
I think this one needs a bit of work (not much!). See inline comment.

About DetectDNP3ObjSetupTest: it doesn't look to me like a test that requires conversion. (What do you say, @inashivb ? )

tests/dnp3-dnp3_data-alert/test.yaml
TheKharleeci: Sid:6 is not expected to match based on the flow and the content. only sid:5 is expected to match

tests/dnp3-dnp3_obj-alert/suricata.yaml
```diff
@@ -0,0 +1,22 @@
+%YAML 1.1
+---
+
+include: ../../etc/suricata-3.1.2.yaml
```
inashivb, line 4: Q: Why was this needed?

tests/dnp3-dnp3_obj-alert/suricata.yaml
```diff
@@ -0,0 +1,22 @@
...
+outputs:
+  - eve-log:
+      enabled: yes
+      filetype: regular
+      filename: eve.json
+      types:
+        - alert:
+            payload: yes
+            packet: yes
+            dnp3: yes
```
inashivb, line 15: This looks way different than what we have in `suricata.yaml` now. Perhaps this kind of structure was followed in old revs? ref: https://github.com/OISF/suricata/blob/master/suricata.yaml.in#L156

TheKharleeci: yes, I just followed the format that was in the previous test. Do I change it for all the dnp3 tests or just the ones I updated.

inashivb: Perhaps, just this one for now. Thanks! :)

tests/dnp3-dnp3_data-alert/test.yaml
jufajardini: You have it sorted out, sorry for the noise :P

TheKharleeci:
Followed by: #766

PR #763, opened 2022-02-24
test: add test for 'unseen' http midstream packets - v1

jufajardini:
This comes from an anonymized pcap shared on Discord. While Wireshark
tags the packets as http traffic, Suri only sees tcp, even with
stream.midstream=true.


jufajardini (commented):
I think this test needs a better name

catenacyber:
Does this require a ticket ?

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/877

PR #764, opened 2022-02-25
smb2-named-pipe-unicode: new test for smb_named_pipe keyword

zer1t0:
A test for the `smb_named_pipe` keyword using SMB2 protocol.


catenacyber:
Looks nice to me

jasonish:
Merged. Thanks!

PR #765, opened 2022-02-25
Createst add midstream arg - v1

jufajardini:
As this seems to be quite a common option to use, it seemed that we could benefit from having
the possibility to set this via command-line argument. The default is still not to have it, but a user
can now add --midstream in case they want to have arg `--set stream.midstream=true` for a given 
s-v test.

Also: updated the Readme to include `strictsums` and `midstream` args. Typo fix, copyright update.

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -142,7 +142,13 @@ def write_to_file(data):
     with open(test_yaml_path, "w+") as fp:
         fp.write("# *** Add configuration here ***\n\n")
         if not args["strictcsums"]:
-            fp.write("args:\n- -k none\n\n")
+            fp.write("args:\n- -k none\n")
+            if args["midstream"]:
+                fp.write("- --set stream.midstream=true\n")
+            fp.write("\n")
+        elif args["midstream"]:
+            fp.write("args:\n- --set stream.midstream=true\n\n")
+
         fp.write(data)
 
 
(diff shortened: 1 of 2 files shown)


createst.py
```diff
@@ -142,7 +142,13 @@ def write_to_file(data):
     with open(test_yaml_path, "w+") as fp:
         fp.write("# *** Add configuration here ***\n\n")
         if not args["strictcsums"]:
-            fp.write("args:\n- -k none\n\n")
+            fp.write("args:\n- -k none\n")
+            if args["midstream"]:
+                fp.write("- --set stream.midstream=true\n")
```
inashivb, line 147: Q: What's the relationship of `strictcsums` and `midstream` for this check to be added here?

jufajardini: They're not directly related, but I want to make sure that if one of `strictsums` has been added as an argument, we will not repeat the `args` line but will still add `midstream` setting if this option has been passed as well.
Wasn't able to think of a way of having them independent without repeating `args` ...

inashivb: I see. ONe of the better ways would be how it's done for `requires` by @TharushiJay : https://github.com/OISF/suricata-verify/pull/790/commits/411c482465c701b706abfb262e5683f13f1cd580
This would avoid the repetition of checks for individual args. :wink:

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/794

PR #767, opened 2022-02-28
tls: only one alert for an invalid handshake

catenacyber:
Instead of 2, when it happens and when the connection gets closed

cf https://github.com/OISF/suricata/pull/7076


catenacyber:
Duplicate of #748

PR #775, opened 2022-03-02
Adds quic ietf v1 test

catenacyber:
Replaces #755 with more checks about detection

PR #780, opened 2022-03-04
Adds quic ietf v1 test

catenacyber:
Replaces #775 with extensions new naming


catenacyber:
Requires suricata PR but GitHub's label does not work for now

PR #781, opened 2022-03-07
detect-engine-state/v2: add tests

TheKharleeci:

Conversion:

detect-engine-state-01: DeStateSigTest03-DeStateSigTest06
detect-engine-state-02: DeStateSigTest07
detect-engine-state-03: DeStateSigTest08- DeStateSigTest10
detect-engine-state-04: DeStateSigTest01-02

Previous PR: #741

Changes: 
- Add tests for DeStateSigTest01-02
-  Add fileinfo check


TheKharleeci:
I am not sure why detect-engine-state-04 does not give any alert. I think I'm doing something wrong with the pcap.

jufajardini:
> I am not sure why detect-engine-state-04 does not give any alert. I think I'm doing something wrong with the pcap.

I noticed that you are using specific calls to populate the packet with HTTP fields. I'd suggest doing like with those other tests where you added the buffers as part of a raw payload to a TCP stream... But I'm not sure how to achieve the same effect here. Maybe you could have 4 packets, where you would incrementally add the buffers together as raw payload... 🤔

We could still end up with a different behavior, since I'm not sure we can recriate the same scenario from `DeStateSigTest01` with an s-v test. But I would give that a try...

TheKharleeci:
Followed By: https://github.com/OISF/suricata-verify/pull/807

PR #743, opened 2022-02-14
Add tests for packet_alert_max - v3

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/694

Changes:
- Add invalid config value (0) for PACKET_ALERT_MAX in suricata.yaml to test that it will fallback to defaults.

Related to https://redmine.openinfosecfoundation.org/issues/4207
and https://redmine.openinfosecfoundation.org/issues/4942

Also reworded (and changed!!) first commit a bit, sorry if that is bad etiquette.

PR #782, opened 2022-03-10
Add tests for packet_alert_max - v4

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/743

Changes:
- Add test for corner case situation described by @catenacyber in https://github.com/OISF/suricata/pull/6896#discussion_r796064516

Related to https://redmine.openinfosecfoundation.org/issues/4207
and https://redmine.openinfosecfoundation.org/issues/4942

Goes with Suricata PR


tests/alert-max/alert-max-append-higher-priority/test.rules
```diff
@@ -0,0 +1,17 @@
+# internal id 0 -- should not trigger
```
catenacyber, line 1: This looks complex... Will it be stable ?

jufajardini: Not sure I am the best one to judge, but Victor agreed that we should try to have this test as an S-V one...

tests/alert-max/alert-max-append-higher-priority/suricata.yaml
```diff
@@ -0,0 +1,21 @@
...
+      enabled: yes
+      filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
+      filename: eve.json
+      types:
+        - alert
+        - http
+
+# Define maximum number of possible alerts that can be triggered for the same
+# packet. Default is 15
+packet-alert-max: 2
```
jufajardini, line 21: packet-alert-max had to be `2` instead of the otherwise obvious 3, because some of the rules are inspected against a fake packet, apparently, leading to a packet-alert-max value = 3 allowing for 4 alerts to be queued.

tests/alert-max/alert-max-append-higher-priority/test.rules
catenacyber: Good enough for me :-)

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/808

PR #784, opened 2022-03-17
Adds an smtp test

catenacyber:
Ticket: 4948

Needs https://github.com/OISF/suricata/pull/7050 to pass


inashivb:
> Needs [OISF/suricata#7050](https://github.com/OISF/suricata/pull/7050) to pass

Q: Why does it pass on all branches without it being merged then? :thinking:

catenacyber:
It should fail if you compile with `--enable-debug-validation`

jasonish:
Merged. Thanks!

PR #785, opened 2022-03-18
GitHub ci jsonschema/v10

catenacyber:
cc @jasonish

Replaces https://github.com/OISF/suricata-verify/pull/756 with rebasing and renaming .xff to alert.xff

### DIFF
--- a/jsonschema-rs-iter/Cargo.toml
+++ b/jsonschema-rs-iter/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "jsonschema-rs-iter"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+jsonschema = "0.13"
+structopt = {version = ">= 0.3"}
+serde_json = "1"
--- a/jsonschema-rs-iter/src/main.rs
+++ b/jsonschema-rs-iter/src/main.rs
@@ -1,0 +1,89 @@
+use std::{error::Error, fs, path::PathBuf, process};
+
+use jsonschema::JSONSchema;
+use structopt::StructOpt;
+
+type BoxErrorResult<T> = Result<T, Box<dyn Error>>;
+
+use std::fs::File;
+use std::io::BufReader;
+
+#[derive(Debug, StructOpt)]
+#[structopt(name = "jsonschema")]
+struct Cli {
+    /// A path to a JSON instance (i.e. filename.json) to validate (may be specified multiple times).
+    #[structopt(short = "i", long = "instance")]
+    instances: Option<Vec<PathBuf>>,
+
+    /// The JSON Schema to validate with (i.e. schema.json).
+    #[structopt(parse(from_os_str), required_unless("version"))]
+    schema: Option<PathBuf>,
+
+    /// Show program's version number and exit.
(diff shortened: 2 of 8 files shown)


catenacyber:
At some point, I think, we will really need an automated tool that generates schema.json out of all the json from S-V tests

jasonish:
> At some point, I think, we will really need an automated tool that generates schema.json out of all the json from S-V tests

That could introduce errors right? I see the schema as a spec that validates the EVE.  If we generate it from the EVE, we're likely to introduce something unexpected at some point.

catenacyber:
> That could introduce errors right? I see the schema as a spec that validates the EVE. If we generate it from the EVE, we're likely to introduce something unexpected at some point.

I mean more a dev tool : for a PR such as adding pgsql parser, it will generate schema.json automatically, so that the diff between old and new schema.json (and thus the extra logging brought by the PR) can easily be reviewed...

Does that make sense to you ?

jsonschema-rs-iter/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "jsonschema-rs-iter"
+version = "0.1.0"
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+jsonschema = "0.13"
+structopt = {version = ">= 0.3"}
```
jasonish, line 10: We should use `clap`, `structopt` is now in maintenance mode.

jsonschema-rs-iter/src/main.rs
```diff
@@ -0,0 +1,89 @@
...
+#[derive(Debug, StructOpt)]
+#[structopt(name = "jsonschema")]
+struct Cli {
+    /// A path to a JSON instance (i.e. filename.json) to validate (may be specified multiple times).
+    #[structopt(short = "i", long = "instance")]
+    instances: Option<Vec<PathBuf>>,
+
+    /// The JSON Schema to validate with (i.e. schema.json).
+    #[structopt(parse(from_os_str), required_unless("version"))]
+    schema: Option<PathBuf>,
```
jasonish, line 20: I think the `schema` should require an argument, and the instances be the positional arguments so I can use `xargs`:
```
find . -name \*.json | xargs ./path/to/jsonschema-rs-iter --schema schema.json
```

jasonish:
> > That could introduce errors right? I see the schema as a spec that validates the EVE. If we generate it from the EVE, we're likely to introduce something unexpected at some point.
> 
> I mean more a dev tool : for a PR such as adding pgsql parser, it will generate schema.json automatically, so that the diff between old and new schema.json (and thus the extra logging brought by the PR) can easily be reviewed...
> 
> Does that make sense to you ?

Maybe? But I also think updating the schema should just become part developing the protocol.

In fact I'm wondering if the schema file should be in the Suricata source repo so it gets versioned with the code.

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/786

jsonschema-rs-iter/Cargo.toml
catenacyber: ok done in next version

jsonschema-rs-iter/src/main.rs
catenacyber: tried something next version, is it better for you ?

PR #787, opened 2022-03-21
tests: add test for bug 5197

inashivb:
This test should fail on all branches as of now.


victorjulien:
Merged in #898, thanks!

PR #786, opened 2022-03-18
GitHub ci jsonschema/v11

catenacyber:
cc @jasonish

- extended the schema to get all S-V tests passing (now there are more pgsql stuff I guess)
- use the rust utility as it is faster than python

Replaces https://github.com/OISF/suricata-verify/pull/785 with better rust program for @jasonish

PR #789, opened 2022-03-21
GitHub ci jsonschema/v12

catenacyber:
cc @jasonish

Replaces https://github.com/OISF/suricata-verify/pull/786 with updated schema.json (maybe quic still to do)


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/797 (and #796)

PR #792, opened 2022-03-22
test-bad-hex-rule-1: add rule with incomplete hex

jasonish:
Add a rule with incomplete hex, for example "|22 2 22|" which
should result in a parse error.

Ticket #5201.


catenacyber:
Suricata PR was merged, could you retry CI ?

jasonish:
Merged.

PR #793, opened 2022-03-23
gh/workflow: use enable-debug-validation w build

inashivb:


inashivb:
Should we do debug-validation with all the jobs or have a separate one?

jasonish:
> Should we do debug-validation with all the jobs or have a separate one?

Any idea if we currenlty pass with debug-validation on?

inashivb:
> > Should we do debug-validation with all the jobs or have a separate one?
> 
> Any idea if we currenlty pass with debug-validation on?

Seems like it. I added it in this job and it seems to have passed. Perhaps, I'm not understanding what you mean. :thinking:

catenacyber:
To me, it looks ready, not just a draft, right ?

inashivb:
> To me, it looks ready, not just a draft, right ?

To me as well. Marked ready for review.

jasonish (approved):

victorjulien:
Merged in #932, thanks!

PR #794, opened 2022-03-23
Createst: add midstream arg - v2

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/765

Changes from previous PR:
- removed redundant checks.

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -1,6 +1,6 @@
 #! /bin/python
 #
-# Copyright (C) 2019 Open Information Security Foundation
+# Copyright (C) 2019-2022 Open Information Security Foundation
 #
 # You can copy, redistribute or modify this Program under the terms of
 # the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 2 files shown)


createst.py
```diff
@@ -141,10 +141,20 @@ def write_to_file(data):
...
+            fp.write("args:\n")
+            if not args["strictcsums"]:
+                fp.write("- -k none\n")
+            if args["midstream"]:
+                fp.write("- --set stream.midstream=true\n")
+            fp.write("\n")
         fp.write(data)
 
+def check_set_args():
+    features = ["midstream"]
```
inashivb, line 154: This should probably also have `strictcsums`.

createst.py
```diff
@@ -141,10 +141,20 @@ def write_to_file(data):
         sys.exit(1)
     with open(test_yaml_path, "w+") as fp:
         fp.write("# *** Add configuration here ***\n\n")
-        if not args["strictcsums"]:
-            fp.write("args:\n- -k none\n\n")
+        if check_set_args():
```
inashivb, line 144: This will never be true for checksums and we'll end up not ever having `-k none` unless we have `--midstream` as an arg.

inashivb (changes requested):
Almost there. Just a bit more work. Thank you. :)

createst.py
jufajardini: I first included it, but then I realized we maybe don't want to have it here, because when the user passes `strictcsums`, `-k none` won't be used as an argument, so my understanding was that we will have that argument when strictcsums is false, only.

jufajardini:
Fixed by: https://github.com/OISF/suricata-verify/pull/795
(hopefully)

PR #795, opened 2022-03-24
createst: add midstream arg - v3

jufajardini:
Previous PR: #794 

Changes from previous PR:
- make sure that `k none` is still set when no argument (`strictcsums`/`midstream`) is passed.

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -1,6 +1,6 @@
 #! /bin/python
 #
-# Copyright (C) 2019 Open Information Security Foundation
+# Copyright (C) 2019-2022 Open Information Security Foundation
 #
 # You can copy, redistribute or modify this Program under the terms of
 # the GNU General Public License version 2 as published by the Free
@@ -362,7 +362,7 @@ def parse_args():
     parser.add_argument("--allow-events", nargs="?", default=None,
                         help="Create filter blocks for the specified events")
     parser.add_argument("--strictcsums", default=None, action="store_true",
-                        help="Stricly validate checksum")
+                        help="Strictly validate checksum")
     parser.add_argument("--midstream", default=False, action="store_true",
                         help="Allow midstream session pickups")
 
--- a/createst.py
+++ b/createst.py
@@ -141,10 +141,29 @@ def write_to_file(data):
         sys.exit(1)
     with open(test_yaml_path, "w+") as fp:
         fp.write("# *** Add configuration here ***\n\n")
-        if not args["strictcsums"]:
-            fp.write("args:\n- -k none\n\n")
+        if check_set_args():
+            fp.write("args:\n")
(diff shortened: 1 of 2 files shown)


createst.py
```diff
@@ -141,10 +141,29 @@ def write_to_file(data):
...
+    """
+    features = ["midstream", "strictcsums"]
+    no_features = True
+    for item in features:
+        if args[item]:
+            if item != "strictcsums":
+                return True
+            no_features = False 
+        elif no_features:
+            return True
```
inashivb, line 166: hmm seems like this acts weird for `strictcsums`. I see an empty `args:` section.
I see that one option working opposite of others may be complicating things. Sorry about that.
So, we'll have to make a special handling for `strictcsums` but it can be done a bit differently. Took me a while but I have (on top of your v2),

```python
diff --git a/createst.py b/createst.py
index 05f4318..4ddbbe6 100755
--- a/createst.py
+++ b/createst.py
@@ -143,18 +143,21 @@ def write_to_file(data):
         if check_set_args():
             fp.write("args:\n")
-            if not args["strictcsums"]:
-                fp.write("- -k none\n")
-            if args["midstream"]:
-                fp.write("- --set stream.midstream=true\n")
-            fp.write("\n")
+        if not args["strictcsums"]:
+            fp.write("- -k none\n")
+        if args["midstream"]:
+            fp.write("- --set stream.midstream=true\n")
+        fp.write("\n")
         fp.write(data)
 
 def check_set_args():
-    features = ["midstream"]
+    features = ["midstream", "strictcsums"]
     for item in features:
         if args[item]:
+            if item == "strictcsums":
+                return False
             return True
+    return True
 
 def test_yaml_format(func):
     """
```

Please note that I have only taken code out of the scope of ` if check_set_args():` as it's only `args:` that really get affected by this. There are no other changes in that function.

Leaving a case table for you to check against (or point out mistakes, if there are any):

Case I: No args
Result: Should have `-k none`

Case II: `--strictcsums` only
Result: Should have no `args` section

Case III: `--midstream` only
Result: Should have `-k none` as well as `- --set stream.midstream=true`

Case IV: `--midstream` + `--strictcsums`
Result: Should only have `- --set stream.midstream=true`

inashivb (changes requested):

createst.py
jufajardini: Thanks a lot for taking your time to walk me through this >__<' 
I swear that I tested all cases yesterday, but I understand what happened with this solution of mine. The order of the elements in the array impacts in its results.
I wanted to check if I was able to come up with another working solution as well, so I'm sharing that one in a new PR now.

jufajardini:
Followed by: https://github.com/OISF/suricata-verify/pull/800

PR #799, opened 2022-03-25

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/494

Changes since v2:
- Weirdly enough the rule started to fire when proto was changed from `dcerpc` to `smb`. (Does this need a ticket? @victorjulien  or is that how rule language should have been?)
- Fixed `dcerpc` event as it must be nested in `smb`

Issues:
This seems to have been fixed in `master` and `master-6.0.x` by some patch. I'm guessing some of the fixes from Eloy.. They have not yet been backported to `master-5.0.x` so it continues to fail there.


inashivb:
cc: @catenacyber @jlucovsky

PR #805, opened 2022-04-13
test/from_base64: from_base64 transform test

jlucovsky:
Issue: 5220
This commit adds a simple test case for the from_base64 transform.


jasonish:
False sense of test passing. It checks for a file that doesn't exist.  I wonder if we should do that anymore, due to this false sense that it passes.  I'd rather have it fail until its really passes.

jlucovsky:
> False sense of test passing. It checks for a file that doesn't exist. I wonder if we should do that anymore, due to this false sense that it passes. I'd rather have it fail until its really passes.

The Suricata PR hasn't been merged yet and is still in draft status.

catenacyber:
What is the draft PR number ?

catenacyber:
Seems to be https://github.com/OISF/suricata/pull/7233

catenacyber:
Closing as https://github.com/OISF/suricata/pull/7233 was closed, right @jlucovsky ?

PR #807, opened 2022-04-19
detect-engine-state/v3: add tests

TheKharleeci:

Conversion:

detect-engine-state-01: DeStateSigTest03-DeStateSigTest06
detect-engine-state-02: DeStateSigTest07
detect-engine-state-03: DeStateSigTest08- DeStateSigTest10
detect-engine-state-04: DeStateSigTest01-02

Previous PR: https://github.com/OISF/suricata-verify/pull/781

Changes:

-  add buffers as part of the raw payload to TCP stream for detect-engine-state-04


tests/engine-state/detect-engine-state-04/writepcap.py
```diff
@@ -0,0 +1,23 @@
...
+data += ('Http Body!')
+
+data2 = ('GET /?var=val HTTP/1.1\r\n')
+data2 += ('User-Agent: Firefox/1.0\r\n')
+data2 += ('Cookie: dummy2\r\nContent-Length: 10\r\n\r\nHttp Body!')
+
+load_layer("http")
+pkts += Ether(dst='ff:ff:ff:ff:ff:ff', src='00:01:02:03:04:05')/ \
+    Dot1Q(vlan=6)/ \
+    IPv6(dst='1.2.3.4', src='5.6.7.8')/TCP(sport=6666, dport=63, flags='P''A')/Raw(load=data)
```
jufajardini, line 18: Yus, we've got alerts! :D
I'm not totally sure we are covering what DeStateSigTest01 and `DeStateSigTest02` wanted to test, though. It seems to me that there we check the same rules against buffers with different degrees of completion, and expect Suri to only match when the conditions are ideal, so to speak. 

I'm not sure if this is about testing that suri is able to detect when there's a gap, or what, though. Have you tried splitting `data` and `data2` contents over different packets, to see if you got a pcap that showed as "data reassembled from multiple packets" or something on wireshark?

(or maybe I'm over complicating things. @inashivb - what do you think?)

jufajardini (commented):
Good improvements! :) :)
I've left a comment about test detect-engine-state-04, and asked for Shivani's opinion on that.
Rest is looking good to me!

tests/engine-state/detect-engine-state-04/writepcap.py
inashivb: I do not think it is about gap handling but I do agree that we should be splitting the data and sending it to the applayer one line at a time as per the tests in C and check whether the sig fires or not for each buf.

TheKharleeci: Hi. Me again. So when you say I should split data and data2, it means we would have 4 packets?

jufajardini:
Created an individual ticket to track work on this: https://redmine.openinfosecfoundation.org/issues/6146

I'll close this for now, Modupe has told me that she'd prefer to focus on the detect-lua PR, since she has less free time, now.

PR #809, opened 2022-04-25
rules parsing: updates emerging-all.rules

catenacyber:
from http://rules.emergingthreats.net/open/suricata-6.0/emerging-all.rules

needed by https://github.com/OISF/suricata/pull/7302


jasonish:
Merged. Thanks!

catenacyber:
Wonder if we should have something special to automatically get the latest version...

jasonish:
> Wonder if we should have something special to automatically get the latest version...

No, I think it should be a deliberate activity for repeatability. It could introduce issues that may need to be fixed up.

PR #813, opened 2022-04-27
Adds test about DNS probing with junk request

catenacyber:

Replaces #417 with rebase and getting right with latest behavior


catenacyber:
Suricata PR was merged. Tests should pass

catenacyber:
Merged in #852

PR #820, opened 2022-05-04
threshold: adds test for threshold.config with by_rule

catenacyber:


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/874

PR #800, opened 2022-03-25
createst: add midstream arg - v4

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/795

Changes from previous PR:
- make sure that `args` is only printed when we do have command line args to pass

PR #821, opened 2022-05-06
createst: add midstream arg - v5

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/800

Changes from previous PR:
- rebased
- added `min-version` option explanation to the README file

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -1,6 +1,6 @@
 #! /bin/python
 #
-# Copyright (C) 2019 Open Information Security Foundation
+# Copyright (C) 2019-2022 Open Information Security Foundation
 #
 # You can copy, redistribute or modify this Program under the terms of
 # the GNU General Public License version 2 as published by the Free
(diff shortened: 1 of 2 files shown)


createst.py
```diff
@@ -155,6 +160,21 @@ def check_requires():
         if args[item]:
             return True
 
+def check_set_args():
+    """
+    Check if the user wants midstream set to true and/or to have strict
+    checksums
+    """
+    user_args = ["strictcsums", "midstream"]
+    no_user_args = True
```
inashivb, line 169: Can we make it work without this extra variable?

createst.py
```diff
@@ -155,6 +160,21 @@ def check_requires():
...
+    for item in user_args:
+        if args[item]:
+            if item != "strictcsums":
+                return True
+            no_user_args = False
+        elif no_user_args:
+            return True
```
inashivb, line 176: We can probably clean this part a bit then.

inashivb (commented):
Great job! :) Sorry for taking so long to get back. I think it works well now but you can clean the code a bit. Not marking it as changes requested but if you'd like to give cleaning the code a try, go ahead. :)

inashivb:
Continuing our discussion here: does it make sense for us to have a `no-strictcsums` instead of what we have right now? to avoid the complexity that this is bringing in the code.. But, how nice is that from a user's perspective?

jufajardini:
Feedback incorporated in https://github.com/OISF/suricata-verify/pull/851

PR #797, opened 2022-03-25
GitHub ci jsonschema/v13

catenacyber:
Part of #789 with other commit in #796

PR #826, opened 2022-05-12
GitHub ci jsonschema/v14

catenacyber:
Follows #797 with transferring schema.json to suricata repo


jasonish:
I had wanted to get working on some better integration.  But I'd like to see this just work for the most part. So on run,
- just use `cargo run`, doesn't require the tool to be installed.
- will need a rust version check, as we still build and test on Rust 1.48.0, but this doesn't build with that.
- (can we rename the tool, eve-validator or something like that?)
I think it has to just work with little concern from the end user and or CI.

jasonish:
Actually `cargo run` leads to some interesting output.  Probably a single cargo build then use the path to the build directory, to avoid having to install it.

jasonish:
Easy to make this output better?

```
===> tls13-draft28: OK
/home/jason/oisf/code/suricata-verify/tests/tls13-draft28-frames/output/eve.json - VALID
===> tls13-draft28-frames: OK
/home/jason/oisf/code/suricata-verify/tests/vxlan-decoder-01/output/eve.json - VALID
/home/jason/oisf/code/suricata-verify/tests/udp-hdr-keyword/output/eve.json - VALID
===> vxlan-decoder-01: OK
===> udp-hdr-keyword: OK
```
Just the OK is what we need if good. Otherwise the eve validation should be silent.

catenacyber:
> Probably a single cargo build then use the path to the build directory, to avoid having to install it.

How so ? That looks more difficult to me. because I do not know the path to the output of `cargo build` on different environments. Do you know how to make this ? Like `cargo build --target-dir=.` but with a better option than `--target-dir` so that I know where the binary really is

> * will need a rust version check, as we still build and test on Rust 1.48.0, but this doesn't build with that.

How so ?

Continued in #827 with the rest done

PR #828, opened 2022-05-16
schema: run eve-validator as part of regular runs

jasonish:
- jsonschema: prints warning if we really do not have it
- jsonschema: use rust utility
- jsonschema: only run for version >= 7
- schema.json: use file from suricete source tree
- schema.json: transfer to suricata
- eve-validation: always build and run eve-validator
- check-eve: make the schema a required argument
- eve-validator: ignore/target; commit Cargo.lock


check-eve.py
```diff
@@ -36,7 +36,8 @@ def validate_json(args, json_filename, schema):
     errors = []
 
     if not args.python_validator:
-        cp = subprocess.run(["eve-validator", "-q", "-s", schema, "--", json_filename])
+        progname = os.path.join(TOPDIR, "eve-validator", "target", "release", "eve-validator")
```
catenacyber, line 39: That will not work in my setup where I have `export CARGO_BUILD_TARGET="x86_64-apple-darwin"`

catenacyber:
Thanks for improving on this...
I have only one concern about the unpredictability of the build dir for the rust validator...

check-eve.py
jasonish: Are you cross compiling?

jasonish: Would it be OK if the script cleared that var first?  This is being built to run on the host that is doing the building, so I don't see a case where a custom target would be required.

catenacyber: > Are you cross compiling?

No, but (IIRC) this is required by Suricata fuzzing (to rebuild std with sanitizers)

> Would it be OK if the script cleared that var first? This is being built to run on the host that is doing the building, so I don't see a case where a custom target would be required.

For me, I guess so.

But my point was rather that I do not know how to get the build directory out of rust in every case...

jasonish: Could just use this var in the target directory.  I don't think I've ever seen a case where the `--target` arg or var resulted in an output path that didn't match.

catenacyber: If this env var is the only way to modify it, I am ok... But it may be in some cargo configuration file as well, do you know ?

jasonish:
You can also do `cargo build --target`.. But it also puts it into a directory of the same name. But we don't provide a way to pass that through, so the env var is the only way to pass that through.

It can be specified in a config file but I bet thats a not very common option.. Then I do have a command line flag to disable validation.  I prefer this trade-off over having to install it myself.

catenacyber:
So, good enough for me if there is no perfect solution for cargo :-)

Or `find . -name "eve-validator" -type f | while read I; do cp $I .; done`
...

PR #830, opened 2022-05-25
eve-validation: always run if schema available - ish/v2

jasonish:
Previous PR: #828

Changes:
- Clear CARGET_BUILD_TARGET as we're always building for the localhost, at
  least for now.
- Remove check for v7, this will run as long as their is a schema.json to
  check.


catenacyber:
cf https://github.com/OISF/suricata/pull/7412

PR #845, opened 2022-06-08

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/799

Changes since v2:
- sid checks
- added readme
- uncommented second rule
- min version check for 6.0 as 5 is near EOL and it seems complicated to fix there


catenacyber:
Merged in #852

PR #851, opened 2022-06-10
createst: add midstream arg - v6

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/821

Changes from previous PR:
- cleaner and more concise version of `check_set_args()` as @inashivb request

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -141,12 +141,17 @@ def write_to_file(data):
         sys.exit(1)
     with open(test_yaml_path, "w+") as fp:
         fp.write("# *** Add configuration here ***\n\n")
-        if not args["strictcsums"]:
-            fp.write("args:\n- -k none\n\n")
         if check_requires():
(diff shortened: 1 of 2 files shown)


createst.py
```diff
@@ -155,6 +160,23 @@ def check_requires():
...
+    """
+    Element order matters here. 'strictcsums' must be the last element.
+    """
+    user_args = ["midstream", "strictcsums"]
+    for item in user_args:
+        if args[item]:
+            if item == "midstream":
+                return True
+            elif item == "strictcsums":
+                return False
```
inashivb, line 177: Since we only have to treat `strictcsums` as a special case, we can probably skip the check for `midstream` as we return `True` by default from this fn.

inashivb (commented):
Looks great. A  bit sad that we have to ensure the args order in array but I cannot quickly come up w a better solution. :/

PR #856, opened 2022-06-14
Add http host keyword tests

jmtaylor90:


catenacyber:
Looks like this needs `min-version: 7` in some test.yaml, right ?

jmtaylor90:
will check on that and update the PR. thanks @catenacyber

jmtaylor90:
continued in https://github.com/OISF/suricata-verify/pull/885

PR #860, opened 2022-06-19
Dataset ops v2.1

regit:
Implement tests for ip.src and ip.dst combined to ip and ipv6 dataset.


catenacyber:
Should not this fail until suricata PR is merged ?

regit:
No it detect the file I did add and skip the test.

PR #863, opened 2022-06-27
createst: add option for custom suricata.yaml

inashivb:


jlucovsky (commented):
nit: please update the copyright years for `createst.py`

createst.py
```diff
@@ -355,6 +357,8 @@ def parse_args():
                         help="Stricly validate checksum")
     parser.add_argument("--min-version", default=None, metavar="<min-version>",
                         help="Adds a global minimum required version")
+    parser.add_argument("--c", metavar="<path-to-suricata.yaml>",
```
victorjulien, line 360: double dash c `--c` seems a bit of an anti pattern to me. Generally we reserve single letter options for single dash. Maybe we should use `--cfg`? or `--suricata-yaml`?

victorjulien:
Works as expected, thanks. Just an option name comment.

PR #867, opened 2022-07-01
Adds regression test against forced filestore

catenacyber:
Cf https://redmine.openinfosecfoundation.org/issues/5408


victorjulien:
Test fails for 5 due to different number of fileinfo records than expected.

catenacyber:
Do you mean 5 with a back ported fix ?

victorjulien:
No I don't know why the test doesn't match, but just wanted to point out it fails in a different way than the segv in 6 and master.

catenacyber:
Ok, adding `min-version: 6` in https://github.com/OISF/suricata-verify/pull/869

PR #870, opened 2022-07-01
createst: add midstream arg - v7

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/851

Changes from previous PR:
- make `check_set_args()` more readable, and dictionary element's order irrelevant, again

### DIFF
--- a/createst.py
+++ b/createst.py
@@ -141,12 +141,17 @@ def write_to_file(data):
         sys.exit(1)
     with open(test_yaml_path, "w+") as fp:
         fp.write("# *** Add configuration here ***\n\n")
-        if not args["strictcsums"]:
-            fp.write("args:\n- -k none\n\n")
         if check_requires():
             fp.write("requires:\n")
         if args["min_version"]:
             fp.write("   min-version: %s\n\n" % args["min_version"])
+        if check_set_args():
+            fp.write("args:\n")
+            if not args["strictcsums"]:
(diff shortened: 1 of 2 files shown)


inashivb (approved):
LGTM. Thank you! :)

createst.py
```diff
@@ -141,12 +141,17 @@ def write_to_file(data):
...
             fp.write("requires:\n")
         if args["min_version"]:
             fp.write("   min-version: %s\n\n" % args["min_version"])
+        if check_set_args():
+            fp.write("args:\n")
+            if not args["strictcsums"]:
+                fp.write("- -k none\n")
+            if args["midstream"]:
+                fp.write("- --set stream.midstream=true\n")
+            fp.write("\n")
```
jasonish, line 154: I do find the usage of this function `check_set_args()` hard to follow. It appears to be used only to determine if we should output an `args` object?

If so, I find the following pattern easy to follow, and less error prone to extension as there are no special cases:
```
suricata_args = []
if not args["strictcsums"]:
    suricata_args.append("-k none")
if args["midstream"]:
    suricata_args.append("--set stream.midstream=true")
if suricata_args:
    fp.write("\n".join(["- {}".format(a) for a in suricata_args]))
```

jufajardini:
Thanks for the feedback 😅 

Replaced by: https://github.com/OISF/suricata-verify/pull/882

PR #874, opened 2022-07-06
threshold: adds test for threshold.config with by_rule

catenacyber:
#820 rebased now that suricata PR was merged


jasonish:
Merged. Thanks!

PR #877, opened 2022-07-08
test: add test for 'unseen' http midstream packets - v2

jufajardini:
Previous PR: #763 

Describe changes:
- redid the sv tests, as after reviewing it didn't seem that the first version was actually showing anything useful
- now we have two tests, one with each shared pcap. `01` shows the unseen `http` packets behavior, `02` shows how Suri is able to identify them when part of a larger stream.

** To note **
I was not able to get Suri to log the payload, in any of the tests.
I also did not manage to get alerts with the rule shared on the Discord chat, nor with another sid that showed up in one of the logs shared by them...

Redmine ticket: [Bug #5437](https://redmine.openinfosecfoundation.org/issues/5437)


jufajardini:
Should the test `bug-1450-02` require min-version: 6 or something like that?

catenacyber:
Should not CI fail until ticket is resolved ?

jufajardini:
> Should not CI fail until ticket is resolved ?

You're right, I wasn't sure how to proceed with the creation of this test, tbh.
Maybe I could add a check for the payload field? 🤔

catenacyber:
> You're right, I wasn't sure how to proceed with the creation of this test, tbh. Maybe I could add a check for the payload field? 🤔

I had not looked into the ticket yet.
Now that I read it, I am not sure I understand the issue...

If this is about the `payload` field, I think there are other open issues about it already...

jufajardini:
> 

The `payload` field may be a side effect. The issue is that there's some `http` traffic in the anonymized pcap that is not being picked as `http` by Suri, even if we pass `stream.midstream=true`. 

I said I think you are right because I thought that we should have a test that is able to show that Suri has this bug. But I wasn't able to...

jufajardini:
> > You're right, I wasn't sure how to proceed with the creation of this test, tbh. Maybe I could add a check for the payload field? thinking
> 
> I had not looked into the ticket yet. Now that I read it, I am not sure I understand the issue...
> 
> If this is about the `payload` field, I think there are other open issues about it already...

The `payload` field may be a side effect. The issue is that there's some `http` traffic in the anonymized pcap that is not being picked as `http` by Suri, even if we pass `stream.midstream=true`. 

I said I think you are right because I thought that we should have a test that is able to show that Suri has this bug. But I wasn't able to...

catenacyber:
Am I understanding correctly that you do not manage to reproduce the bug described in the redmine ticket https://redmine.openinfosecfoundation.org/issues/5437 ?

Do I read correctly the redmine ticket that the bug is that some flows do not have `app-layer: http` ? These flows do not have 3way handshake but suricata uses `stream.midstream=true`

jufajardini:
> Am I understanding correctly that you do not manage to reproduce the bug described in the redmine ticket https://redmine.openinfosecfoundation.org/issues/5437 ?
> 
The part I was not able to reproduce is the have payload to be printed. In both tests, that is not shown in these tests, while apparently it should be seen for the full pcap (test 02)
> Do I read correctly the redmine ticket that the bug is that some flows do not have `app-layer: http` ? These flows do not have 3way handshake but suricata uses `stream.midstream=true`
If I understand your question correctly, yes, the bug is the fact that this HTTP flow isn't recognized by Suri, regardless of having `stream.midstream` set to true...

catenacyber:
> > Do I read correctly the redmine ticket that the bug is that some flows do not have `app-layer: http` ? These flows do not have 3way handshake but suricata uses `stream.midstream=true`
> If I understand your question correctly, yes, the bug is the fact that this HTTP flow isn't recognized by Suri, regardless of having `stream.midstream` set to true...

So, the S-V test should be about event_type==flow and app_layer==http, right ?

jufajardini:
> > > Do I read correctly the redmine ticket that the bug is that some flows do not have `app-layer: http` ? These flows do not have 3way handshake but suricata uses `stream.midstream=true`
> 
> > If I understand your question correctly, yes, the bug is the fact that this HTTP flow isn't recognized by Suri, regardless of having `stream.midstream` set to true...
> 
> So, the S-V test should be about event_type==flow and app_layer==http, right ? And it should fail on master

thanks, I was feeling rather lost with how to approach this one. I'll incorporate those changes :)

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/926

PR #885, opened 2022-07-19
Add http host tests v3

jmtaylor90:
fixes from comments in #856

### DIFF
--- a/tests/test-bad-http-host-rule-1/test.yaml
+++ b/tests/test-bad-http-host-rule-1/test.yaml
@@ -1,0 +1,23 @@
+requires:
+  min-version: 7.0.0
+
+  features:
+    - HAVE_LIBJANSSON
+
+command: |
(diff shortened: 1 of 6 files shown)


tests/test-bad-http-host-rule-1/test.yaml
```diff
@@ -0,0 +1,23 @@
+requires:
+  min-version: 7.0.0
+
+  features:
+    - HAVE_LIBJANSSON
+
+command: |
+  ${SRCDIR}/src/suricata --set classification-file="${SRCDIR}/classification.config" --set reference-config-file="${SRCDIR}/reference.config" -l ${OUTPUT_DIR} -c ${TEST_DIR}/suricata.yaml -r ${TEST_DIR}/ -S ${TEST_DIR}/test.rules
```
catenacyber, line 8: Why do we need a specific command ?

jmtaylor90: I don't know that it's needed, it's likely just from an existing test case I copied to create these two tests or I just copied the test.yaml from the readme. If it isn' t needed, I will do a new PR without.

catenacyber: Would you please try without ?
This is a non-standard option that is rarely used ;-)

jmtaylor90:
continued in https://github.com/OISF/suricata-verify/pull/886

PR #887, opened 2022-07-22
Add http host tests v5

jmtaylor90:
- removed libjansson to address comment in previous PR


catenacyber:
Thanks for the improvements

PR #888, opened 2022-07-25
tests: add tests for reject action - v1

jufajardini:
Trying to add more tests that use rules with the `reject` action, as the bug
reports that the reject action is no longer working.

I think it is better to create a few more, still.

Bug #5458


catenacyber:
Is it normal that the tests pass when the redmine ticket is not closed ? https://redmine.openinfosecfoundation.org/issues/5458

jufajardini:
> Is it normal that the tests pass when the redmine ticket is not closed ? https://redmine.openinfosecfoundation.org/issues/5458

From what Victor explained to me, checking the reject action with an SV test is probably more involved than this, since reject packets don't usually show up in packet captures without some workaround. I'm guessing that's why the tests pass, regardless of the PR. 

As we went with unittests, for now, I'll close this PR. Thanks for pinging :D

PR #238, opened 2020-05-29
tests/bsize Add test cases for bsize

jlucovsky:
This commit adds several test cases for the `bsize` keyword.

[Suricata PR #4949](https://github.com/OISF/suricata/pull/4949)

PR #893, opened 2022-07-27
add bsize keyword tests v2

jmtaylor90:
This adds signature tests for the various bsize keyword values.


tests/test-bsize-values/test.yaml
```diff
@@ -0,0 +1,31 @@
+args:
+- -k none
+
+requires:
+   min-version: 7.0.0
```
catenacyber, line 5: Does size not work on previous versions ?

catenacyber:
cc @jlucovsky as you work on `bsize` in some PRs, right ?

catenacyber:
cf #238

tests/test-bsize-values/test.yaml
jmtaylor90: I will double check, I thought I had 5 as the min version on this.

jmtaylor90: The 6.x checks fail if I lower the version.

catenacyber: Right, `<=` must be min 7 even if bsize < existed before :-)

jlucovsky: Older versions only supported `>` and `<`

@jmtaylor90 One path is to create a test with just these for 6 and newer and then a test with the other operands for 7 and newer.

jmtaylor90: Sure, I can do that. I will resubmit with the appropriate tests split out.

jmtaylor90:
continued in #922

PR #872, opened 2022-07-04
Output emptyarrays 5167 v2

catenacyber:
cf https://github.com/OISF/suricata/pull/7604

Replaces #871 with better jq

PR #897, opened 2022-07-29
Output emptyarrays 5167 v3

catenacyber:
cf https://redmine.openinfosecfoundation.org/issues/5167

Replaces #872 with making the check only for Suricata 7

The other option would be to backport https://github.com/OISF/suricata/pull/7612

### DIFF
--- a/check-eve.py
+++ b/check-eve.py
@@ -40,11 +40,12 @@ def validate_json(args, json_filename, schema):
     status = "OK"
     errors = []
 
-    cp = subprocess.run(["jq", "-e", 'path(.. | select(type == "array" and length == 0)) | join(".")', json_filename])
-    if cp.returncode != 4 and cp.stdout is not None:
-        status = "FAIL"
-        errors.append("jq returned %d" % cp.returncode)
-        errors.append(cp.stdout)
+    if args.minseven:
+        cp = subprocess.run(["jq", "-e", 'path(.. | select(type == "array" and length == 0)) | join(".")', json_filename])
+        if cp.returncode != 4 and cp.stdout is not None:
+            status = "FAIL"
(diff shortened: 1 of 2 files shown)


check-eve.py
```diff
@@ -77,6 +84,7 @@ def main():
     parser.add_argument("file", nargs="?", default=[])
     parser.add_argument("-q", dest="quiet", action="store_true")
     parser.add_argument("-s", dest="schema", action="store", required=True)
+    parser.add_argument("--min-seven", dest="minseven", action="store_true")
```
jasonish, line 87: I'd rather see this as `--check-empty-array` or something like that.

jasonish:
Can we use this to validate the length in the schema? https://json-schema.org/understanding-json-schema/reference/array.html#length

check-eve.py
catenacyber: Ok.
I thought it would fit for other additional checks...

catenacyber:
>  I don't think we depend on jq at all in S-V

Oh, I thought so...
test.yaml checks are only run with pure python then ?

> Drop the Python eve validator

Ok, but that is not directly related, or is it and am I missing something ?

catenacyber:
> Can we use this to validate the length in the schema? https://json-schema.org/understanding-json-schema/reference/array.html#length

That works nice
```
suricata-verify/tests/filesize-keyword/output/eve.json - INVALID. Errors:
1./files [{"filename":"\\eicar","gaps":false,"size":68,"state":"CLOSED","stored":false,"tx_id":6}] has less than 1000 items
```

So, I only need a suricata PR, not a S-V one

catenacyber:
Replaced by https://github.com/OISF/suricata/pull/7918

PR #907, opened 2022-08-17
tests: add ips midstream exception policy tests - v1

jufajardini:
Add tests for the exception policy for midstream streams (https://redmine.openinfosecfoundation.org/issues/5468)

exception-policy-midstream-01: tests the fail open behavior
exception-policy-midstream-02: tests the fail closed behavior
exception-policy-midstream-03: tests the default behavior

Tests will pass for 7 and fail for 6 (until backported).


victorjulien:
can you add a test for the case of:
http midstream (SYN/ACK pickup)
pass policy (so no detection)
rule that would have matched if not for the pass policy
http record still logged in eve.http

you can probably base this on one of the testmyids pcaps

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/916

PR #915, opened 2022-08-22
tests: test flow droped but applayer event logged - v1

jufajardini:
It seems that Suricata will log an applayer event for a dropped flow,
for the second packet of the flow. This test demonstrates such behavior,
so we can investigate it.

This PR adds tests for appprotos `http` and `smb`. Must find pcaps for other protocols that require `midstream=true` and in which the stream is first seen for `ACK` state.

Related to
[Task #5510](https://redmine.openinfosecfoundation.org/issues/5510)


catenacyber:
Should this be a draft PR ?

jufajardini:
> Should this be a draft PR ?

Thanks :)

catenacyber:
Juliana, what is the status for this draft ? Are you expecting something from us ?

jufajardini:
> Juliana, what is the status for this draft ? Are you expecting something from us ?

Sorry for the very late answer. Probably waiting to understand if this is a real issue/if this test is good enough to prove anything...
As I'm addressing a bug that might be connected to this, I'll try to use this test. This may help me figure out how good it is :P

catenacyber:
Should the tests fail if there is a problem for now ?

jufajardini:
New PR with tests failing to better illustrate situation: https://github.com/OISF/suricata-verify/pull/1169

PR #917, opened 2022-08-24
tests: add dsize option signature tests

jmtaylor90:


tests/test-dsize-values/test.yaml
```diff
@@ -0,0 +1,37 @@
+# *** Add configuration here ***
+
```
jufajardini, line 2: nit: remove these lines

jufajardini (changes requested):
Aside a nit comment, it looks good to me! :)

tests/test-dsize-values/test.yaml
jmtaylor90: oops, will do. thanks for reviewing!

jmtaylor90:
continued in #918

PR #493, opened 2021-04-29
filemagic: test interaction with flowbits

catenacyber:
Replace #492 with needing feature `MAGIC`

https://github.com/OISF/suricata/pull/6092

PR #924, opened 2022-09-01
tests: add flowbit bug test

regit:


tests/flowbit-chained/test.rules
```diff
@@ -0,0 +1,3 @@
+alert http any any -> any any (msg:"digicert.com bypass"; http.host; content:".digicert.com"; endswith; flow:established,to_server; flowbits: set, stamus.pass; sid:1; rev:1; priority:3;)
+alert http any any -> $HOME_NET any (msg:"HTTP learning"; flow:established,to_client; http.server; flowbits:isnotset, stamus.pass; dataset:set,http,type string,state output/http.intel; sid:2; rev:1; priority:2;)
+#alert http any any -> $HOME_NET any (msg:"HTTP learning"; flow:established,to_client; http.server; flowbits:isnotset, stamus.pass, data; dataset:set,http,type string,state output/http.intel; sid:2; rev:1; priority:2;)
```
victorjulien, line 3: this test doesn't use the new functionality?

catenacyber:
#493 may be related

tests/flowbit-chained/test.rules
regit: No, I wanted to exhibit the problem for test purpose but it would make sense indeed to add uncomment third signature and get it tested.

catenacyber:
I confirm this is the same bug as #493 and it is fixed by https://github.com/OISF/suricata/pull/7317 (@regit if you want to try it)

catenacyber:
@regit should this be closed in favor of #493 ?

regit:
This is the same case indeed. I'm good with proposal.

PR #925, opened 2022-09-01
runner: catch sub-process errors; allow falsey pcap filename - v1

jasonish:
- runner: error out on errors from the runner
- runner: don't fail if the pcap filename is falsey
- runner: print suricata command line in verbose mode


victorjulien:
Merged in #932, thanks!

PR #926, opened 2022-09-02
tests: add test for unseen http midstream traffic - v3

jufajardini:
Add two tests about the bug 5437, where a midstream HTTP traffic isn't recognized by Suri. 
bug-5437-01 showcases the bug
bug-5437-02 showcases Suri properly identifying the packets, once the pcap also has the underlying TCP stream

Previous PR: #877 

Updates from last PR: incorporate feedback from past PR, which makes the tests make sense, now (I hope)

Related to bug https://redmine.openinfosecfoundation.org/issues/5437


catenacyber:
So, I understand the root cause of this.

If we start a flow with a midstream packet with TCP `FIN` flag set, we simply ignore it from function `StreamTcpPacketStateNone`

Fixed by https://github.com/OISF/suricata/pull/8320

catenacyber:
Could we get rid of the suricata.yaml ? And have the rules not using variables like `HOME_NET` ?

jufajardini:
> Could we get rid of the suricata.yaml ? And have the rules not using variables like `HOME_NET` ?

I'll see to it!

jufajardini:
Feedback incorporated with https://github.com/OISF/suricata-verify/pull/1051

PR #928, opened 2022-09-06
tests: add basic test for reject action for exception policy - v1

jufajardini:
I wanted to at least show that what we can measure is working. Not sure if we can do more.

This test only ensures that Suricata does drop packet/flow after reaching an exception condition with a 'reject' policy.

Related to
Task #5503


jufajardini:
PR has been merged: https://github.com/OISF/suricata/pull/7833

victorjulien:
Merged in #932, thanks!

PR #935, opened 2022-09-15
framework: sets one minute timeout for process

catenacyber:
so as to catch unhandled exception if stdout/stderr has not an ascii character


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/936

PR #936, opened 2022-09-15
framework: more robustness against failing Suricata

catenacyber:
so as to catch unhandled exception if stdout/stderr has not an ascii character

Adds a timeout of 5 minutes per test

Replaces #935

### DIFF
--- a/run.py
+++ b/run.py
@@ -51,6 +51,8 @@ LINUX = sys.platform.startswith("linux")
 suricata_bin = "src\suricata.exe" if WIN32 else "./src/suricata"
 suricata_yaml = "suricata.yaml" if WIN32 else "./suricata.yaml"
 
+PROC_TIMEOUT=300
+
 if LINUX:
     manager = mp.Manager()
     lock = mp.Lock()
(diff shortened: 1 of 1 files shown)


jasonish:
Would https://github.com/OISF/suricata-verify/pull/901 catch the error?

catenacyber:
I do not think so.
You can test with current Suricata master branch, test ut-complete

jasonish:
I'm not sure I agree with classifying these as io errors.  Instead we should handle this more gracefully. I'm thinking we open the stdout and stderr log files in binary mode, so we can pass the bytes through as is.  Then decode for display purposes.

For example: https://github.com/OISF/suricata-verify/pull/937

The timeout is still a good idea though.

catenacyber:
Problem with #937 is that it tests pass on master when I think they should fail

jasonish:
Please see https://github.com/OISF/suricata-verify/pull/937#issuecomment-1249897016.  I think I'm OK with validating the data as UTF-8.. But it needs to happen after writing the output to the file so we maintain the bad data in a log.

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/948

PR #937, opened 2022-09-15
runner: handle binary output from suricata stderr/stdout

jasonish:


jasonish:
An alternate solution to #936 that handles non-unicode output more gracefully and doesn't count it as an error.

catenacyber:
This passes on suricata master when I think it should fail.

IIRC, this kind of error has happened in the past and we were happy to have Suricata-verify providing these input to fix Suricata

catenacyber:
Anyways, maybe there should be 2 commits because there are 2 bugs : the line decoding, and the never ending S-V

jasonish:
> > This passes on suricata master when I think it should fail.
> 
> How so?

If its because you think Suricata logging non-utf8 characters to its stdout/stderr, I'm not sure I agree.  At least not logged as an I/O error.  Maybe some other check that validates the stderr/stdout for utf-8 and log it as such.

But I'm not sure I agree that this should fail on master.  I think it was just incomplete data handling in the runner.  We could simply log something here when it fails to decode as well, but shouldn't fail the test.

catenacyber:
> If its because you think Suricata logging non-utf8 characters to its stdout/stderr, I'm not sure I agree. At least not logged as an I/O error. Maybe some other check that validates the stderr/stdout for utf-8 and log it as such.

I can not find the previous similar bug I thought existed.
So maybe, you are right :-)

But then, S-V has been passing this check with all tests, and here it is finding a real bug for me, that we do not log the right string.
So, I still think there is some value in what it brought.
I am bad at naming. Do you have a better idea than io_error ?

> We could simply log something here when it fails to decode as well, but shouldn't fail the test.

Would we fix this if it was just a log instead of a hard failure ?

jasonish:
> > If its because you think Suricata logging non-utf8 characters to its stdout/stderr, I'm not sure I agree. At least not logged as an I/O error. Maybe some other check that validates the stderr/stdout for utf-8 and log it as such.
> 
> I can not find the previous similar bug I thought existed. So maybe, you are right :-)
> 
> But then, S-V has been passing this check with all tests, and here it is finding a real bug for me, that we do not log the right string. So, I still think there is some value in what it brought. I am bad at naming. Do you have a better idea than io_error ?

utf8_decode_error?

> 
> > We could simply log something here when it fails to decode as well, but shouldn't fail the test.
> 
> Would we fix this if it was just a log instead of a hard failure ?

It depends. Do we think that non-utf8 output on stderr/stdout should be a hard error?  Suricata has no issue with it.  With this fix, S-V has no issue with it.  So its more a "policy" decision.  IMO its not a bug, but maybe its something we shouldn't allow.

jasonish:
Merged this as-is for now, fixing as a bug in S-V.  The idea here was to pass the output of Suricata stderr and stdout verbatim to log files. This should have never gone through utf-8 decoding in the first place, probably a left over from Python 2 days.

UTF-8 validation could be added on here I suppose, using an error feedback mechanism like in #936 as the pipe_readeer should never fail, as it will lead to keeping the Suricata process around forever.

PR #944, opened 2022-09-21
Tls frag/v7

victorjulien:


tests/tls-client-hello-frag-01/test.yaml
```diff
@@ -0,0 +1,13 @@
...
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 1
+  - filter:
+      count: 1
+      match:
+        event_type: tls
+        tls.sni: www.google.com
```
jasonish, line 13: @victorjulien Is this test for 7 only?

victorjulien: ah ya probably. Does it fail in 6?

PR #716, opened 2022-02-07
tests: Add test case for unreachable content 

jlucovsky:
Continuation of #673

This PR tests the dsize validation logic in Suricata PR 6838 (link below).

Changes:
- Existing test cases for dsize validation were restricted to Suricata versions less than 7
- These test cases were uses to create version 7 and later tests that expect the modified error message
- Typo fixups :-)

This PR accompanies [Suricata PR #6838](https://github.com/OISF/suricata/pull/6838)

PR #950, opened 2022-09-26
tests: Add test case for unreachable content

jlucovsky:
Continuation of #716

This PR tests the dsize validation logic in Suricata PR 6838 (link below).

Changes:
- Existing test cases for dsize validation were restricted to Suricata versions less than 7
- These test cases were uses to create version 7 and later tests that expect the modified error message
- Typo fixups :-)

Updates
- Add test case for rule with mix of content and negated content.

This PR accompanies [Suricata PR #7932](https://github.com/OISF/suricata/pull/7932)


catenacyber:
Merged

PR #957, opened 2022-10-05
Rename long running test to run earlier

jasonish:
This means we won't be waiting on it at the end, as it likely finished
already given the multi-threaded nature of S-V.


catenacyber:
Needing rebase

PR #959, opened 2022-10-14
tests: add test to verify unmatched signatures does not write to dataset

norg:
Suricata Verify test related to https://redmine.openinfosecfoundation.org/issues/5576


tests/datasets-validate-postmatch/test.rules
```diff
@@ -0,0 +1 @@
+alert http any any -> $HOME_NET any (msg:"HTTP learning"; flow:established,to_client; http.content_type; content:"noone"; http.server; content:"ECS"; fast_pattern; dataset:set,http,type string,state output/http.intel; sid:2; rev:1; priority:2;)
```
inashivb, line 1: We should keep the filename to `http.intel` only as `output` dir doesn't exist initially and is written everytime we run the test.

inashivb (changes requested):
Not moved forward with this yet. Was just trying and found a complaint of rule failure by Suricata. :)

inashivb (approved):
The earlier review was incorrect. This is fine but needs tweaks if run with gdb

catenacyber:
cf https://github.com/OISF/suricata/pull/8114

victorjulien:
PR is too old to restart the CI, can you resubmit?

catenacyber (changes requested):
Rebase + Resubmit was requested

PR #965, opened 2022-10-19
tests: filestore alert event

regit:
Suricata verify test for https://redmine.openinfosecfoundation.org/issues/4881


catenacyber:
Technically, this is not a bug because the file is not yet _stored_ when Suricata produces the alert
- you get ` "event_type": "alert",` with `"stored": false,` at  `"pcap_cnt": 91,
- you get ` "event_type": "fileinfo",` with `"stored": true,` at  `"pcap_cnt": 94,
 
Indeed, you get the alert as soon as possible when tou see the content, but the file is only marked as stored when it is over (and remaining content has been processed)

So, the question is do we want to change this behavior, or introduce a new key for logging ?

Fixed by https://github.com/OISF/suricata/pull/8321 with behavior change

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1058

PR #967, opened 2022-10-20
Add unit tests for bug 4759

zagge-cgeo:


tests/bug-4759/test.yaml
```diff
@@ -0,0 +1,302 @@
+# *** Add configuration here ***
+
```
jufajardini, line 2: Could you please remove these lines? They're a leftover from the createst script and should not go in the final tests.

jufajardini (changes requested):
We're trying to make a good practice of adding a README file to suricata-verify tests, to say where the pcap comes from and any other relevant info (if any - this part isn't mandatory).

Could you please add that, too? Thanks :)

PR #972, opened 2022-10-22
tests: add test for issue 4751

hsadia538:
Previous PR #969 

Redmine [ticket](https://redmine.openinfosecfoundation.org/issues/4751)

Describe changes:
- Update commit message
- Update Readme file
- Update `test.yaml` to add `min-version`

PR #975, opened 2022-10-23
tests: add test for issue 4751

hsadia538:
Previous PR https://github.com/OISF/suricata-verify/pull/972

Redmine [ticket](https://redmine.openinfosecfoundation.org/issues/4751)

Describe changes:

- Update test.yaml


hsadia538:
Do we need a ticket for this one or should I add something to the main issue ticket? ( i already added the pr link)

inashivb:
> Do we need a ticket for this one or should I add something to the main issue ticket? ( i already added the pr link)

We do not need a separate ticket for s-v tests added right with the change/feature.

jasonish:
If I'm reading this correctly this configures Suricata to only log the https records right? If so, it would be good to test for that as well.  One way is to create a check on the number of dns records.  It should equal the number of https records.

hsadia538:
> If I'm reading this correctly this configures Suricata to only log the https records right? If so, it would be good to test for that as well. One way is to create a check on the number of dns records. It should equal the number of https records.

I have created a Suricata verify test to verify the output for ` rrtype : HTTPS.`
For context: This aims to parse rrtype 65 as HTTPS.

jasonish:
Ok, the pcap contains a single packet, so what I suggested doesn't make sense here.

jasonish (approved):

jasonish:
Merged. Thanks!

PR #980, opened 2022-10-25
detect-icmp-seq: add test

AkakiAlice:
Task: #5597

Add test for DetectIcmpSeqMatchTest01


tests/detect-icmp-seq/test.yaml
```diff
@@ -0,0 +1,18 @@
+# *** Add configuration here ***
+
```
jufajardini, line 2: Can you please remove these? They're from the createst script, and are not needed ;)

tests/detect-icmp-seq/test.yaml
```diff
@@ -0,0 +1,18 @@
...
+args:
+- -k none
+
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: alert
```
jufajardini, line 10: If we have two rules, we should have two checks here, one for each rule sid. You can see how to do that check in the test bug-1401-01 (or many others)

jufajardini (changes requested):
Thanks for this work!

We're trying to make it a good practice to have a README file added to the SV tests, stating where was the pcap from (for instance, created with Scapy, or shared on bug report); and any test behavior that is being tested, if one wants to add such info. Could you add that?

AkakiAlice:
> 

Sure! I will add README file

tests/detect-icmp-seq/test.yaml
AkakiAlice: Now I think I get it! In the ``DetectIcmpSeqMatchTest01`` unit test it said that sid 2 should not alert so I thought its rule doesn't need a check, but instead I need to include the check with the count being 0, right?

jufajardini: Yep, that's it :)

PR #987, opened 2022-10-28
createst: add midstream arg - v9

jufajardini:
Add optional argument '--midstream' to add that as an argument to test.yaml.
Default is still midstream false.

usage:
`createst.py [--midstream]`

before:
```
args:
- -k none
```

after:
```
args:
- -k none
- --set stream.midstream=true
```

Previous PR: #882 

Update from Last PR: rebased, adjusted README file to (hopefully) avoid merging conflicts.

PR #991, opened 2022-10-28
createst: add midstream arg - v10

jufajardini:
Add optional argument '--midstream' to add that as an argument to test.yaml.
Default is still midstream false.

usage:
createst.py [--midstream]

before:
```
args:
- -k none
```
after:
```
args:
- -k none
- --set stream.midstream=true
```
Previous PR: https://github.com/OISF/suricata-verify/pull/987

Update from Last PR: rebased.


jasonish:
Merged. Thanks!

PR #992, opened 2022-10-28
detect-bytemath: add tests

hsadia538:
Redmine [Ticket](https://redmine.openinfosecfoundation.org/issues/5589)

Conversion:
- detect-bytemath-03: `DetectByteMathPacket01` from detect-bytemath.c
- detect-bytemath-04: `DetectByteMathPacket02`  from detect-bytemath.c


hsadia538:
This PR is testing multiple operations and is based of off two tests in ` detect-bytemath.c` but I think it can be combined in one.
Let me know if its the right approach.

tests/detect-bytemath-04/README.md
```diff
@@ -0,0 +1,11 @@
+Description
+===========
+Test byte_match payload keyword.
```
inashivb, line 3: What's the difference between both the tests?

hsadia538: I have tested (-) in one (+) in other but I think I can combine them in one and increase the number of rules?

jlucovsky: IMO, it's ok to combine tests `-03` and `-04` provided the rule count doesn't get "too high". 
It's also ok to keep the tests separate -- if they remain separate, I'd suggest changing the test directory name slightly to `detect-bytemath-sub-03` and `detect-bytemath-add-04`

tests/detect-bytemath-03/test.rules
```diff
@@ -0,0 +1,9 @@
+alert tcp any any -> any any (ipv4.hdr; byte_extract: 1, 0, extracted_val, relative;byte_math: bytes 1, offset 1, oper -, rvalue extracted_val, result var;byte_test: 1, <, var, 9;msg:\
+"Byte extract and byte math with byte test verification\
+";sid:1;)
+alert tcp any any -> any any (ipv4.hdr; byte_extract: 1, 0, extracted_val, relative;byte_math: bytes 1, offset 1, oper -, rvalue extracted_val, result var;byte_test: 1,!=, var, 9;msg:\
+"Byte extract and byte math with byte test verification\
+";sid:2;)
+alert tcp any any -> any any (ipv4.hdr; byte_extract: 1, 0, extracted_val, relative;byte_math: bytes 1, offset 0, oper -, rvalue extracted_val, result var;byte_test: 1,>, var, 0;msg:\
+"Byte extract and byte math with byte test verification\
+";sid:3;)
```
jlucovsky, line 9: byte_test works with unsigned values; the first rule:
- (byte_extract )Value 69 is extracted
-  (byte_match) 34 - 69 = -35
- (byte_test) 6 < -35 (where -35 is represented in a `uint64_t` value)

If the intent was to have unsigned values compared than this is ok. If the intent was to compare 6 and -35 as signed values, then the rule would require adjustment to avoid the negative value computation.

The other rules are fine:
- (2nd rule) 69 is extracted, byte_math computes -35 and then -35 is tested for inequality with 6
- (3rd rule) 69 extracted; byte_math computes 0 and then the check of 69 > 0 succeeds.

tests/detect-bytemath-03/test.yaml
```diff
@@ -0,0 +1,19 @@
+requires:
+  min-version: 7
```
jlucovsky, line 2: `byte_math` has been around for a bit so checking for the byte_math source file could be used instead of a version:
```
requires:
   files:
       - src/detect-bytemath.c
    
```

tests/detect-bytemath-03/test.rules
hsadia538: I was not comparing unsigned values.
So no changes are needed.
Based on comments I will be renaming the directories instead of combining the tests.
Thank you for the review!

hsadia538:
Followed by #994

PR #996, opened 2022-11-01
tests: update pcre test for new output

victorjulien:
Disable on 6 until we can do version checks per shell check.


victorjulien:
Merged after rebase. @inashivb I did this one before #999 was available. Could you update this test to use that and make sure it works in both 6 and 7?

inashivb:
Replaced w #1015

PR #983, opened 2022-10-28
createst: add exclude-fields option - v2

jufajardini:
This is a continuation of PR #342 by @TharushiJay 

Changes from last PR:
- rebase
- update the readme to include the new argument
- update the commit message as per the review (include before and after results in the commit message)
- include ticket number in original commit message

PR #342, opened 2020-10-18
createst: add exclude-fields option

TharushiJay:
Adds the --exclude-fields option to exclude specified fields from the filter blocks.

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4062

`createst.py mytest mypcap --exclude-fields dest_port, src_port`

Before : 
```
requires:
  min-version: 5.0.0
  features:
    - HAVE_LIBJANSSON

args:
 - -k none

checks:
- filter:
    count: 1
    match:
      alert:
        action: allowed
        category: access to a potentially vulnerable web application
        gid: 1
        rev: 1
        severity: 2
        signature: no1
        signature_id: 9000000
      app_proto: http
      dest_ip: 10.100.0.8
      dest_port: 44270
      event_type: alert
      http:
        hostname: www.abcdefghij.com
        http_content_type: text/html
        http_method: GET
        http_refer: http://www.abcdefghij.com/abdeltat/login
        http_user_agent: Mozilla/5.0 (X11; U; Linux i686; fr; rv:1.9.0.6) Gecko/2009011912
          Firefox/3.0.6
        length: 1483
        protocol: HTTP/1.1
        status: 401
        url: /publication/pub.home/home.html
      pcap_cnt: 14
      proto: TCP
      src_ip: 162.2.41.200
      src_port: 80
```

After : 
```
requires:
  min-version: 5.0.0
  features:
    - HAVE_LIBJANSSON

args:
 - -k none

checks:
- filter:
    count: 1
    match:
      alert:
        action: allowed
        category: access to a potentially vulnerable web application
        gid: 1
        rev: 1
        severity: 2
        signature: no1
        signature_id: 9000000
      app_proto: http
      dest_ip: 10.100.0.8
      event_type: alert
      http:
        hostname: www.abcdefghij.com
        http_content_type: text/html
        http_method: GET
        http_refer: http://www.abcdefghij.com/abdeltat/login
        http_user_agent: Mozilla/5.0 (X11; U; Linux i686; fr; rv:1.9.0.6) Gecko/2009011912
          Firefox/3.0.6
        length: 1483
        protocol: HTTP/1.1
        status: 401
        url: /publication/pub.home/home.html
      pcap_cnt: 14
      proto: TCP
      src_ip: 162.2.41.200
```

PR #997, opened 2022-11-01
createst: add exclude-fields option - v3

jufajardini:
This is a continuation of PR https://github.com/OISF/suricata-verify/pull/342 by @TharushiJay

Last PR: #983 

Changes from last PR:
- rebase


jasonish:
Non top-level fields don't work, for example:
```
--exclude-fields flow.src_port
```
does not exclude that field from the filter.  I took a quick stab at it, but get lost as soon as I see Python decorators.  But the fix is probably a refactor to check the fully assembled key names against the exclusion list.

jufajardini:
Closing this for now as it needs re-work and I haven't been able to look into it in a while...

PR #1000, opened 2022-11-03
smtp: more valid SMTP for protocol-changed test

catenacyber:
That is one server banner, plus one response per request


catenacyber:
See https://github.com/OISF/suricata/pull/8141

catenacyber:
Merged in #1013

PR #1002, opened 2022-11-12
Feat 5630/v2

inashivb:
- #999 
- #996

### DIFF
--- a/run.py
+++ b/run.py
@@ -395,13 +395,26 @@ class FileCompareCheck:
 
 class ShellCheck:
 
-    def __init__(self, config, env):
+    def __init__(self, config, env, suricata_config):
         self.config = config
         self.env = env
+        self.suricata_config = suricata_config
 
     def run(self):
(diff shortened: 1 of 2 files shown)


catenacyber:
Should CI be green ?

inashivb:
> Should CI be green ?

I think it will be if we merge the framework patch (#999) and then test the test.

catenacyber:
> > Should CI be green ?
> 
> I think it will be if we merge the framework patch (#999) and then test the test.

This PR has both and fails CI.
Is #996 waiting for a suricata PR ?

inashivb:
> > > Should CI be green ?
> > 
> > 
> > I think it will be if we merge the framework patch (#999) and then test the test.
> 
> This PR has both and fails CI. Is #996 waiting for a suricata PR ?

I'm rechecking. No Suricata PR is needed, I think. I must have misunderstood the expected behavior.

PR #1006, opened 2022-11-16
Add test for brotli content encoding

dcbeckett:
Currently Suricata 6 and Suricata 7.0.0-beta1 fails this HTTP/1.1 test as it doesn't support brotli encoding
Feature #5692


catenacyber:
cc @cccs-rmorti ;-)

catenacyber:
David, I do not think this feature will come in the short term.
I think it will come after HTTP1 parsing has been moved to rust...

cccs-rtmorti:
> cc @cccs-rtmorti ;-)

This doesn't look like it will be too hard to add, since there is already a nice crate for it: https://crates.io/crates/brotli

catenacyber:
And brotli crate is already used in HTTP2 ;-)

catenacyber:
Rebased in https://github.com/OISF/suricata-verify/pull/2441

PR #1008, opened 2022-11-17
app-layer-template: only for version < 7

jasonish:
The C templates have been removed from 7.0.

### DIFF
--- a/tests/app-layer-template/test.yaml
+++ b/tests/app-layer-template/test.yaml
@@ -5,6 +5,7 @@ args:
 
 checks:
 - filter:
+    lt-version: 7
     count: 1
     match:
       dest_ip: 10.16.1.10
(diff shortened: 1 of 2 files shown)


tests/app-layer-template/test.yaml
```diff
@@ -5,6 +5,7 @@ args:
 
 checks:
 - filter:
+    lt-version: 7
```
catenacyber, line 8: It looks to me that we have twice the same test for `lt-version: 7` and `min-version: 7`
What am I missing ?

catenacyber:
> The C templates have been removed from 7.0.

Have they already ?

tests/app-layer-template/test.yaml
jasonish: `template.request` and `template.response` differ.

jasonish:
> > The C templates have been removed from 7.0.
> 
> Have they already ?

Premature optimization.  Still requires the PR.

catenacyber:
PR is open https://github.com/OISF/suricata/pull/8182

PR #1009, opened 2022-11-18
tests: add tests to verify decoding of padded UDP Ethernet packets

lukashino:
Prepare tests for [Redmine ticket 5693](https://redmine.openinfosecfoundation.org/issues/5693)


lukashino:
I think as the solution around shorter UDP packets changed, this is no longer needed.

PR #1022, opened 2022-11-30
rfb: adds a check for community_id field in a rfb event

catenacyber:
Should be good for master, and not for master6


catenacyber:
Requires https://github.com/OISF/suricata/pull/8226

victorjulien:
Merged in #1030, thanks.

PR #1023, opened 2022-11-30
http: check that custom logging does not collide

catenacyber:
for content range header for instance

I do not know if this will be backported as this is a breaking change of something broken...


catenacyber:
Requires https://github.com/OISF/suricata/pull/8227

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1026

PR #1026, opened 2022-12-02
http: check that custom logging does not collide

catenacyber:
for content range header for instance

Replaces #1023 with new naming for log fields


catenacyber:
Suricata PR was merged

jasonish:
Merged with https://github.com/OISF/suricata-verify/pull/1243.

PR #1027, opened 2022-12-02
smtp: do not check for empty events

catenacyber:
This will be needed by https://github.com/OISF/suricata/pull/8230

### DIFF
--- a/tests/smtp-file-data-01/test.yaml
+++ b/tests/smtp-file-data-01/test.yaml
@@ -59,15 +59,3 @@ checks:
       src_ip: 1.1.205.22
       src_port: 4053
       tx_id: 0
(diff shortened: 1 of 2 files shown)


victorjulien:
What is the purpose here?

catenacyber:
> What is the purpose here?

This will be needed by https://github.com/OISF/suricata/pull/8230 that is the mime rust parser cf https://redmine.openinfosecfoundation.org/issues/3487

Current master creates an empty SMTP transaction after one is finished and these S-V checks (createst) enforce this behavior. My current rust mime parser does not make SMTP create this empty transaction...

catenacyber:
Seems unneeded anymore cf #1035

PR #1035, opened 2022-12-08
body_md5 : take every line into account,

catenacyber:
also lines that are headers of sub mime parts

and do not take new line post boundary

### DIFF
--- a/tests/smtp-long-DATA-line/test.yaml
+++ b/tests/smtp-long-DATA-line/test.yaml
@@ -30,7 +30,6 @@ checks:
       dest_ip: 192.168.1.4
       dest_port: 3326
       event_type: anomaly
-      pcap_cnt: 40
       proto: TCP
       src_ip: 217.12.11.66
       src_port: 587
@@ -45,7 +44,6 @@ checks:
       dest_ip: 192.168.1.4
       dest_port: 3326
       event_type: anomaly
-      pcap_cnt: 40
       proto: TCP
       src_ip: 217.12.11.66
       src_port: 587
--- a/tests/smtp-md5/test.yaml
+++ b/tests/smtp-md5/test.yaml
@@ -13,7 +13,7 @@ checks:
         smtp.mail_from: <gurpartap@patriots.in>
         smtp.rcpt_to[0]: <raj_deol2002in@yahoo.co.in>
         email.subject_md5: "c2239a92bde29f0a9f9173193cc2fe00"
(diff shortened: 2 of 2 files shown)


catenacyber:
@jasonish thoughts about this behavior change ?

catenacyber:
cf https://github.com/OISF/suricata/pull/8450

catenacyber:
> @jasonish thoughts about this behavior change ?

ping ?

jasonish:
> > @jasonish thoughts about this behavior change ?
> 
> ping ?

I think I need more discussion of why the change in md5.. Determine what may be more correct.  My expectation is we'd parse the headers, then hash the bytes to the end of the file.  Is that new behavior?

catenacyber:
> 
> I think I need more discussion of why the change in md5.. Determine what may be more correct. My expectation is we'd parse the headers, then hash the bytes to the end of the file. Is that new behavior?

Yes, this is the new behavior.
Difference with old behavior is when we have multiple/mixed recursive MIME.
The global md5 does not take into account the internal headers

In the example, we have something like

```
MIME Multipart Media Encapsulation, Type: multipart/mixed, Boundary: "----=_NextPart_000_0004_01CA45B0.095693F0"
- Preamble: 546869732069732061206d756c746970617274206d65737361676520696e204d494d4520…
- Encapsulated multipart part:  (multipart/alternative)
- Encapsulated multipart part:  (text/plain)
```

The `email.body_md5` must take into account the mime headers of the encapsulated parts

jasonish:
> > I think I need more discussion of why the change in md5.. Determine what may be more correct. My expectation is we'd parse the headers, then hash the bytes to the end of the file. Is that new behavior?
> 
> Yes, this is the new behavior. Difference with old behavior is when we have multiple/mixed recursive MIME. The global md5 does not take into account the internal headers
> 
> In the example, we have something like
> 
> ```
> MIME Multipart Media Encapsulation, Type: multipart/mixed, Boundary: "----=_NextPart_000_0004_01CA45B0.095693F0"
> - Preamble: 546869732069732061206d756c746970617274206d65737361676520696e204d494d4520…
> - Encapsulated multipart part:  (multipart/alternative)
> - Encapsulated multipart part:  (text/plain)
> ```
> 
> The `email.body_md5` must take into account the mime headers of the encapsulated parts

That sounds more correct to me, so I'm in favour of the change. Sounds like the previous behaviour was attempting to be smart.

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1252

PR #1036, opened 2022-12-09
tests: add tests for exception policy master switch - v1

jufajardini:
Add tests for the new master switch config for the Exception policies.

Tests for:

- 'auto' behavior (drop-packet/drop-flow)
- 'performance' (bypass)
- disabled
- overwrite setting with Exception policy set

I think it could be a good idea to add another test for when the master switch is disabled and the exception policy isn't present.

Task #5219


jufajardini:
Add min-version requirement to all tests.

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1037

PR #1040, opened 2022-12-14
tests: add tests for exception policy master switch - v4

jufajardini:
Previous PR: #1039

Changes from previous PR:
- update tests that used  'disabled' in favour of the already existing 'ignore'

PR #1044, opened 2022-12-19
tests: add tests for exception policy master switch - v5

jufajardini:
Previous PR: #1040 

Changes from last PR:
- add drop.reason to the first test, remove `set stream.midstream=true` to avoid confusion about what the test is about.

Suricata PR: https://github.com/OISF/suricata/pull/8287


catenacyber:
Does this require a new Suricata PR Juliana ?

catenacyber:
Is this related to #1178 ?

jufajardini:
> Is this related to #1178 ?

No, this is for testing the master configuration for the exception policies (to set the same value to all different exception policies from one place). But we're missing backporting this, due to a bug in the midstream exception policy (latest PR for that: #9011)

jufajardini:
rebased and replaced: https://github.com/OISF/suricata-verify/pull/1264

PR #1045, opened 2022-12-22
rfb: test rfb frames

hsadia538:
Ticket: https://redmine.openinfosecfoundation.org/issues/5717

- Test rfb Pdu frames
Requires: https://github.com/OISF/suricata/pull/8313

PR #1048, opened 2022-12-26
rfb: test rfb frames

hsadia538:

Previous PR: #1045 

Describe new changes:
- Update `test.yaml `and `test.rules`

Requires: https://github.com/OISF/suricata/pull/8337


tests/rfb-frames/test.rules
```diff
@@ -0,0 +1,3 @@
+alert rfb any any -> any any (msg:"RFB Frame 1" ;flow:to_client; frame:pdu; content:"|01 02|"; startswith; sid:1;)
+alert rfb any any -> any any (msg:"RFB Frame 2"; flow:to_server; frame:pdu; content:"|7A 6F 36|"; offset:2; sid:2;)
+alert rfb any any -> any any (msg:"RFB Frame 3"; frame:pdu; content:"|30 30 38 0A|"; endswith; sid:3;)
```
inashivb, line 3: Could we also have some not matching rules? e.g. content has more data that follows that packet frame. Should something like that be a match?
Others like not giving an offset but content is something from the middle of the pdu, etc.

inashivb (changes requested):
Looks good at the first sight. Might be good to add some rules that won't alert and add checks corresponding to them.

tests/rfb-frames/test.rules
hsadia538: Do you mean adding the not matching rules and saying this rule should have 0 matching packets in the yaml file?

hsadia538: > e.g. content has more data that follows that packet frame. Should something like that be a match?
Do you mean a wrong field of content?

inashivb: Yes.

hsadia538:
Followed by #1050

PR #1050, opened 2023-01-02
rfb: test rfb frames

hsadia538:

Previous PR: #1048 

Describe new changes:
- Update `test.yaml `and `test.rules`
- Add new rules with 0 match count.

Requires: https://github.com/OISF/suricata/pull/8337


jufajardini (changes requested):
This is looking good, kudos! 

I think we can add some checks about the frame event type, like it's done with 1c1ffcf3bb0812, to exemplify usage - and even check from frame type, in case we end up adding more frames to RFB :)

hsadia538:
> 1c1ffcf

Do you mean how the rule tests event types like REGISTER and INVITE? 
I am not sure if its possible to do the same for rfb. As far as I have observed in SIP pcaps, REGISTER is part of the packet but IN RFB, the state is determined based on packet data after processing and not the actual values so it will not be possible to find a specific state in a packet(I might be wrong)

hsadia538:
> and even check from frame type, in case we end up adding more frames to RFB :)

are you referring to request.line, request.header in SIP? I am not sure what you mean by checking from frame types?

jufajardini:
> 

See if the other comment answers this one?

PR #1054, opened 2023-01-04
tests: add test for bug 5379

inashivb:
Should fail on master and 6.0.x for now.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5379


inashivb:
Interesting. Was expecting it to fail on 6..

PR #1058, opened 2023-01-06
tests: filestore alert event

catenacyber:
Replaces #965 using `storing` instead of `stored`


inashivb (commented):
Looking good and indeed much less confusing.

Could we add a README to cite the source of pcap and any other important info?

catenacyber:
> Could we add a README to cite the source of pcap and any other important info?

Done in https://github.com/OISF/suricata-verify/pull/1143

PR #1060, opened 2023-01-10
Stats flow ctr 5769 v1

catenacyber:
Requires https://github.com/OISF/suricata/pull/8328


catenacyber:
Suricata PR was merged

check-eve.py
```diff
@@ -40,6 +40,23 @@ def validate_json(args, json_filename, schema):
...
+            if "stats" in obj:
+                d = obj["stats"]["app_layer"]["flow"]
+                stats = {}
+                for proto in d:
+                    stats[proto] = d[proto]
+                for k in flows:
+                    if k not in stats:
+                        errors.append("Different values for %s : %d flows and %d in stats" % (k, flows[k], 0))
+                    elif stats[k] != flows[k]:
+                        errors.append("Different values for %s : %d flows and %d in stats" % (k, flows[k], stats[k]))
```
jasonish, line 58: It looks like these are occurring, but never printed or cause failure.  Probably the "status" var has to be updated?

Might make sense to break out to its own function with a bit of function doc to says whats going on as well.

catenacyber: Indeed

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1061

PR #1061, opened 2023-01-11
framework: consistency check for flow app-layer stats

catenacyber:
Replaces #1060 with really erroring if needed

PR #1062, opened 2023-01-11
framework: consistency check for flow app-layer stats

catenacyber:
Replaces #1061 by taking care that `dns` flows are split in stats in `dns_udp` and `dns_tcp`

### DIFF
--- a/check-eve.py
+++ b/check-eve.py
@@ -36,10 +36,45 @@ try:
 except:
     HAVE_PY = False
 
+def stat_normalize_proto(p):
+    if p.endswith("_tcp"):
+        return p[:-4]
+    if p.endswith("_udp"):
+        return p[:-4]
+    return p
+
+def validate_json_stats(json_filename):
+    errors = []
+    with open(json_filename) as f:
+        flows = {}
+        for line in f:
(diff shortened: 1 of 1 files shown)


catenacyber:
Interesting : http-connect-simple counts 2 http flows in stats because of protocol detection being run again...
Thoughts in this ?

catenacyber:
@jasonish do you think 
`jq 'select(.event_type=="flow" and .app_proto=="ftp") | .app_proto'  log/eve.json | wc -l` and `jq 'select(.event_type=="stats") | .stats."app_layer".flow.ftp' log/eve.json` should always yield the same results ? Also when there is a protocol change ?

jasonish:
> @jasonish do you think `jq 'select(.event_type=="flow" and .app_proto=="ftp") | .app_proto' log/eve.json | wc -l` and `jq 'select(.event_type=="stats") | .stats."app_layer".flow.ftp' log/eve.json` should always yield the same results ? Also when there is a protocol change ?

My guess is probably, but maybe some edge cases where not?

catenacyber:
So, closing this as what we want to check is not universally true...

PR #1064, opened 2023-01-13
mqtt: test mqtt frames

hsadia538:
Ticket: https://redmine.openinfosecfoundation.org/issues/5731

Draft PR to gather feedback

Describe changes:
- tests mqtt frames

PR #1065, opened 2023-01-16
mqtt: test mqtt frames v2

hsadia538:
Ticket: https://redmine.openinfosecfoundation.org/issues/5731

Previous PR: #1064 

Describe changes:
- Add tests for mqtt frames

Requires: https://github.com/OISF/suricata/pull/8382


jufajardini (approved):
For this first set of frame types, this is looking good, to me :)

jufajardini:
In light of the last discussions around truncated messages, I think it would be good to have a test with frames and those truncated types, so we could see what are the frames registering in such a case. :)

jufajardini (changes requested):
Please, see my comment :)

hsadia538:
> In light of the last discussions around truncated messages, I think it would be good to have a test with frames and those truncated types, so we could see what are the frames registering in such a case. :)

Ok I will incorporate this in a new SV PR and then you will re-run CI checks for this one?

jufajardini:
> > In light of the last discussions around truncated messages, I think it would be good to have a test with frames and those truncated types, so we could see what are the frames registering in such a case. :)
> 
> Ok I will incorporate this in a new SV PR and then you will re-run CI checks for this one?

I think that can work. If not, we'd need a new Suricata PR.

PR #1070, opened 2023-01-20
Bug 5379/v3

inashivb:
Should fail on master and 6.0.x for now.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5379

Suri fix: https://github.com/OISF/suricata/pull/8410

Previous PR: https://github.com/OISF/suricata-verify/pull/1055

Changes since v2:
- Add test for TCP as well


tests/tcp-5379/test.yaml
```diff
@@ -0,0 +1,53 @@
+args:
+- -k none
+
+checks:
```
victorjulien, line 4: needs a check for making sure sid 2 did not match

PR #1073, opened 2023-01-22
tests: smb2 file sha logging test

victorjulien:


tests/smb2-09-trunc-file-logging/suricata.yaml
```diff
@@ -0,0 +1,201 @@
+%YAML 1.1
```
catenacyber, line 1: Do we need a big suricata.yaml ?

PR #1074, opened 2023-01-24
tls: add test for tls keywords v1

hsadia538:
Description:
Add tests for tls keywords: `tls.subject ` `tls.issuerdn`


tests/tls-issuerdn/test.yaml
```diff
@@ -0,0 +1,17 @@
+requires:
```
jufajardini, line 1: I recently learned that when we want to use a pcap from a pre-existing sv test, we can also add:
```yaml
pcap: ../existing-test-name/pcap-file.pcap
```
:)

jufajardini: As seen in: https://github.com/OISF/suricata-verify/blob/2eca390bdd825fd7c709a18ef0884146c3cb4c7e/tests/detect-udp-flow-rule-01/test.yaml

tests/tls-issuerdn/test.yaml
```diff
@@ -0,0 +1,17 @@
...
+  features:
+    - HAVE_LIBJANSSON
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: tls
+        tls.session_resumed: true
+
```
jufajardini, line 11: Suggestion: let's `tls.issuerdn` here, as well, as this also helps to showcase how the output is the same, regardless of using the legacy keyword or the more recent one. ;)

tests/tls-subject/test.yaml
```diff
@@ -0,0 +1,16 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: tls
+        tls.session_resumed: true
```
jufajardini, line 10: Similar idea as above: let's check for `tls.suject` here :)

jufajardini (changes requested):
We're on the right path, good job!

Some inline comments, but just to improve on the work you've done.
Partly related: while reviewing this PR I noticed that at least the current `tls` keywords for these legacy keywords are not covered by suricata-verify tests. Could you please also add tests for those, and check if we are missing any other in the process?
The ones I checked were: `tls.cert_issuer` and `tls.cert_subject`.

The commit messages need two minor fixes:
- they are missing the blank line between the commit subject and the commit body. (if you open the branch in tig, you'll notice that the `Ticket` part is appended to the commit title, due to the lack of that extra line). 
- let's also go back to the `Ticket #xxx` format, as this is more according to our guidelines :)

hsadia538:
> We're on the right path, good job!
> 
> Some inline comments, but just to improve on the work you've done. Partly related: while reviewing this PR I noticed that at least the current `tls` keywords for these legacy keywords are not covered by suricata-verify tests. Could you please also add tests for those, and check if we are missing any other in the process? The ones I checked were: `tls.cert_issuer` and `tls.cert_subject`.
> 
Do you want me to add the tests for those in the same commits?

tests/tls-issuerdn/test.yaml
hsadia538: so you mean I dont need to give the pcap at all and reference it here ?

hsadia538: Yes it works. thanks

hsadia538:
Followed by #1075

PR #1083, opened 2023-01-25
Add check for duplicate JSON keys - v1

jasonish:
Add a new script to test for duplicate JSON keys.

TODO: Hook info normal S-V runs somehow.

### DIFF
--- a/duplicate-key-check.py
+++ b/duplicate-key-check.py
@@ -1,0 +1,42 @@
+#! /usr/bin/env python3
+#
+# Check JSON objects for duplicate keys.
+#
+# NOTE: This will only report the first duplicate key found in a line
+# of JSON. More duplicates may be found after fixing the first.
+
+import json
+import sys
+
+class DuplicateKeyError(Exception):
+    pass
+
+def validate_file(filename):
+    with open(filename) as json_in:
+        for line in json_in:
+            json.loads(line, object_pairs_hook=hook)
+
+def main():
+    failures = 0
+    for filename in sys.argv[1:]:
+        try:
+            validate_file(filename)
+        except DuplicateKeyError as err:
+            print("ERROR: Duplicate key: {}: {}".format(filename, err))
+        except Exception as err:
+            print("ERROR: Unknown: {}: {}".format(filename, err))
+            failures += 1
+
+    return failures
+
+def hook(pairs):
+    d = {}
+    for k, v in pairs:
+        if k in d:
+            equal = d[k] == v
+            raise DuplicateKeyError("key={}, equal={} current value={}, new value={}".format(k, equal, d[k], v))
+        d[k] = v
+    return d
+
+if __name__ == "__main__":
+    sys.exit(main())


victorjulien:
I'm fine with this as a first step, but really feel it should be managed by `run.py` so we don't need to do the additional step of running this check. Probably needs master-6.0.x to be clean first.

jasonish:
> I'm fine with this as a first step, but really feel it should be managed by `run.py` so we don't need to do the additional step of running this check. Probably needs master-6.0.x to be clean first.

Yeah, makes sense.  Probably as part of eve-validation.  Will come back to that soon, integrating properly is a bigger job than creating this script! :)

catenacyber:
It is no longer requires suricata fix, but requires backports, right ?

jasonish:
> It is no longer requires suricata fix, but requires backports, right ?

Correct.  Lets just put this on hold til after these releases are out.

catenacyber:
> Correct. Lets just put this on hold til after these releases are out.

Ok.
By the way, why is CI green if a backport is still missing ?

jasonish:
> > Correct. Lets just put this on hold til after these releases are out.
> 
> Ok. By the way, why is CI green if a backport is still missing ?

This script isn't hooked into this CI yet, for now its just a utility script.  My original PR message is in error.

catenacyber:
Should not jsonschema tell us about duplicate keys ?

catenacyber:
> No, jsonschema runs on the data after loaded by Python. The Python loader throws out duplicate keys.

jsonschema runs on the file directly cf https://github.com/OISF/suricata-verify/blob/master/eve-validator/src/main.rs#L52

Now that it is the fast rust version instead of the slow python one

jasonish:
> Can we disallow dups for eve anyway? It's almost certainly unintented?

Verification is probably the best way (ie: this PR),  It would be hard to do with JsonBuilder given we're not inserting to a map that might know what keys are already there, so its state tracking would have to become much more advanced.

jasonish:
aaaasdfffttrreewwqqzzxxcVVVVR

> > Can we disallow dups for eve anyway? It's almost certainly unintented?
> 
> Verification is probably the best way (ie: this PR), It would be hard to do with JsonBuilder given we're not inserting to a map that might know what keys are already there, so its state tracking would have to become much more advanced.

Alternative answer.. Of course, its just software, but not a quick or easy change to JsonBuilder.

victorjulien:
Right, I was thinking about verification stage indeed.

catenacyber:
And so, verification with serde would not tell us about duplicate keys ?

jasonish:
> And so, verification with serde would not tell us about duplicate keys ?

Not easily.  Using serde to decode to a generate `serde_json::Value` will not show these errors.

If you create structs for all possible data types and derive the serde implementation, it will error out on duplicate keys.  Maybe we could generate a struct from the schema and deserialize every line to it?  I'm not sure how robust the existing, if any jsonschema -> rust code gen tools there are though.

catenacyber:
Oh, so python errors on duplicate keys but not rust `serde_json` 😢 

So, this is a good idea

jasonish:
> Oh, so python errors on duplicate keys but not rust `serde_json` cry

Python doesn't either. This script hooks into the deserializer to raise the error.

duplicate-key-check.py
```diff
@@ -0,0 +1,42 @@
...
+            print("ERROR: Unknown: {}: {}".format(filename, err))
+            failures += 1
+
+    return failures
+
+def hook(pairs):
+    d = {}
+    for k, v in pairs:
+        if k in d:
+            equal = d[k] == v
```
catenacyber, line 36: This seems redundant with the warning message below

duplicate-key-check.py
```diff
@@ -0,0 +1,42 @@
...
+        for line in json_in:
+            json.loads(line, object_pairs_hook=hook)
+
+def main():
+    failures = 0
+    for filename in sys.argv[1:]:
+        try:
+            validate_file(filename)
+        except DuplicateKeyError as err:
+            print("ERROR: Duplicate key: {}: {}".format(filename, err))
```
catenacyber, line 25: Should we increase failures here as well ?

duplicate-key-check.py
```diff
@@ -0,0 +1,42 @@
...
+def hook(pairs):
+    d = {}
+    for k, v in pairs:
+        if k in d:
+            equal = d[k] == v
+            raise DuplicateKeyError("key={}, equal={} current value={}, new value={}".format(k, equal, d[k], v))
+        d[k] = v
+    return d
+
+if __name__ == "__main__":
```
catenacyber, line 41: Who calls this ?

duplicate-key-check.py
jasonish: Ah, yeah.. Debugging.

duplicate-key-check.py
jasonish: We'd have to hook it in, this is just a script for detecting duplicates.

catenacyber:
Tests pass because the new script is not hooked in yet

victorjulien:
> Tests pass because the new script is not hooked in yet

Updated my script to exclude drafts.

catenacyber:
@jasonish what is the status if this ?
This looks a good idea, what is missing before merging ?

jasonish:
Replaced by https://github.com/OISF/suricata-verify/pull/1572.

PR #1076, opened 2023-01-24
tls: add test for tls keywords v3

hsadia538:
Previous PR : #1075 
Describe Changes:

- Update `test.rules` to fix build error

PR #1084, opened 2023-01-25
tls: add test for tls keywords v4

hsadia538:
Previous PR : https://github.com/OISF/suricata-verify/pull/1076
Describe Changes:

- Update `test.rules` to fix typo

### DIFF
--- a/tests/tls-issuerdn/README
+++ b/tests/tls-issuerdn/README
@@ -1,0 +1,13 @@
+Description
+===========
+Tests the `tls.issuerdn` legacy keyword
+Tests the `tls.cert_issuer` new keyword
+Both represent TLS/SSL certificate IssuerDN field
+
+PCAP
+====
+PCAP comes from an [existing TLS test](https://github.com/OISF/suricata-verify/blob/master/tests/tls/tls.pcap)
+
+Redmine ticket
+==============
+https://redmine.openinfosecfoundation.org/issues/5544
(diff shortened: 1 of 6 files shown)


hsadia538:
Not sure if its related or not but the build failure is due to the test `test-bad-hex-rule-2`

hsadia538:
After incorporating feedback from last PR , I added tests for the new keywords that replace the legacy one. Should I change the commit message to reflect that?

catenacyber:
Is this expecting a Suricata PR to pass ?

jufajardini:
> Is this expecting a Suricata PR to pass ?

Shouldn't be, as this is just increasing testing coverage for existing keywords...

jufajardini:
> After incorporating feedback from last PR , I added tests for the new keywords that replace the legacy one. Should I change the commit message to reflect that?

sorry for taking so long to answer D:
Better to, yes. :)
It is ok to have that in the commit body, if it doesn't fit in the title.

jufajardini:
> Not sure if its related or not but the build failure is due to the test `test-bad-hex-rule-2`

I've triggered a re-run for that failure. If this still fails, I would suggest rebasing and submitting a new PR - which would also work well since you're about to change the commit messages, too :)

hsadia538:
I see that the checks have passed

hsadia538:
Followed by #1106

PR #1086, opened 2023-01-26
detect-lua: add tests

TheKharleeci:
Includes test for LuaMatchTest01-LuaMatchTest06

Previous PR: https://github.com/OISF/suricata-verify/pull/688


jufajardini (changes requested):
I have pointed out some aspects that need more work on, and which I think would allow us to at least have alerts for all rules.

If ensuring that the pcap has both packets `to_server` isn't enough to trigger all rules, I suggest maybe using a different pcap - maybe even using one from some existing SV test. :)

Would be great if we could manage to have the alert checks also check the packet number, but not sure how easy that would be - I tried a bit, but the pcaps I tried didn't have `pcap_cnt` fields in the alert outputs, so maybe we can ignore it by now...

inashivb:
Replaced  w https://github.com/OISF/suricata-verify/pull/1165

PR #1092, opened 2023-01-27
Add command line options for generating JUnit xml and copying output files of failed tests

TAnhMinh:
The option --generate-xml will cause generation of a JUnit xml file summarizing the result of all run tests.
The option --copy-failed will copy output files of all failed tests into the 'failed' directory.


jasonish:
This pull-request looks like a mix of private changes that should not have escaped a private Gitlab instance mixed perhaps with stuff of use in general?

I wonder if this could be limited to minor changes that would then allow you to have custom scripts to post-process into your desired XML output and integrate with your CI?

TAnhMinh:
Yes, you are right, thank you. I divided the changes into two new merge branches: lua-require and new-cmd-options, There are respective pull requests.

PR #1095, opened 2023-01-30
Adds test about http mime with truncated file

catenacyber:
due to request.body_limit configuration value

Requires https://github.com/OISF/suricata/pull/8474

@jasonish can you explain me why it is `outputs.1.eve-log.types.2.files.force-hash=sha256` ? (I mean the 1 and 2)


jasonish:
> @jasonish can you explain me why it is `outputs.1.eve-log.types.2.files.force-hash=sha256` ? (I mean the 1 and 2)

`outputs` and `types` are arrays so we have to provide the index into the array to change the value of.

catenacyber:
Yes, but why index 1 and 2, instead of 0 and 0 ?

jasonish:
> Yes, but why index 1 and 2, instead of 0 and 0 ?

`eve-log` is at index `1` or `outputs` (`fast` is at index `0`).  I think the `.2.files.force-hash` is wrong. `files` is at index 6, at least in master.  And 5 in in master-6.0.x.

```
./src/suricata -c ./suricata.yaml --dump-config|grep -e '^outputs.*eve-log.*files'
```
to see how Suricata is interpreting the configuration file.

jasonish:
> > eve-log is at index 1 of outputs
> 
> So `1` and `outputs` are redundant somehow ?
> 
> > I think the .2.files.force-hash is wrong.
> 
> But it works and .1.files.force-hash does not work... thinking And I copied it from another test

Nope. This is rooted probably in a bad YAML layout from the start that we have yet to go back and address.  It might be more clear with a direct translation to JSON.

- Note how ouputs is an array of objects.  So outputs[1] is an object, that has one key of "eve-log".  So in JSON notation, and perhaps how we should have created the `--set` command you would be setting:

```
outputs[1]["eve-log"]["types"][6]["files"]["force-hash"] = "sha256"
```
but our notation, for better or worse is:
```
outputs.1.eve-log.types.6.files.force-hash=sha256
```

```json
{
  "outputs": [
    {
      "fast": {
        "enabled": "yes",
        "filename": "fast.log",
        "append": "yes"
      }
    },
    {
      "eve-log": {
        "enabled": "yes",
        "filetype": "regular",
        "filename": "eve.json",
        "pcap-file": false,
        "community-id": false,
        "community-id-seed": 0,
        "xff": {
          "enabled": "no",
          "mode": "extra-data",
          "deployment": "reverse",
          "header": "X-Forwarded-For"
        },
        "types": [
          {
            "alert": {
              "tagged-packets": "yes"
            }
          },
          {
            "frame": {
              "enabled": "no"
            }
          },
          {
            "anomaly": {
              "enabled": "yes",
              "types": null
            }
          },
          {
            "http": {
              "extended": "yes"
            }
          },
          {
            "dns": null
          },
          {
            "dns": null
          },
          {
            "tls": {
              "extended": "yes"
            }
          },
          {
            "files": {
              "force-magic": "no"
            }
          },
          {
            "smtp": null
          },
          "ftp",
          "rdp",
          "nfs",
          "smb",
          "tftp",
          "ike",
          "dcerpc",
          "krb5",
          "bittorrent-dht",
          "snmp",
          "rfb",
          "sip",
          "quic",
          {
            "dhcp": {
              "enabled": "yes",
              "extended": "no"
            }
          },
          "ssh",
          {
            "mqtt": null
          },
          "http2",
          {
            "pgsql": {
              "enabled": "no"
            }
          },
          {
            "stats": {
              "totals": "yes",
              "threads": "no",
              "deltas": "no"
            }
          },
          "flow"
        ]
      }
    }
]}
```

catenacyber:
Thank you very much Jason for the explanation.
It makes sense now.
I wonder how to fix this bad yaml layout in the future... Is there a ticket for it ?

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1109

jasonish:
> Thank you very much Jason for the explanation. It makes sense now. I wonder how to fix this bad yaml layout in the future... Is there a ticket for it ?

Its very vague, but we have https://redmine.openinfosecfoundation.org/issues/2321.

Its not so straight-forward with the `outputs` though, as we allow multiple `eve-log` sections to be setup with different configurations, so just moving outputs to a map is not an option either.

PR #1098, opened 2023-01-30
s-v features + s-v bug fixes + tests for Bug 5379/v9

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/1097

Changes since v8:
- Add explicit checks for None along with 0


run.py
```diff
@@ -187,16 +187,14 @@ def is_equal(self, a, b):
...
-        if a.minor is not None and b.minor is not None:
-            if a.minor != b.minor:
-                return False
+        if a.minor is not None and b.minor is not None and a.minor >= 0 and b.minor >= 0 and a.minor != b.minor:
+            return False
 
-        if a.patch is not None and b.patch is not None:
-            if a.patch != b.patch:
-                return False
+        if a.patch is not None and b.patch is not None and a.patch >= 0 and b.patch >= 0 and a.patch != b.patch:
```
inashivb, line 196: This makes no sense. It isn't different from the previous solution.

PR #1109, opened 2023-02-02
Adds test about http mime with truncated file

catenacyber:
due to request.body_limit configuration value

Replaces #1095 with using suricata.yaml to set value for outputs


victorjulien:
Merged in #1181, thanks!

PR #1111, opened 2023-02-02
framework: explicit utf-8 encoding for reading json

catenacyber:
As the default encoding is platform dependent

cc @jasonish


jasonish:
Fixes the issue on Debian 9 and CentOS 7 when I test locally. Also running in a dummy PR here: https://github.com/jasonish/suricata/actions/runs/4076505190

jasonish:
Looks all good, will merge.

PR #1114, opened 2023-02-04
log: Updates needed since Suri stops if logs fail to open at startup

jlucovsky:
This PR adjusts existing test cases that now fail due to the fix for 5836. Those changes cause log initialization issues to stop Suricata if they occur at startup.


victorjulien:
Was merged through #1132/#1137

PR #1118, opened 2023-02-09
Detect tls v7

hsadia538:
Previous PR : https://github.com/OISF/suricata-verify/pull/1108
Describe Changes:

- Update `tls` folder structure
- Fix comments
- Fix nit issues

PR #1121, opened 2023-02-10
Detect tls v8

hsadia538:
Previous PR : https://github.com/OISF/suricata-verify/pull/1118
Describe Changes:

- Update `tls` folder `suricata.yaml` files
- Fix README


tests/tls/tls-issuerdn/README
```diff
@@ -0,0 +1,13 @@
+Description
+===========
+Tests the `tls.issuerdn` legacy keyword
+Tests the `tls.cert_issuer` new keyword
+Both represent TLS/SSL certificate IssuerDN field
+
+PCAP
+====
+PCAP comes from an [existing TLS test](https://github.com/OISF/suricata-verify/blob/master/tests/tls-cert-issuer/tls.pcap)
```
inashivb, line 9: Link probably still needs to be fixed.

inashivb (commented):
Looking good, mostly, now. Good job! :)
Some suggestions:
As a general rule of thumb, one should not introduce inconsistencies/bugs and fix them as a part of the same PR unless it makes sense logically or the two ops were performed by diff people.
So, here, for example, the right way would be to first makes the tls folder and move tests there and then add the new tests there. Explanation why it would make better sense:
If I see the commit for `tls.cert_issuer`, it still points to incorrect pcap in the `test.yaml`: https://github.com/OISF/suricata-verify/pull/1121/commits/4355b54aa4d2619808da4cc442edd359e4ed575c
which is later fixed in the same PR as a part of https://github.com/OISF/suricata-verify/pull/1121/commits/04a65894d83c6585297da2090edff852fb7b0d6b
If we were running tests per commit here, this change would be breaking the CI.

hsadia538:
Followed by #1122

PR #1122, opened 2023-02-13
Detect tls v9

hsadia538:
Previous PR : https://github.com/OISF/suricata-verify/pull/1121
Describe Changes:

- Update commit issues

PR #1126, opened 2023-02-17
Detect tls v10

hsadia538:
Previous pr #1122 

Describe Changes:

- Fix pcap path for `tls-random ` in README file


inashivb (commented):
Looking pretty good now. Great job! :)
One nit:
The order of commits is still in reverse (ref: https://github.com/OISF/suricata-verify/pull/1121#pullrequestreview-1295114382). Although it doesn't matter much in this repo, do you want to make it the right order?

Here, logical issue is that we try to reference pcap for https://github.com/OISF/suricata-verify/pull/1126/commits/2cc2a72adc1c5b56c00ecb609279a3571f656ce2 from a path that doesn't exist when this commit is made (the `tls` folder does not exist).
It all works out since the commit to move the tests to folders is also a part of the same PR.

hsadia538:
Followed by #1128

PR #1127, opened 2023-02-17
Filemagic keywords v2

hsadia538:
Previous PR: #1125 
Describe Changes:

- Update README file
- Add test for file.data `prefilter` keyword.

PR #1130, opened 2023-02-20
Filemagic keywords v3

hsadia538:
Previous PR: #1127
Describe Changes:

- Update comments for `file_data prefilter` tests
- Add `test.yaml` to add checks for `filestore` and incorporate stats.

### DIFF
--- a/tests/file-data-prefilter/README.md
+++ b/tests/file-data-prefilter/README.md
@@ -1,0 +1,11 @@
+Description
+===========
+Tests the `prefilter` keyword for `file_data` and `file.data`.
+
+PCAP
+====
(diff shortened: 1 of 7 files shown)


inashivb (changes requested):
Great job! :) Looking good to me, mostly.
Very minor: the commit message indicates that both tests are for file magic. That needs to be updated. We could prefix both with `file:` instead.

hsadia538:
> Great job! :) Looking good to me, mostly. Very minor: the commit message indicates that both tests are for file magic. That needs to be updated. We could prefix both with `file:` instead.

I found the file_data prefilter code logic within the `detec-filemagic.c `file that's why I prefixed it that way.

hsadia538:
followed by #1131

PR #1133, opened 2023-02-27
Fix missing require in .lua files. V1

TAnhMinh:
Without "local io = require("io")", the tests that use .lua files did not work with Lua 5.2.4


jasonish:
This PR needs to be rebased.

PR #1134, opened 2023-02-27
Add command line options for generating JUnit xml and copying output files of failed tests V1

TAnhMinh:
The option --generate-xml will cause generation of a JUnit xml file summarizing the result of all run tests.
The option --copy-failed will copy output files of all failed tests into the 'failed' directory.

You can add suricata-verify-tests.xml and failed/ to .gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ requires:
   # Require that the Suricata version be less than a version.
   lt-version: 6
 
-  # Test is only for this version. For example, 4.0 would match any 4.0
+  # Test is only for this version. For example, 4.0 would match any 4.0 
   # release, but 4.0.3 would only match 4.0.3.
   version: 4.0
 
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -400,26 +404,13 @@ def run(self):
...
-        lt_version = self.config.get("lt-version")
-        if req_version is not None:
-            shell_args["version"] = req_version
-        if min_version is not None:
-            shell_args["min-version"] = min_version
-        if lt_version is not None:
-            shell_args["lt-version"] = lt_version
-        check_requires(shell_args, self.suricata_config)
-
         try:
```
jufajardini, line 414: We don't want to remove these ;)

TAnhMinh: Oh, thank you. That was a mistake - new attempt was made in V2.

README.md
```diff
@@ -147,19 +147,17 @@ checks:
...
+Script to create a test directory with test.yaml for a given PCAP. This
+needs to be run from a valid Suricata source directory.
 
 ### Usage
 ```
 usage: createst.py [-h] [--output-path <output-path>] [--eventtype-only]
-                   [--allow-events [ALLOW_EVENTS]] [--rules <rules-file>]
-                   [--strictcsums] [--min-version <min-version>]
-                   [--midstream]
+                   [--allow-events [ALLOW_EVENTS]]
```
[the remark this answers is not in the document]
TAnhMinh: Yes, thank you, that was not intentional - new attempt was made in V2

PR #1139, opened 2023-03-08
smtp-eve: fix filesize and version check

inashivb:
Earlier, the CRLFs that were a part of the file were also stripped off as a part of finding and stripping the delimiters in the MIME handler. This was fixed as a part of
https://redmine.openinfosecfoundation.org/issues/5725. This patch fixes the test too to reflect the fix.

Ticket: [5821](https://redmine.openinfosecfoundation.org/issues/5821)

### DIFF
--- a/tests/smtp-eve/test.yaml
+++ b/tests/smtp-eve/test.yaml
@@ -64,7 +64,7 @@ checks:
       src_port: 1470
       tx_id: 0
 - filter:
-    version: 6.0.0 # FIXME ref: https://redmine.openinfosecfoundation.org/issues/5821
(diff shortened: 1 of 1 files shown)


tests/smtp-eve/test.yaml
```diff
@@ -64,7 +64,7 @@ checks:
       src_port: 1470
       tx_id: 0
 - filter:
-    version: 6.0.0 # FIXME ref: https://redmine.openinfosecfoundation.org/issues/5821
+    min-version: 6.0.0
```
inashivb, line 67: Remove

PR #1147, opened 2023-03-17
tests: add test for bug 5867 false positive drop log events - v1

jufajardini:
Test for [bug 5867](https://redmine.openinfosecfoundation.org/issues/5867)

These tests fail, as of now.


jufajardini:
Blip fixed on https://github.com/OISF/suricata-verify/pull/1151

PR #1149, opened 2023-03-20
datasets: test to load invalid encoded string

inashivb:


tests/datasets-invalid-encoding/test.yaml
```diff
@@ -0,0 +1,10 @@
+requires:
+  files:
+    - src/datasets.c
+
+pcap: ../datasets-02-load/input.pcap
```
inashivb, line 5: no need for this

victorjulien: can you fix it up?

PR #1153, opened 2023-03-23
Http2 header 5780 v1

catenacyber:
Complex to merge for 6 and 7.
Should we just get rid of these rules ? Or use full `http.header`


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1232

PR #1155, opened 2023-03-23
tests: test defrag exception policy with drop-flow - v1

jufajardini:
Defrag memcap and flow memcap do not support flow action for the exception policies, as there is no flow when the exception condition is hit. In such cases, the exception policy must be considered as `drop-packet`. This commit changes the defrag exception policy test to check if this behavior is working.

### DIFF
--- a/tests/exception-policy-defrag-01/test.yaml
+++ b/tests/exception-policy-defrag-01/test.yaml
@@ -8,7 +8,7 @@ args:
 - -k none
 # pretend pretend error in the first fragment
 - --simulate-packet-defrag-memcap=1
-- --set defrag.memcap-policy=drop-packet
+- --set defrag.memcap-policy=drop-flow
 checks:
   - filter:
       count: 0


tests/exception-policy-defrag-01/test.yaml
```diff
@@ -8,7 +8,7 @@ args:
 - -k none
 # pretend pretend error in the first fragment
 - --simulate-packet-defrag-memcap=1
-- --set defrag.memcap-policy=drop-packet
+- --set defrag.memcap-policy=drop-flow
```
jufajardini, line 11: Not sure if we'd rather just error out if someone uses an exception policy value that isn't allowed for the given policy.

But we'd still need to check this, for the case of `auto`in IPS mode, for the master switch.

jufajardini:
Suricata PR was merged: https://github.com/OISF/suricata/pull/8631

victorjulien:
Merged in #1174, thanks!

PR #1159, opened 2023-03-30
detect/content: Add negated endswith tests

jlucovsky:
Issue: 5541

This commit adds test cases for the issue discovered in 5541 so that negated endswith are handled properly.

3 tests
- Negated endswith content that *should* match
- Negated endswith content that *shouldn't* match
- Negated endswith content that *should* match with a content match following it


jlucovsky:
Continued in #1185

PR #1160, opened 2023-03-31
tests: add exception/policy stats counters - v1

jufajardini:
- Add exception policy counters to the existing exception-policy tests
- Add tests for flow-memcap command-line option and flow-memcap exception policy

`exception-policy-flow-memcap` doesn't work for now, as the values change for every run of the test - not sure how to fix that....


jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1178

PR #1161, opened 2023-03-31
Adds test about ftp port when memcap is reached

catenacyber:
Ticket: #5701


victorjulien:
Merged in #1174, thanks!

PR #1166, opened 2023-04-05
libsuricata: Integrate suricata-verify

amirabell:
Integrate suricata-verify with libsuricata by
adding the `run_library` script that allows to
run the suricata-verify suite with `suricata_client`. Modify run.py to support `stream` files.
Add a list of excluded tests `library-excluded-tests` when running in library mode because some functionalities are still missing.
Adapt some tests to run in both executable and library mode. More in details:
 * bug-130, bug-3286*: remove `pcap_cnt` field not supported in library mode.
 * stream-http-extraction-multiple, stream-http-extraction-pipelined, stream-http-signature-matching: add tests for the library stream interface.
 * All the other tests: add callback section to suricata.yaml.


amirabell:
Corresponding suricata PR: https://github.com/OISF/suricata/pull/8683

jufajardini (changes requested):
Hi, thanks for this contribution :)

Is it possible to separate the commit that adds changes to the framework from the one that updates the tests?

PR #1167, opened 2023-04-05
Suricata library v2

amirabell:
Integrate suricata-verify with libsuricata.

Previous PR: https://github.com/OISF/suricata-verify/pull/1166


tests/bug-3286-02-linux-evasion/test.yaml
```diff
@@ -17,17 +17,10 @@ checks:
         event_type: alert
         alert.signature_id: 1
   - filter:
-      count: 1
```
catenacyber, line 20: Why are these checks removed ?

amirabell: Libsuricata has no concept of pcap_cnt because it is not aware of the packet acquisition method. Therefore pcap_cnt is always 0 and not logged. In the above test the alerts are triggered for the same signature and the filter are differentiated by the pcap_cnt. Without the pcap_cnt available I merged the two filters in one and increased the counter to achieve the same result. It is true that the test is slightly less precise as we don't check that we generate an alert at pcap_count 2 and an alert at pcap_count 3 but we only check we have 2 alerts for that specific signature.

catenacyber: > It is true that the test is slightly less precise as we don't check that we generate an alert at pcap_count 2 and an alert at pcap_count 3 but we only check we have 2 alerts for that specific signature.

Looks problematic...

> Libsuricata has no concept of pcap_cnt

Could it be added ?

amirabell: Yes it could if we want.
I did not added it because it is a counter specific to pcap replaying. It is not present in other runmodes.
In libsuricata it should be a global packet counter.

catenacyber: I think we want to keep the ability to test at which packet in a pcap the alert is triggered

catenacyber:
@jasonish are you working on this SV PR about library ?

jasonish:
> @jasonish are you working on this SV PR about library ?

I think we can close it now.

It will be tricky, S-V depends on the Suricata application.  Adapting it for contrived library examples doesn't make that much sense I don't think.

PR #1169, opened 2023-04-05
tests: test flow droped but applayer event logged - v2

jufajardini:
It seems that Suricata will log an applayer event for a dropped flow, for the second packet of the flow. This test demonstrates such behavior, so we can investigate it.

Related to
Task #5510
------------
Previous PR (#915) had tests that passed, because the checks accounted for the events that are possibly logged wrongly. This new PR expects that we don't see such events. Therefore:
New PR with tests failing, to illustrate the situation/possible bug.


jufajardini:
Merged with tests for 5802 in https://github.com/OISF/suricata-verify/pull/1221

catenacyber:
Closing if this was merged, right ?

PR #1123, opened 2023-02-13
Mqtt frames v4

hsadia538:
Ticket: https://redmine.openinfosecfoundation.org/issues/5731

Previous PR: #1112 

Describe changes:
- Add tests for mqtt `trunc.data` frames for truncated messages

Requires: https://github.com/OISF/suricata/pull/8540

PR #1172, opened 2023-04-15
Mqtt frames v5

hsadia538:
Ticket: https://redmine.openinfosecfoundation.org/issues/5731

Previous PR: #1123 

Describe changes:
- Remove tests for mqtt `trunc.data` frames for truncated messages
- Add boundary tests for truncated data

Requires: https://github.com/OISF/suricata/pull/8540


tests/mqtt-frames-truncated/test.rules
```diff
@@ -0,0 +1,17 @@
...
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 4"; frame:header; content:"|10 1c|"; sid:4;)
+
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 5"; frame:pdu; content:"|17 0C E2|"; sid:5;)
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 6"; frame:pdu; content:"|00 00 54 46|"; sid:6;)
+
+# pre-boundary test for truncated data
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 7"; frame:data; content:"|0a|"; sid:7;)
+
+# At boundary test for truncated data
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 8"; frame:data; content:"|51|"; sid:8;)
```
inashivb, line 14: Could we add a few more bytes to ensure it is the truncated data that we want to be looking at. Unless `51` is a very unique character that only happens once at the boundary for MQTT?

tests/mqtt-frames-truncated/test.rules
```diff
@@ -0,0 +1,17 @@
...
+
+# post-boundary test for truncated data
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 9"; frame:data; content:"|5b|"; sid:9;)
```
inashivb, line 17: Same comments as above.

tests/mqtt-frames-truncated/test.rules
```diff
@@ -0,0 +1,17 @@
+alert mqtt any any -> any any (msg:"mqtt truncated Frame 1"; frame:pdu; content:"|10 1c|"; startswith; sid:1;)
```
inashivb, line 1: nit: msg indicates all these are truncated frames which from the matches does not seem to be true?

tests/mqtt-frames/test.yaml
```diff
@@ -0,0 +1,52 @@
+requires:
+  min-version: 7
+
+args:
+ - -k none
```
inashivb, line 5: nit: README mentions the pcap already exists. Can we try to deduplicate it w the `pcap` key here?

inashivb (changes requested):
Really nice work, @hsadia538 ! :)
Some questions and nits inline. Thank you for your work!

tests/mqtt-frames-truncated/test.rules
hsadia538: I am not sure what do you  mean by adding more bytes. For this purpose of this rule, I counted the bytes up till a point where there is a boundary. Since after last update to frames, we are no longer making frames for truncated data so after this specific byte we don't get a match

tests/mqtt-frames-truncated/test.rules
hsadia538: I can add more bytes to the post boundary test

hsadia538:
followed by #1177

tests/mqtt-frames-truncated/test.rules
catenacyber: It looks like we get a match even after this specific byte cf https://github.com/OISF/suricata-verify/pull/1365

inashivb: yep. did think that might happen. Thanks for picking this up! :)

PR #1117, opened 2023-02-08
rfb: test rfb frames v5

hsadia538:
Ticket: https://redmine.openinfosecfoundation.org/issues/5717

Previous PR: #10509

Describe new changes:
- Rebase branch

Requires: https://github.com/OISF/suricata/pull/8526

PR #1176, opened 2023-04-17
rfb: test rfb frames

hsadia538:

Previous PR: #1117

Describe new changes:
- Rebase branch


tests/rfb-frames/test.yaml
```diff
@@ -0,0 +1,30 @@
+requires:
+  min-version: 7
```
inashivb, line 2: nit: Deduplicate the pcap

inashivb (changes requested):
LGTM. Nice work! :)
Just one minor change requested inline.
Hopefully, I'll add a CI job that'll start to fail if one duplicates a pcap so CI would be able to point it out to you..

PR #1179, opened 2023-04-18
tests: add test for bug 2917

inashivb:
Ticket: https://redmine.openinfosecfoundation.org/issues/2917

Previous PR: https://github.com/OISF/suricata-verify/pull/1120

Changes since v1:
- Entire test has been converted from a shell check looking for error in logs to filters on engine analysis


inashivb:
Seems like the logging format has changed quite a bit since 6.. Changing the min version

PR #1180, opened 2023-04-18
tests: add test for bug 2917

inashivb:
Ticket: https://redmine.openinfosecfoundation.org/issues/2917

Previous PR: https://github.com/OISF/suricata-verify/pull/1179

Changes since v2:
- min-version locked to 7 as engine output differs a lot b/w 6 and 7


tests/bug-2917/test.yaml
```diff
@@ -0,0 +1,215 @@
+requires:
+  min-version: 7
+
+pcap: false
+
+args:
+  - --engine-analysis
```
inashivb, line 7: @victorjulien does anything seem out of the ordinary in the following checks per rule?

tests/bug-2917/test.rules
```diff
@@ -0,0 +1,15 @@
+alert http $HOME_NET any -> $EXTERNAL_NET $HTTP_PORTS (msg:"TEST 1"; flow:to_server,established; content:"GET"; http_method; flowbits:set,test; sid:1;)
+
+# does not work - based on the original rules
```
victorjulien, line 3: what does not work?

tests/bug-2917/test.yaml
```diff
@@ -0,0 +1,215 @@
...
+
+checks:
+  - filter:
+      filename: rules.json
+      count: 1
+      match:
+        app_proto: http
```
victorjulien, line 14: I find this sorting very hard to read. The default output order should ideally be preserved here

inashivb: Sorry. Didn't realize I did that.

tests/bug-2917/test.rules
inashivb: This is where, before `227d1d967a5902` (that suppressed the error), one would get the error `Unable to find the sm in any of the sm lists`.
The error disappeared if things were done differently (in 3 different ways) in the following 3 rules.

victorjulien: can you update the message to reflect that in the past this generated an error/warning?

PR #1185, opened 2023-04-19
detect/content: Add negated endswith tests

jlucovsky:
Continuation of 1159

This commit adds test cases for the issue discovered in 5541 so that negated endswith are handled properly.

3 tests

Negated endswith content that should match
Negated endswith content that shouldn't match
Negated endswith content that should match with a content match following it

Updates:
- Added `gt-version` verb in run.py
- Modified ends-with test cases to to support 6.0.x


jasonish:
Merged via https://github.com/OISF/suricata-verify/pull/1199. Thanks!

PR #1201, opened 2023-05-12
Stage pull requests

jasonish:
Staging:
- https://github.com/oisf/suricata-verify/pull/223


tests/tls-directionless/test.yaml
```diff
@@ -0,0 +1,15 @@
+# currently failing
+skip: true
```
victorjulien, line 2: only passes due to this

jasonish: Ah, well that prevents any form of automated rechecking...

victorjulien: agree, we shouldn't use this as a pattern

PR #1202, opened 2023-05-15
smtp: Add test to match on attachment with md5

thomasjwinter:
Based on the filemd5 test but using smtp attachment instead. The SMTP transaction contains the EICAR file as an attachment and the expected md5 to match used is the standard md5 for the EICAR.

This test currently fails because the smtp/mime-decoder is including the \r\n delimiter in the attachment buffer which causes a different md5 to be calculated compared to the standard EICAR md5.


catenacyber:
Thanks for this test.

> This test currently fails because the smtp/mime-decoder is including the \r\n delimiter in the attachment buffer which causes a different md5 to be calculated compared to the standard EICAR md5.

catenacyber:
And does https://github.com/OISF/suricata/pull/8886 solve the issue ?

thomasjwinter:
@catenacyber redmine ticket is here https://redmine.openinfosecfoundation.org/issues/6023

No, your branch does not fix the issue. EICAR md5 is still calculated to e7e5fa40569514ec442bbdf755d89c2f.

catenacyber:
Thank you Thomas

Are you planning to do a Suricata fix as well ?
It looks like the issue is with SMTP supplying data to MIME line after line, so MIME decoder cannot know that it should or should not take the end of line, because it cannot know wether the next line is a boundary...

thomasjwinter:
@catenacyber sorry, I am not planning to do a Suricata fix as well.

I attempted a fix here https://redmine.openinfosecfoundation.org/attachments/2786 but this in resulted in some extra suricata-verify failures https://redmine.openinfosecfoundation.org/attachments/2804
I'm not familiar enough with smtp/mime encoding to understand what to do here. It may be the case sometimes you need to include the delimiter and sometimes you shouldn't.

catenacyber:
No worries, I will take it, just wanted to avoid duplicating the work if you were already on it.

Thanks for the test

catenacyber:
Replaced by #1265

PR #1214, opened 2023-05-24
datasets: test to load invalid encoded string

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/1149

Changes since v1:
- Remove pcap reference as it is not needed

### DIFF
--- a/tests/datasets-invalid-encoding/README.md
+++ b/tests/datasets-invalid-encoding/README.md
@@ -1,0 +1,9 @@
+# Test Description
+
+This test demonstrates that datasets would error out in case they're
+given a bad base64 encoded string to load.
+
+
+## Related issues
+
+https://redmine.openinfosecfoundation.org/issues/5885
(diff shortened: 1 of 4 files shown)


victorjulien:
Does cause suri to generate some error message that we can match on in test.yaml?

inashivb:
> 

Yes. thank you for pointing that out. This test is absolutely incorrect. The exit code 1 is not because of the error but because the engine refused to start w/o a pcap file. So, I think I'll have to rollback to v1 + add a grep for the error message.

victorjulien:
> > 
> 
> Yes. thank you for pointing that out. This test is absolutely incorrect. The exit code 1 is not because of the error but because the engine refused to start w/o a pcap file (both have exit code 1 btw). So, I think I'll have to rollback to v1 + add a grep for the error message.

instead of grep please do a engine json output plus normal test.yaml matching

PR #1216, opened 2023-05-24
Add tests for exception policy midstream bug 5825 - v1

jufajardini:
Related to #5825

TODOs:
- add tests for IPS mode
- (probably) remove eve.json config from `suricata.yaml` for the tests that error out and don't use it
- (possibly) update IDS `drop` tests with proper expected results if we keep the current behavior of suri setting these to `ignore`
- (possibly) adding the missing tests for IDS (thought they might be repetitive, dunno)

### DIFF
--- a/tests/exception-policy-ids-midstream-disabled/exception-policy-midstream-disabled-drop-packet/README.md
+++ b/tests/exception-policy-ids-midstream-disabled/exception-policy-midstream-disabled-drop-packet/README.md
@@ -1,0 +1,15 @@
+# Test
+
+Check that the midstream exception policy is properly applied in case Suricata
+has stream midstream pick-up sessions disabled. In this test the exception policy
+for midstream sessions is set to ``drop-packet``. This test is for IDS mode.
+
+# Behavior
+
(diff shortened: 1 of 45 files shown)


tests/exception-policy-ids-midstream-disabled/exception-policy-midstream-disabled-drop-packet/README.md
```diff
@@ -0,0 +1,15 @@
...
+
+Check that the midstream exception policy is properly applied in case Suricata
+has stream midstream pick-up sessions disabled. In this test the exception policy
+for midstream sessions is set to ``drop-packet``. This test is for IDS mode.
+
+# Behavior
+
+We expect Suri to error out without starting as ``drop-packet`` isn't a valid
+exception policy value for the midstream exception policy.
+
```
jufajardini, line 11: We must decide the behavior for ``drop-packet`` and ``drop-flow``, as the current one is different from this approach.

jufajardini:
Closing, next week should see a reviewable PR.

PR #1218, opened 2023-05-25
decode-event: Add test for too small UDP and TCP packets

coledishington:
Add test for IPv4 and IPv6 packets that set proto/next-header to UDP and TCP but have no payload.

Bug: #6086.


catenacyber:
Thanks for this @coledishington 

What happens if we have 4 bytes of UDP header (instead of 0) ?

coledishington:
@catenacyber Removed the makefile and added a packets with partial UDP headers. The partial UDP headers trigger the same counters/rules. Changes are in https://github.com/OISF/suricata-verify/pull/1231/files

PR #1220, opened 2023-05-26
Datasets test/v4

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/1214

Changes since v2:
- allow matching substring in yaml
- add match with suricata.json

Note: substring matching in test.yaml allows to match on certain error strings that may contain variables like directory paths which may change from one environment to another.

### DIFF
--- a/tests/datasets-invalid-encoding/test.yaml
+++ b/tests/datasets-invalid-encoding/test.yaml
@@ -1,0 +1,20 @@
+requires:
+  files:
+    - src/datasets.c
+
+pcap: ../datasets-02-load/input.pcap
+
+args:
+ - -k none
+
+exit-code: 1
+
+checks:
+  - filter:
+      filename: suricata.json
+      count: 1
+      match:
+        log_level: "Error"
+        event_type: "engine"
+        engine.message.__slice[27]: "bad base64 encoding ua-seen"
+        engine.module: "datasets"
--- a/tests/datasets-invalid-encoding/README.md
+++ b/tests/datasets-invalid-encoding/README.md
@@ -1,0 +1,9 @@
+# Test Description
+
+This test demonstrates that datasets would error out in case they're
+given a bad base64 encoded string to load.
+
+
+## Related issues
+
+https://redmine.openinfosecfoundation.org/issues/5885
--- a/tests/datasets-invalid-encoding/datasets.csv
+++ b/tests/datasets-invalid-encoding/datasets.csv
@@ -1,0 +1,1 @@
+Y3VybC83Lj!QzLjA=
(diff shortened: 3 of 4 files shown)


tests/datasets-invalid-encoding/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+exit-code: 1
+
+checks:
+  - filter:
+      filename: suricata.json
+      count: 1
+      match:
+        log_level: "Error"
+        event_type: "engine"
+        engine.message.__slice[27]: "bad base64 encoding ua-seen"
```
victorjulien, line 19: this feels over-engineered, can we just do an exact match?

inashivb: Full `engine.message` for me locally looks like: 
```shell
bad base64 encoding ua-seen//home/ard/Orgs/oisf/public/suricata-verify/tests/datasets-invalid-encoding/datasets.csv
```
I added this because,
1. message will change as per the env this is run in (dir)
2. yaml cannot expand on env vars so can't use $PWD here
3. you mentioned we don't want to grep so no shell checks with grep

How do you propose we deal with this?

inashivb (commented):
Forgot committing the config file 🤦🏽‍♀️

tests/datasets-invalid-encoding/test.yaml
jasonish: > @jasonish do you have a Python trick up your sleeve to somehow match only on the substring or substitute the $PWD without much of a change in the framework? We're considering going back to shell checks with grep now..

No, I don't.  I think this slice syntax is a little hard to reason about.  I wonder if finding a substring would be a good enough test?

Introduce `find.message.__find: "bad base 64 encoding ua-seen"`

Expanding environment variable on the value side of the YAML would require some expression syntax for expansion, simply swapping `$SOMETHING` for an environment variable of `$SOMETHING` could break other things.

I'm sure we'd have a use for finding substrings later on.

jasonish: If we wanted to do environment variable substitution on the right hand side, I'd use GitHub Actions as an example...

```
engine.message: "${{env.CURDIR}} blah blah"
```
where `${{env.CURDIR}}` is used for environment substation.  Might seem a little heavy on syntax for now, but keeps us open to other expansion syntax..

For example, I also see myself doing something like GitHub Actions matrix notation (example: https://github.com/jasonish/docker-suricata/blob/main/.github/workflows/build.yml#L15), and variable expansion is key to this working, `${{ matrix.NAME }}`...

inashivb: Thank you! This env way looks fun. I'm not well informed of all the params that might show up as per the env in engine log. So, for now, I'm gonna implement the `__find` which indeed sounds much better and more versatile. Not sure why didn't I do that in the first place! :D

PR #1223, opened 2023-06-01
Add and update tests for midstream exception policy bug 5825 - v2

jufajardini:
Previous (draft) PR: https://github.com/OISF/suricata-verify/pull/1216

Changes:
- add tests to check on midstream behaviors in IDS and IPS mode, with the different variations of midstream policy/midstream enabled/disabled IPS (IDS mode is missing cases for 'drop-flow' in ids mode. Felt redundant, but now I feel I must add those. 
- tests for bug-5825 (midstream exception policy config conflict) added into dedicated directory
- test 'exception-policy-ids-midstream-disabled-reject' fails possibly due to bug where we don't differentiate the reject if we're in IDS or IPS mode?
- test 'bug-2491-02' set midstream policy to 'ignore' as 'drop-flow' would lead to Suri refusing to init for invalid config
- Better explain 'exception-policy-midstream-03' behavior and scenario in the README

TODO:
- add test cases for 'drop-flow' in ids mode
- double-check if tests do what we expect them to

Suricata PR: https://github.com/OISF/suricata/pull/8948

### DIFF
--- a/tests/bug-2491-02/test.yaml
+++ b/tests/bug-2491-02/test.yaml
@@ -4,7 +4,7 @@ requires:
 args:
 - --set stream.async-oneside=true
 - --set stream.midstream=true
-- --set stream.midstream-policy=drop-flow
+- --set stream.midstream-policy=ignore
 
 checks:
   - filter:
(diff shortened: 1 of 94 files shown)


jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1226

PR #1227, opened 2023-06-02
tests: update ips-state-1 test - v1

jufajardini:
This test indicated that there were FP drops for http and that another check was failing, but currently the test passes and we don't observe such things. 

The flow event for 'tls' still doesn't show the 'action' field, though, as the flag for ACTION_DROP isn't marked as true (as the README says). Should we add a check for that and make it so that the test fails again?

### DIFF
--- a/tests/ips-state-1/README.md
+++ b/tests/ips-state-1/README.md
@@ -5,9 +5,5 @@ be full passed with no alerts, while the TLS flow should be dropped.
 
 ## Current Observations
 
-- HTTP response packets are being logged as dropped, however the transaction is
-  logged suggesting the drop is only in logging only, but not actually
-  occurring.
-
 - All the TLS packets apear to be getting dropped, but `flow.action` is never
   set to true.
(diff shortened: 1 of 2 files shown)


catenacyber:
Is there a redmine ticket associated ?

jufajardini:
> Is there a redmine ticket associated ?

Sorry, I think I missed the notification for this one. I haven't created a ticket because I don't know if there is something to be added/ changed, my intention was more to update test details, considering the observed behavior and what the test comments said...

catenacyber:
So, should this PR be a draft ?
Can I help somehow ?

jufajardini:
> So, should this PR be a draft ? Can I help somehow ?

Thanks, not sure if it needs to be a draft, it does need review/ comparing what we had with what is there now, to see if the changes make sense, and if I should add that check I mentioned, to make the test fail again...

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1496

PR #1229, opened 2023-06-05
test: add test for bug 6109 - v1

jufajardini:
In IDS mode, with midstream disabled, it seems that the exception policy 'reject' is applying actions that should only be taken in IPS mode. This leads to the flow being dropped (or logged as such?) in IDS mode. This test showcases this.

Bug #6109 (https://redmine.openinfosecfoundation.org/issues/6109)

This test fails, currently.

### DIFF
--- a/tests/bug-6109-reject-policy-ids/README.md
+++ b/tests/bug-6109-reject-policy-ids/README.md
@@ -1,0 +1,20 @@
+# Test
+
+Check that the midstream exception policy is properly applied in case Suricata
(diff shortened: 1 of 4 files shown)


catenacyber:
Suricata PR was merged, and now these tests pass locally for me.
@jufajardini would you rebase and force-push to get CI running and checking all is green ?

jasonish:
Merged via https://github.com/OISF/suricata-verify/pull/1263. Thanks!

PR #1231, opened 2023-06-06
decode-event: Add test for too small UDP and TCP packets

coledishington:
Add test for IPv4 and IPv6 packets that set proto/next-header to UDP and TCP but have no payload.

Bug: #6086.


catenacyber:
Suricata PR was merged, but not back ported... (do not know if it should be back ported)

catenacyber:
Rebased in https://github.com/OISF/suricata-verify/pull/1259

PR #1232, opened 2023-06-07
http2: move http.request_header keyword to new test

catenacyber:
Replaces #1153 with creating a new S-V test instead of reusing other ones to test for the keyword

https://redmine.openinfosecfoundation.org/issues/5780


victorjulien:
Merged in #1238, thanks!

PR #1028, opened 2022-12-02
tests: Update versions to accommodate dsize/bsize validation backport

jlucovsky:
This PR updates the versions associated with tests that depend on the new/old bsize/dsize validation. The backport (issue 4317) of the dsize/bsize validation logic to 6.0.x requires tests with different behavior to be version-filtered.

PR #1237, opened 2023-06-09
Jlucovsky 4317/2

catenacyber:
Rebase #1028 cc @jlucovsky


catenacyber:
Closing as Jeff told me that 

> I think the decision was to not backport the dsize/bsize changes because that could cause some rules currently loaded to not be loaded for 6.0.x deployments

PR #1165, opened 2023-04-03
Detect lua 4911 v3

TheKharleeci:
Includes test for LuaMatchTest01-LuaMatchTest06

Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/4911

Previous PR: https://github.com/OISF/suricata-verify/pull/1086

PR #1242, opened 2023-06-11
detect-lua: add tests

TheKharleeci:
Includes test for LuaMatchTest01-LuaMatchTest06

Previous PR: https://github.com/OISF/suricata-verify/pull/1165


tests/detect-lua/detect-lua-04/test.yaml
```diff
@@ -0,0 +1,17 @@
+args:
```
jufajardini, line 1: It seems to me that the pcaps for this test and for `detect-lua-01` are the same. In this case, here, you could add:
```yaml
pcap: ../detect-lua-01/lua-http.pcap
```

Then you can delete the pcap for this test. That would help us reuse the existing pcap and remove duplicated files ;)

jufajardini (changes requested):
Thanks for adding the README files :)

I think we are almost there. A few comments, but these are more for polishing, now that the tests are passing and we have the kickstart of the README files.

- I've noticed that some of those have basically the same description. I've added some ideas on what could be present here, as well: https://github.com/OISF/suricata-verify/pull/1086/files#r1099095555. Could you please add more info so one can understand what's the difference between the tests?
Info on where did the pcap come from is also good to be present there.
- This PR doesn't have any scapy scripts, have you not used any, or it slipped off the radar between the PRs?

catenacyber:
Closing as stale, feel free to reopen

PR #1249, opened 2023-06-14
Datasets test/v6

inashivb:
Previous PR: #1220 

Changes since v4:
- add a `find` method instead of `slice` method

### DIFF
--- a/run.py
+++ b/run.py
@@ -382,7 +382,10 @@ def find_value(name, obj):
                 return len(obj)
             except:
                 return -1
-
+        if part == "__find":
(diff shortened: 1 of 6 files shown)


catenacyber:

inashivb:
> Is there a redmine ticket for this ?

No. When I was working on base64 issues, Victor had asked for this test to see if datasets behaved as expected and they did. So, we have this test just to increase the coverage, I guess

jasonish:
Merged via https://github.com/OISF/suricata-verify/pull/1263. Thanks!

PR #1255, opened 2023-06-20
tls: adds a test for certificate without issuer

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5439

Looks like the bug was already solved ?


catenacyber:
Test passes on 7, but not on 6

victorjulien:
Merged in #1304, thanks!

PR #1264, opened 2023-06-22
tests: add tests for exception policy master switch - v6

jufajardini:
Add tests for the new master switch config for the Exception policies. Added also a check in the exception-policy test that checks the default behavior, to ensure that things work as expected with the master switch disabled and the exception policies not set.

Task #5219

Previous PR: 
https://github.com/OISF/suricata-verify/pull/1044

Changes from previous PR: 
- rebased

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5219


jufajardini:
I noticed that we hadn't merged the SV tests for the master switch, probably because the previous PR started to fail when we disabled the midstream policies and I didn't manage to go back to it...

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1280

PR #1272, opened 2023-06-28
runner: allow a test to be retried

jasonish:
Add a new parameter, retry that takes count. If the checks fail, the
test will be re-run. This could help us deal with failures in tests
that are sensitive to timing.

Example test that often fails in CI:
https://github.com/OISF/suricata/actions/runs/5404663247/jobs/9819234965?pr=9093#step:16:140

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ checks:
 
       # The number of records this filter should match.
       count: 1
-	  
+
       # The fields to match on.
       match:
         # Example match on event_type:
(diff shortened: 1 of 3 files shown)


catenacyber:
Have we ever managed to reproduce these failures locally ?

jasonish:
> Have we ever managed to reproduce these failures locally ?

Only as far as running S-V with a concurrency level of 80 on a 20 core system in an attempt to oversubscribe it, but never fails locally.  We once had a similar job sometime ago I could trigger by do something similar locally, but not this one.

victorjulien (approved):

catenacyber:
What is preventing a merge if this got approved ?

victorjulien:
Merged in #1272, thanks!

PR #1274, opened 2023-06-30
rfb: add test case for logging of partial txs

satta:
## Ticket

This PR introduces a test case referencing the following change: https://github.com/OISF/suricata/pull/9106, in particular addressing the request in comment https://github.com/OISF/suricata/pull/9106#issuecomment-1613893802.

The tx involves an Apple macOS desktop sharing server, which uses a non-RFC authentication type we do not implement. This raises an "unimplemented_security_type" event and -- in the Suricata master before https://github.com/OISF/suricata/pull/9106, would cause the transaction not to be logged at all, not even the version strings, which have been parsed successfully.
After the Suricata change, a partial event with the data available so far would be logged (i.e. version strings for server and client and the unsupported security type number).


victorjulien:
@satta this fails on both master and master-6.0x. Does this mean we need a backport for master-60x as well?

satta:
It needs https://github.com/OISF/suricata/pull/9106 to be merged into master to not fail.

I think I could also just skip this test on 6.x since all the RFB changes would probably be a bit too big for a 6.0.x backport I guess. What do you think? I can add a version check quickly.

victorjulien:
> It needs [OISF/suricata#9106](https://github.com/OISF/suricata/pull/9106) to be merged into master to not fail.
> 
> I think I could also just skip this test on 6.x since all the RFB changes would probably be a bit too big for a 6.0.x backport I guess. What do you think? I can add a version check quickly.

Sounds good, lets limit it to 7 for now.

satta:
Next PR: #1279

PR #1279, opened 2023-07-02
rfb: add test case for logging of partial txs - v2

satta:
## Ticket

Changes w.r.t. previous PR:

* Add version constraints allowing to skip this test.

The tx involves an Apple macOS desktop sharing server, which uses a non-RFC authentication type we do not implement. This raises an "unimplemented_security_type" event and -- in the Suricata master before https://github.com/OISF/suricata/pull/9106, would cause the transaction not to be logged at all, not even the version strings, which have been parsed successfully.
After the Suricata change, a partial event with the data available so far would be logged (i.e. version strings for server and client and the unsupported security type number).

Previous PR: #1274 

changes to the ticket status: 
https://github.com/OISF/suricata/pull/9114


victorjulien:
Merged in #1286, thanks!

PR #1280, opened 2023-07-03
tests: add tests for exception policy master switch - v7

jufajardini:
Add tests for the new master switch config for the Exception policies. Added also a check in the exception-policy test that checks the default behavior, to ensure that things work as expected with the master switch disabled and the exception policies not set.

Task #5219

Previous PR:
https://github.com/OISF/suricata-verify/pull/1264

Changes from previous PR:
- add tests that checks for midstream exception policy with default config and checks that Suri doesn't warn on setting packet actions for policies that don't support flow actions (cf https://github.com/OISF/suricata/pull/9059#pullrequestreview-1504741786)
- rebased
- updated test for bug 6149 to check for log message indicating that master switch is set to 'auto', as per proposed changes (see Suri PR https://github.com/OISF/suricata/pull/9130)

Ticket
Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5219


victorjulien:
Merged in #1286, thanks!

PR #1287, opened 2023-07-05
tests: test for address lists

jasonish:
Add tests for address lists including negation. The idea is to better understand how address lists are constructed when negation is included.

So far nothing stands out as odd, all of the combinations are processed as expected. However, deeper nesting may show issues.


catenacyber:
Is something expected here Jason ?
Or could this draft be closed ?

jasonish:
Closing for now, part of some new rule parsing work.

PR #1306, opened 2023-07-11
mime: add test for bug 6207

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6207


inashivb:
It does not trigger in 6 like https://github.com/OISF/suricata/actions/runs/5511424793/jobs/10046956946?pr=9204
Only has some missing events and extra anomaly events that we log for invalid base64 in 6

PR #1317, opened 2023-07-19
workflows: disable docs during make

inashivb:
Requires https://github.com/OISF/suricata/pull/9258


catenacyber:
Replaced by #1319 ?

PR #1260, opened 2023-06-20
B1tg issue/6008 smb v4

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6008

Rebase of https://github.com/OISF/suricata-verify/pull/1211

PR #1318, opened 2023-07-19
tests: add test for bug 6008 SMB_COM_WRITE_ANDX data offset issue

catenacyber:
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6008

Passing part of #1260 

But do we want to add a 22Mbytes pcap ?

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -131,7 +131,6 @@ jobs:
                 pkgconfig \
                 python3-devel \
                 python3-jsonschema \
-                python3-sphinx \
                 python3-yaml \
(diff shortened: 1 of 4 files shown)


victorjulien:
> But do we want to add a 22Mbytes pcap ?

It was trivial to trim it. See #1340. Please check this next time.

catenacyber:
> > But do we want to add a 22Mbytes pcap ?
> 
> It was trivial to trim it. See #1340. Please check this next time.

Indeed, but I was not sure we wanted it as we lose some context : Trickbot-goes-from-client-to-domain-controller

PR #1320, opened 2023-07-20
decode-event: Add test case for GRE packet too small

coledishington:
Add test for IPv4 and IPv6 packets that set proto/next-header to GRE but have an invalid payload.

Bug: #6222

## Ticket

changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6222


victorjulien:
Merged in #1321, thanks!

PR #1343, opened 2023-08-03
iprep: test reputation & category file newline handling

coledishington:
The commit 'e7c0f0ad9 src: remove multiple uses of atoi' caused a regression in parsing of ip-rep reputation config files.

Previously, parsing ip-rep values in reputation config files succeeded for CR and CR+LF due to the use of atoi() in parsing ip-rep values, when the line was split by SRepSplitLine the \r at following the reputation score was ignored.

Bug: #6243

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6243


thomasjwinter:
https://github.com/OISF/suricata/pull/9357

victorjulien:
Merged in #1372, thanks!

PR #1350, opened 2023-08-10
decode-event: change to match on protocol gre specifically

thomasjwinter:
Instead of matching on all packets or having to specify the GRE protocol number directly, we can use use the protocol keyword.

Feature: #6261

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6261

### DIFF
--- a/tests/decode-too-small/test.rules
+++ b/tests/decode-too-small/test.rules
@@ -1,3 +1,3 @@
 alert tcp any any -> any any (msg:"TCP packet too small"; decode-event:tcp.pkt_too_small; sid:1;)
 alert udp any any -> any any (msg:"UDP packet too small"; decode-event:udp.hlen_too_small; sid:2;)
-alert ip any any -> any any (msg:"GRE packet too small"; decode-event:gre.pkt_too_small; sid:3;)
+alert gre any any -> any any (msg:"GRE packet too small"; decode-event:gre.pkt_too_small; sid:3;)


jufajardini (changes requested):
nit: commit message has a typo (double `use`).

Can we have more tests showcasing GRE usage? I saw that test `bug-5633` also seems to have GRE traffic, not sure if it's the best, but would be nice to see this PR show a bit more of https://github.com/OISF/suricata/pull/9364 and that tests the expected behavior.

thomasjwinter:
I tried changing those tests to use this new gre protocol but got failures. I then tried using ```ip_proto:47``` as Julian suggested and got the same failures. Is that a reproduction of [that bug](https://redmine.openinfosecfoundation.org/issues/5633) as both of the changed rules were pass rules? I don't understand what should be actually happening in this case or why changing the rule to match on all IP vs just GRE causes the change in behaviour when all the packets in the pcap are gre encapsulated.
```
 ../suricata-verify-tjwgh/run.py  bug-5633

    Finished release [optimized] target(s) in 0.03s
Number of concurrent jobs: 8
===> bug-5633-gre-01: Sub test #1: FAIL : expected 0 matches; got 2 for filter {'count': 0, 'match': {'event_type': 'alert'}}
===> bug-5633-gre-02: Sub test #1: FAIL : expected 0 matches; got 4 for filter {'count': 0, 'match': {'event_type': 'alert'}}

PASSED:  0
FAILED:  2
SKIPPED: 0

```

jufajardini:
> 

Without seeing the changes you've made to the rules/tests, my assumption is that, as the pass rule specifies what's the IP that's covered by it, some packets have a different IP, and therefore are matching on a broader gre/ip rule, as that's what I observed with some local tests.

If I remove the pass rule, and have a gre rule in place, my eve looks like:
```json
{"timestamp":"2022-11-24T15:51:38.513355+0000","pcap_cnt":1,"event_type":"alert","vlan":[6],"src_ip":"1.1.1.2","src_port":0,"dest_ip":"1.1.1.1","dest_port":0,"proto":"GRE","pkt_src":"wire/pcap","alert":{"action":"allowed","gid":1,"signature_id":2,"rev":0,"signature":"","category":"","severity":3}}
{"timestamp":"2022-11-24T15:51:38.513355+0000","flow_id":797468085676186,"event_type":"alert","src_ip":"2.2.2.3","src_port":12345,"dest_ip":"2.2.2.2","dest_port":514,"proto":"UDP","pkt_src":"gre tunnel","alert":{"action":"allowed","gid":1,"signature_id":1,"rev":0,"signature":"","category":"","severity":3},"tunnel":{"src_ip":"1.1.1.2","src_port":0,"dest_ip":"1.1.1.1","dest_port":0,"proto":"GRE","depth":1,"pcap_cnt":1,"pkt_src":"wire/pcap"},"app_proto":"failed","direction":"to_server","flow":{"pkts_toserver":1,"pkts_toclient":0,"bytes_toserver":32,"bytes_toclient":0,"start":"2022-11-24T15:51:38.513355+0000","src_ip":"2.2.2.3","dest_ip":"2.2.2.2","src_port":12345,"dest_port":514}}
```
So my understanding is that the pass rule is blocking traffic like the second event, but not the first. Now, if Suri should interpret all the traffic the same, here, and this is a bug, that's something I would have to spend more time to understand...

catenacyber:
Closing as https://github.com/OISF/suricata/pull/9364 was closed and both PRs are referenced in https://redmine.openinfosecfoundation.org/issues/6261#note-1

PR #1352, opened 2023-08-18
tests: deduplicate pcaps, cleanup extras

inashivb:
Previous PR: #1351 

Changes since v1:
- Fix pcap references for 6


jufajardini:
While reviewing this, I noticed that if a pcap file is removed, we'll have two behaviors for the tests that use it:
- the one which contained the pcap file will be skipped, with a message (in non-quiet mode) stating missing pcap
- the ones with a link to it will fail, with the error message going to `stderr`

Makes me wonder if we could/should make this a bit more user-friendly. Maybe just documenting that we deduplicate pcap files, and if a file is removed or renamed, such and such behaviors would be observed, in the failing tests?

inashivb:
> While reviewing this, I noticed that if a pcap file is removed, we'll have two behaviors for the tests that use it:
> 
>     * the one which contained the pcap file will be skipped, with a message (in non-quiet mode) stating missing pcap
> 

I think if this is happening, we should fail if pcap is not present in a dir unless `pcap:false` option is given in test.yaml.

>     * the ones with a link to it will fail, with the error message going to `stderr`
> 

This sounds about right.

> 
> Makes me wonder if we could/should make this a bit more user-friendly. Maybe just documenting that we deduplicate pcap files, and if a file is removed or renamed, such and such behaviors would be observed, in the failing tests?

Sure. Although, this is first in line for the actual work in which we'll fail the CI in case a newly added test has a pcap that already exists in s-v repo. So, something in lines of that?

jufajardini:
> I think if this is happening, we should fail if pcap is not present in a dir unless `pcap:false` option is given in test.yaml.

> This sounds about right.
> 
> Sure. Although, this is first in line for the actual work in which we'll fail the CI in case a newly added test has a pcap that already exists in s-v repo. So, something in lines of that?

Yeah, I think that the behaviors look correct, I pointed these out more for cases of people making local changes that could lead to such failures and to help them understand what those could be, so a step that would happen ideally before the CI... What do you think is the best approach, here?

inashivb:
> 
> Yeah, I think that the behaviors look correct, I pointed these out more for cases of people making local changes that could lead to such failures and to help them understand what those could be, so a step that would happen ideally before the CI... What do you think is the best approach, here?

Indeed. The plan is to have something in the repo much like the shell script we have for clang formatting. If anyone wants to check for dups then, they could run that script every time or make it a part of their git hooks.

jufajardini:
> > Yeah, I think that the behaviors look correct, I pointed these out more for cases of people making local changes that could lead to such failures and to help them understand what those could be, so a step that would happen ideally before the CI... What do you think is the best approach, here?
> 
> Indeed. The plan is to have something in the repo much like the shell script we have for clang formatting. If anyone wants to check for dups then, they could run that script every time or make it a part of their git hooks.

I see, thanks, sorry for not grasping this part before xD
I'm worried about errors that happen when someone deletes or renames a pcap that has been deduped, as the errors are different for the test with the original test or the others that add a reference to it.
So there's the error for when there is a duplicated pcap that has to be deduped, and there are the errors when/if a deduped pcap is changed or deleted. 

Just wanted to make sure this latter scenario is properly documented or explained, as with this the tests get more "entangled", and one change may impact something unsuspected :P

jasonish:
> Makes me wonder if we could/should make this a bit more user-friendly. Maybe just documenting that we deduplicate pcap files, and if a file is removed or renamed, such and such behaviors would be observed, in the failing tests?

But such failures should never make it into the master branch right?

jasonish:
I had this staged, but am now unstaging.  There appears to be intermittent failures in tests where the path points to another test that itself is also path.

In particular, the geoip test.

jufajardini:
> > Makes me wonder if we could/should make this a bit more user-friendly. Maybe just documenting that we deduplicate pcap files, and if a file is removed or renamed, such and such behaviors would be observed, in the failing tests?
> 
> But such failures should never make it into the master branch right?

Right, I'm thinking more of when someone is creating new tests or editing old ones and ends up stumbling on stuff like this.

jufajardini:
> I had this staged, but am now unstaging. There appears to be intermittent failures in tests where the path points to another test that itself is also path.
> 
> In particular, the geoip test.

I'll submit a fix. For this one, at least...

jufajardini:
Trying to fix the geoip fail-case: https://github.com/OISF/suricata-verify/pull/1381

catenacyber:
Should this PR be closed in favor of #1381 ?

inashivb:
> Should this PR be closed in favor of #1381 ?

Indeed

PR #1354, opened 2023-08-19
multi-tenant: Basic multi-tenant selector tests

jlucovsky:
This commit adds basic tests for vlan-selector configurations (including vlan-pair, added in 6237)

Issue: 6237

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6237


jasonish:
What happens when you have 2 inner vlan's with the same ID, but different outer IDs?  Or is that not a normal deployment scenario?

jlucovsky:
Good point.  It’s a customer issue  and they must be aware of that.  Should the doc bring that point up more strongly?-jeffOn Aug 19, 2023, at 3:10 PM, Jason Ish ***@***.***> wrote:﻿
What happens when you have 2 inner vlan's with the same ID, but different outer IDs?  Or is that not a normal deployment scenario?

catenacyber:
Suricata PR for reference is https://github.com/OISF/suricata/pull/9401

catenacyber:
https://github.com/OISF/suricata/pull/10015

catenacyber:
Should this have `min-version: 8` ?

catenacyber (changes requested):
Should be green for 7, right ?

jlucovsky:
Closed for further evaluation.

PR #1358, opened 2023-08-23
tests: add a test for bug 6278

lukashino:
PR tests "run-as":
  - non-existent user
  - NULL user (empty user string)

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6278


lukashino:
Continues in #1412

PR #1370, opened 2023-08-31

coledishington:

Bug: #6260

Redmine ticket:


tests/flow-pkt-recursion/flow-pkt-recursion-terminated-excluded/test.yaml
```diff
@@ -0,0 +1,19 @@
+# Exclude recursion-level from flow matching
+# Re-use the pcap from the middleware-included test.
+command: |
+  ${SRCDIR}/src/suricata -c "${SRCDIR}/suricata.yaml" -l "${OUTPUT_DIR}" \
+   -r "${TEST_DIR}/../flow-pkt-recursion-terminated-included/test.pcap" \
+   --set "threshold-file=${SRCDIR}/threshold.config" \
+   --set "classification-file=${SRCDIR}`[ -f ${SRCDIR}/etc/classification.config ] && printf '/etc'`/classification.config" \
+   --set "reference-config-file=${SRCDIR}`[ -f ${SRCDIR}/etc/reference.config ] && printf '/etc'`/reference.config" \
+   --set 'decoder.recursion-level.use-for-tracking=false'
```
catenacyber, line 9: You can use `args` instead of the whole command here

(and `pcap:` to reuse another pcap)

tests/flow-pkt-recursion/flow-pkt-recursion-middleware-excluded/test.yaml
```diff
@@ -0,0 +1,18 @@
...
+   --set "reference-config-file=${SRCDIR}`[ -f ${SRCDIR}/etc/reference.config ] && printf '/etc'`/reference.config" \
+   --set 'decoder.recursion-level.use-for-tracking=false'
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        proto: ICMP
+        flow.pkts_toserver: 1
```
catenacyber, line 17: There does not seem to be a difference with flow-pkt-recursion-middleware-included/test.yaml

Is there ?

coledishington: Correct, there is no difference in the expected flow matching result when the Suricata device is not a tunnel terminator. This is what flow-pkt-recursion-middleware-excluded and flow-pkt-recursion-middleware-included are showing, that for flows on the same recursion level there if no difference with this new config option decoder.recursion-level.use-for-tracking disabled (it is enabled by default).

coledishington:
> You can use `args` instead of the whole command here
> 
> (and `pcap:` to reuse another pcap)
Thanks for the comments. Applied change in new pull request https://github.com/OISF/suricata-verify/pull/1373.

tests/flow-pkt-recursion/flow-pkt-recursion-middleware-excluded/test.yaml
catenacyber: Ok. Is this no difference in behavior not proved by existing tests already ?..

coledishington: Thanks for your comments. The existing test would need to be using tunneled unencrypted packets and explicitly check flow matching to already be proved. A test that matches this behavior might exist, but I thought it was specific enough traffic to add in a test for this feature. Happy to remove if you don't agree.

catenacyber: Thanks for the explanation, let's keep it :-)

PR #1373, opened 2023-09-04

coledishington:

Bug: #6260

Redmine ticket:


jufajardini:
Test readme looks well explained to me.
Tests are ensuring no dedups.

catenacyber:
I do not know about the suricata PR, but I think this SV PR would need to have `min-version: 8`

coledishington:
Rebased and updated PR in https://github.com/OISF/suricata-verify/pull/1597

PR #1390, opened 2023-09-19
Adds test about mime when stream depth is reached

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6367

Suricata PR incoming

### DIFF
--- a/tests/mime/mime-stream-depth/README.md
+++ b/tests/mime/mime-stream-depth/README.md
@@ -1,0 +1,8 @@
+# Description
+
+Test mime base64 decoding completeness when stream depth is reached
+See https://redmine.openinfosecfoundation.org/issues/6367
+
+# PCAP
+
+The pcap comes from QA TLPW1
--- a/tests/mime/mime-stream-depth/test.yaml
+++ b/tests/mime/mime-stream-depth/test.yaml
@@ -1,0 +1,15 @@
+requires:
+  min-version: 7
+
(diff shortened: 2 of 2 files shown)


tests/mime/mime-stream-depth/README.md
```diff
@@ -0,0 +1,8 @@
+# Description
+
+Test mime base64 decoding completeness when stream depth is reached
```
inashivb, line 3: Is this on the standard stream depth setting for now? 64 mb?

While I see that we're not trying to store any files, would filestore's stream depth settings also affect this (as it should ideally override stream's depth settings)?

So, should we have a suricata.yaml in this test? Or would we never change the standard settings for depths? wdyt?

catenacyber: Yes, it is for standard depth which is 1Mb

> would filestore's stream depth settings also affect this (as it should ideally override stream's depth settings)?

Maybe

> So, should we have a suricata.yaml in this test? 

Yes, or `--set stream.depth=1mb`

catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1393

tests/mime/mime-stream-depth/README.md
catenacyber: `stream.reassembly.depth` in facts

PR #1393, opened 2023-09-22
Adds test about mime when stream depth is reached

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6367

Suricata PR is https://github.com/OISF/suricata/pull/9502

#1390 with `stream.reassembly.depth=1mb` enforced by the test instead of relying on default configuration value


victorjulien:
Merged in #1403, thanks!

PR #1407, opened 2023-10-03
tests: add JA4 test cases

satta:
This PR introduces test cases for TLS and QUIC JA4 calculation.

## Ticket


tests/ja4-quic/test.yaml
```diff
@@ -0,0 +1,11 @@
...
+  min-version: 7.0.0
+  files:
+    - rust/src/quic/ja4.rs
+
+checks:
+  - filter:
+      count: 11
+      match:
+        event_type: quic
+        quic.ja4: q13d0310h3_55b375c5d22e_cd85d2d88918
```
catenacyber, line 11: How can we have `q` and `h` ? Is this not hexadecimal only ?

catenacyber: Could you please add a README for these new tests ?

satta: > How can we have `q` and `h` ? Is this not hexadecimal only ?

No, it isn't. `q` means QUIC (in contrast to `t` which is TLS) and `h3` is part of the advertised ALPN data. See https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4.md for more details.

satta: > Could you please add a README for these new tests ?

Yeah, can do.

satta:
Next PR: #1408

PR #1412, opened 2023-10-05
tests: add a test for bug 6278 v2

lukashino:
PR tests "run-as":
  - non-existent user
  - NULL user (empty user string)

Follow-up of #1358 
Changes from the previous PR:
- rebased

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6278

PR #1413, opened 2023-10-06
tests: add a test for bug 6278 v3

lukashino:
PR tests for "run-as":
- non-existent user
- NULL user (empty user string)

Follow-up of https://github.com/OISF/suricata-verify/pull/1412

Changes from the previous PR:
- new docs
- min-version

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6278

### DIFF
--- a/tests/bug-6278-1/README.md
+++ b/tests/bug-6278-1/README.md
@@ -1,0 +1,8 @@
+# Test Description
+
+Test to make sure Suricata handles well non-existent user as an input 
+in the user field.
+
+## Related Issue
+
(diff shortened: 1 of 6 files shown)


jufajardini (approved):
Overall structure looks good to me, changing review status only because if the Error message is changed, this will need update, too.

victorjulien (changes requested):
needs rebase

catenacyber:
@lukashino looks like this was merged already, right ?
Could you close this PR then ?

lukashino:
Indeed, sorry forgot about this.

PR #1415, opened 2023-10-10
Adds a test about filestore

catenacyber:
## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6388

Requires https://github.com/OISF/suricata/pull/9586

PR #1416, opened 2023-10-10
Filestore 6388 v2

catenacyber:
## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6388
https://redmine.openinfosecfoundation.org/issues/6390
https://redmine.openinfosecfoundation.org/issues/6392

Requires https://github.com/OISF/suricata/pull/9586

https://github.com/OISF/suricata-verify/pull/1415 with another test


catenacyber:
I guess I need the test to not test the eve.json but rather that there is no stored file with a wrong hash.

@inashivb how can I do that ?

catenacyber:
If there is a file `output//filestore/eb/eb076a2ec6ced9ee2e823e098446513cf5b2bb60fbcb04e6c85dc23dedaa414a` its hash must be eb076a2ec6ced9ee2e823e098446513cf5b2bb60fbcb04e6c85dc23dedaa414a (and its size 2048 bytes), when on master6 it is 0 bytes

inashivb:
> I guess I need the test to not test the eve.json but rather that there is no stored file with a wrong hash.
> 
> @inashivb how can I do that ?

You can use shell checks with `test` command. Example: https://github.com/OISF/suricata-verify/blob/master/tests/filestore-alert-log/test.yaml#L12

catenacyber:
Thanks Shivani Replaced by https://github.com/OISF/suricata-verify/pull/1427

PR #1418, opened 2023-10-11
mqtt: add test case for protocol string keyword

satta:
This PR adds a new MQTT test that checks the new `mqtt.connect.protocolstring` keyword that is currently not tested anywhere else.

We only run the test when the associated detection source is available in the tested source tree.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6396


catenacyber:
Requires https://github.com/OISF/suricata/pull/9600 to run and not be skipped

tests/mqtt-connect-rules/test.yaml
```diff
@@ -0,0 +1,60 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
+  files:
+    - rust/src/mqtt/parser.rs
```
catenacyber, line 5: We can just do with the first one

tests/mqtt-connect-rules/test.yaml
```diff
@@ -0,0 +1,60 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
```
catenacyber, line 3: We do not need to specify this feature anymore

tests/mqtt-connect-rules/test.yaml
```diff
@@ -0,0 +1,60 @@
...
+      count: 1
+      match:
+        event_type: alert
+        alert.signature: MQTT CONNECT protocol string SUCCESS
+
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature: MQTT CONNECT protocol string SUCCESS2
```
catenacyber, line 60: Should you test the absence of  sid:3 ?

tests/mqtt-connect-rules/test.yaml
satta: Hmm, the first one? I'd say that the presence of `src/detect-mqtt-connect-protocol-string.c` should be enough to indicate that the feature can be tested. I'll leave that one in.

catenacyber: Indeed. I meant just keep `src/detect-mqtt-connect-protocol-string.c`

tests/mqtt-connect-rules/test.yaml
satta: Done

satta:
Next PR: #1459

PR #1423, opened 2023-10-14
tests: add rule to check for tcp/ack - v1

0xEniola:
Test for rule type for tcp-ack keyword.

Related to
Issue: #6354

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6354

Suricata PR: https://github.com/OISF/suricata/pull/9608

### DIFF
--- a/tests/rules/tcp-ack-keyword/test.yaml
+++ b/tests/rules/tcp-ack-keyword/test.yaml
@@ -1,0 +1,14 @@
+requires:
+    min-version: 7.0.0
+    pcap: false
+    
+args:
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 1
+      lists.packet.matches[0].name: "tcp.ack"
--- a/tests/rules/tcp-ack-keyword/test.rules
+++ b/tests/rules/tcp-ack-keyword/test.rules
@@ -1,0 +1,1 @@
+alert tcp any any -> any any (msg:"Testing ack";ack:273;sid:1;)


tests/rules/tcp-ack-keyword/test.yaml
```diff
@@ -0,0 +1,14 @@
...
+args:
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 1
+      lists.packet.matches[0].name: "tcp.ack"
```
jufajardini, line 14: Would be good to have another line check here with the actual value that `ack` has in the line, (cf https://github.com/OISF/suricata-verify/tree/master/tests/rules/ipopts or other tests there, many of them check not only for the keyword itself, but also any value assigned to it.)

jufajardini (changes requested):
Nice! One comment from my end.

I've linked this PR on your Suricata PR description, so that our CI checks will tests your changes against it. Let's see how that goes!

Meanwhile, I'll try to give some thought on naming for the ack value. Maybe we could have just `value`. 🤔  I'll add that suggestion to the Suri PR, and then you can update the SV test accordingly :)

tests/rules/tcp-ack-keyword/test.yaml
0xEniola: Yes! I wanted to do that, but looking at the codes, I didn't see any parameter that would be used for it.

So I assumed it didn't exist.

inashivb: I don't think this test tests what you have done in your Suricata PR. That should have made this test fail. `tcp.ack` already exists in Suricata's matches which is why the rules as you have written work.
You should have checked for (alongwith what you already have) `lists.packet....ack.ack: 273`.

> I didn't see any parameter that would be used for it.

Because you just added it in your Suricata PR ;)

PR #1414, opened 2023-10-10
Bug 5578/s/v2

inashivb:
## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5578

Previous PR: https://github.com/OISF/suricata-verify/pull/1285

Changes since v1:
- rebase

PR #1424, opened 2023-10-16
Bug 5578/s/v3

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5578

Previous PR: https://github.com/OISF/suricata-verify/pull/1414

Changes since v2:
- Fixed test reference to the one that actually contains the pcap
- added Readme


catenacyber:
Could we have a version that passes with min-version 7 ?
And then, we could decrease it if we fix https://redmine.openinfosecfoundation.org/issues/5578

PR #1433, opened 2023-10-21
tests: add rule to check for tcp_mss - v1

0xEniola:
Test for rule type for tcp-mss keyword

Related to
Issue: #6355

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6355

Suricata PR: https://github.com/OISF/suricata/pull/9673

### DIFF
--- a/tests/rules/tcp-mss/test.yaml
+++ b/tests/rules/tcp-mss/test.yaml
@@ -1,0 +1,14 @@
+requires:
+    min-version: 7.0.0
+    pcap: false
+
+args:
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+#      list.packet.matches[0].name: "tcp.mss"
+      list.packet.matches[0].mss.size: 536
--- a/tests/rules/tcp-mss/test.rules
+++ b/tests/rules/tcp-mss/test.rules
@@ -1,0 +1,1 @@
+alert tcp any any -> any any (msg:"Testing mss"; tcp.mss:<50; sid:1;)


tests/rules/tcp-mss/test.yaml
```diff
@@ -0,0 +1,14 @@
...
+args:
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+#      list.packet.matches[0].name: "tcp.mss"
+      list.packet.matches[0].mss.size: 536
```
jufajardini, line 14: This check will fail with the rule you've provided ;)

jufajardini (changes requested):
Right path, but two adjustments needed ;)

tests/rules/tcp-mss/test.yaml
0xEniola: Actually, I have tried with other working tests, I don't necessarily have to test for `name`, and the check still passes.

But for this tcp-mss.
Even testing for `name` doesn't work.

That's where the issue is, I don't get the reason why it does not even recognise `tcp.mss` as the name, that's why I did that.

Because normally, it will pass even without checking for name matches. 

So I feel there's an issue with my code.

0xEniola:
When suricata-verify is done running, it says “FAILED”, but shows that “Rule matches on packet” in the `rules_analysis.txt` file.

jufajardini:
> When suricata-verify is done running, it says “FAILED”, but shows that “Rule matches on packet” in the `rules_analysis.txt` file.

Can you share the complete output for that, please? I'll double check your code and see if I can figure out what's going on.

0xEniola:
> > When suricata-verify is done running, it says “FAILED”, but shows that “Rule matches on packet” in the `rules_analysis.txt` file.
> 
> Can you share the complete output for that, please? I'll double check your code and see if I can figure out what's going on.

Which of the output? `rules. json` or `rules_analysis.txt`?

tests/rules/tcp-mss/test.yaml
jufajardini: I got it. It's not with the code.
The check should be for `lists`, not `list` ;)
Also, as I pointed out, the `size` check won't work with the rule you've added for this test.

jufajardini: When such failures occur, what I do is open and compare my changes with a similar test, and carefully look for possible differences. That's what I did for this case. ;)

jufajardini:
> > > When suricata-verify is done running, it says “FAILED”, but shows that “Rule matches on packet” in the `rules_analysis.txt` file.
> > 
> > 
> > Can you share the complete output for that, please? I'll double check your code and see if I can figure out what's going on.
> 
> Which of the output? `rules. json` or `rules_analysis.txt`?

The one you mentioned ;)

But I think at least we've figured why the test was failing.

tests/rules/tcp-mss/test.yaml
0xEniola: > I got it. It's not with the code.
> The check should be for `lists`, not `list` ;)
> Also, as I pointed out, the `size` check won't work with the rule you've added for this test.

Wiiiiiiii!

Thank you for catching that 😂.
It's really funny how the issues always be right in my face but still subtle.

But, why isn't the `size` check working though? I've looked through it as well; it says `Rule matches on packets` rules_analysis.txt but still fails.

0xEniola: > When such failures occur, what I do is open and compare my changes with a similar test, and carefully look for possible differences. That's what I did for this case. ;)

I'll do this more.

Thank you.

0xEniola: Here is the output from `rules_analysis.txt` regarding the `size` check.

```
-------------------------------------------------------------------
Date: 22/10/2023 -- 22:38:27
-------------------------------------------------------------------
== Sid: 1 ==
alert tcp any any -> any any (msg:"Testing mss"; tcp.mss:<536; sid:1;)
    Fast Pattern analysis:
        No content present

== Sid: 1 ==
alert tcp any any -> any any (msg:"Testing mss"; tcp.mss:<536; sid:1;)
    Rule is packet inspecting.
    Rule matches on packets.
    Warning: Rule is inspecting both the request and the response.

============
Summary:
============```

0xEniola: OK! I was tinkering around, and figured I'd have to include a test for all the options before the test could pass, which means a test for the `mode`, `arg1` and `arg2`.

Which means, for each option, the code to log them must be present.

So I did `jb_set_uint` for each of them.

And found out `mode: 1` = `<`, `mode: 3` = `>`, and `mode: 5` is used for `min-max`.

0xEniola: And based off of what you did in the ipopts task; I was thinking if I'd have to write additional code in `detect-tcpmss.c` to convert the integer value of the `mode` option, into string literals (the corresponding symbols/modes); similar to what you did with `IpOptsFlagToString` function.

jufajardini: > Here is the output from `rules_analysis.txt` regarding the `size` check.
> 
> ```
> -------------------------------------------------------------------
> Date: 22/10/2023 -- 22:38:27
> -------------------------------------------------------------------
> == Sid: 1 ==
> alert tcp any any -> any any (msg:"Testing mss"; tcp.mss:<536; sid:1;)
>     Fast Pattern analysis:
>         No content present
> 
> == Sid: 1 ==
> alert tcp any any -> any any (msg:"Testing mss"; tcp.mss:<536; sid:1;)
>     Rule is packet inspecting.
>     Rule matches on packets.
>     Warning: Rule is inspecting both the request and the response.
> 
> ============
> Summary:
> ============
> ```

Ok, thanks, I see what you meant, now. So, as you are working with the `engine-analysis` run mode, what you are seeing are the characteristics of this rule, not a detection output. What I mean by this is that the line you've mentioned indicates that this is a _rule type that matches on packets_. 

Does this clarify this bit?

jufajardini: > OK! I was tinkering around, and figured I'd have to include a test for all the options before the test could pass, which means a test for the `mode`, `arg1` and `arg2`.
> 
> Which means, for each option, the code to log them must be present.
> 
> So I did `jb_set_uint` for each of them.
> 
> And found out `mode: 1` = `<`, `mode: 3` = `>`, and `mode: 5` is used for `min-max`.

I'm not sure if `arg2` is used for `tcp.mss`, but if it is, it is indeed better that you add a test to showcase that.
Great catch about the `mode` bit. As I am not familiar with this type of keyword, I wondered how the `>/</` and similar was being represented, but didn't investigate further. This should definitely be present, thanks for pointing out! It also means there's more work to be done on the Suricata side of this task, as you will have to first log these fields, if you want to test the output here. I'll update my review there! :)

jufajardini: > But, why isn't the size check working though? I've looked through it as well; it says Rule matches on packets rules_analysis.txt but still fails.

It's not working, because the rule is for `<50` and you're testing against size `536`. Your size fields are different.

The rule_analysis part I just explained above.

jufajardini: > And based off of what you did in the ipopts task; I was thinking if I'd have to write additional code in `detect-tcpmss.c` to convert the integer value of the `mode` option, into string literals (the corresponding symbols/modes); similar to what you did with `IpOptsFlagToString` function.

Yes, please, as I've mentioned in another comment, that will indeed be needed.

0xEniola: > > But, why isn't the size check working though? I've looked through it as well; it says Rule matches on packets rules_analysis.txt but still fails.
> 
> It's not working, because the rule is for `<50` and you're testing against size `536`. Your size fields are different.
> 
> The rule_analysis part I just explained above.

Yes! I'm aware.

I was still tinkering around when I submitted this.

jufajardini: Thanks for asking this. 
Any reason not to add the full sentence? It feels to me that this makes it slightly harder to understand, in a situation where I'm not sure there's a need for concision - since this mode is only used for testing, so output size isn't so critical, I think 🤔

0xEniola: > Thanks for asking this. 
> Any reason not to add the full sentence? It feels to me that this makes it slightly harder to understand, in a situation where I'm not sure there's a need for concision - since this mode is only used for testing, so output size isn't so critical, I think 🤔 

Oh yes! Actually, understanding is paramount.

I'll change it to full sentences then.
- `less than`
- `greater than`
- `range`

PR #1434, opened 2023-10-21
tests: suricata hint missing runmode v1

comfort619:
Ticket: 5711

Test for missing suricata capture runmode hint

Suricata PR: https://github.com/OISF/suricata/pull/9674

### DIFF
--- a/tests/runmode_test/README.md
+++ b/tests/runmode_test/README.md
@@ -1,0 +1,14 @@
+# Suricata Missing Runmode Issue
+
+This repository contains test files and documentation for addressing the "Missing runmode" issue 5711 in Suricata.
+
+## Issue Description
+
+The issue relates to Suricata not providing adequate feedback when the runmode is missing, leading to confusion for users.
+
+## Test Files
+
+- `test.yaml`: This file defines a verification test that checks for the "Missing runmode" error.
+- `runmode_test`: A directory where you can place any necessary configuration files or rules for the test.
+
+
--- a/tests/runmode_test/test.yaml
+++ b/tests/runmode_test/test.yaml
@@ -1,0 +1,10 @@
+requires:
+  min-version: 7.0.0
+  pcap: false
+
+checks:
+  - shell:
+      args: suricata
+      expect: 1  # Expect an exit code of 1, indicating an error
+      expect-output: "ERROR: Missing runmode"  # Verify that the error message is present
+


tests/runmode_test/README.md
```diff
@@ -0,0 +1,14 @@
...
+## Issue Description
+
+The issue relates to Suricata not providing adequate feedback when the runmode is missing, leading to confusion for users.
+
+## Test Files
+
+- `test.yaml`: This file defines a verification test that checks for the "Missing runmode" error.
+- `runmode_test`: A directory where you can place any necessary configuration files or rules for the test.
+
+
```
jufajardini, line 14: It's a great practice to have a README file, but these explanations are very generic, so we can keep these out. :)

tests/runmode_test/test.yaml
```diff
@@ -0,0 +1,10 @@
+requires:
+  min-version: 7.0.0
+  pcap: false
+
+checks:
+  - shell:
+      args: suricata
```
jufajardini, line 7: Since the Suricata verify suite will run Suricata, you don't have to invoke shell to run it.

Instead, the suricata-verify way would be to have a line before the checks where you would pass the args for Suricata to run with. Check for instance https://github.com/OISF/suricata-verify/blob/master/tests/config-includes-array/test.yaml#L6-L7, these means that Suricata-verify will run Suricata as
`suricata --dump-config -c suricata.yaml`.

The `-c suricata.yaml` part is added by suricata-verify, which will use a suricata.yaml file present in the added test directory, if you provide one.

tests/runmode_test/test.yaml
```diff
@@ -0,0 +1,10 @@
...
+      expect: 1  # Expect an exit code of 1, indicating an error
```
jufajardini, line 8: When you want to indicate that a test expects Suricata to exit with error code 1, you should add an `exit-code: 1` line, outside the checks. See:
https://github.com/OISF/suricata-verify/blob/master/tests/detect-strip_whitespace-01/test.yaml#L5C1-L5C1

tests/runmode_test/test.yaml
```diff
@@ -0,0 +1,10 @@
...
+      expect-output: "ERROR: Missing runmode"  # Verify that the error message is present
+
```
jufajardini, line 10: We don't have a check option like that in Surica-verify, you could have a look at https://github.com/OISF/suricata-verify/pull/1431/files to see how to check for log output error messages in SV tests :)

jufajardini (changes requested):
Good try at creating a test like this! 

We must adjust it to work with how Suricata-verify expects and runs things, though ;)

tests/runmode_test/test.yaml
comfort619: Okay, thank you for the feedback

PR #1438, opened 2023-10-24
tests: add rule type check for flowbits

hadiqaalamdar:
Task#6309
Related to
Issue: #6309

Suricata PR: https://github.com/OISF/suricata/pull/9685

### DIFF
--- a/tests/rules/flowbits/test.rules
+++ b/tests/rules/flowbits/test.rules
@@ -1,0 +1,5 @@
+alert ip any any -> any any (msg:"isset rule need an option"; flowbits:isset; sid:1;)
+alert ip any any -> any any (msg:"isnotset rule need an option"; flowbits:isnotset; sid:2)
+alert ip any any -> any any (msg:"set rule need an option"; flowbits:set; sid:3;)
+alert ip any any -> any any (msg:"unset rule need an option"; flowbits:unset; sid:4;)
+alert ip any any -> any any (msg:"toggle rule need an option"; flowbits:toggle; sid:5;)
--- a/tests/rules/flowbits/test.yaml
+++ b/tests/rules/flowbits/test.yaml
@@ -1,0 +1,43 @@
+requires:
+    min-version: 7.0.0
+    pcap: false
+
+args:
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 1
+      lists.packet.matches[0].name: "flowbits"
+      lists.packet.matches[0].flowbits.option: "isset"
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 2
+      lists.packet.matches[0].name: "flowbits"
+      lists.packet.matches[0].flowbits.option: "isnotset"
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 3
+      lists.packet.matches[0].name: "flowbits"
+      lists.packet.matches[0].flowbits.option: "set"
+- filter:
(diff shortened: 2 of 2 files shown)


tests/rules/flowbits/test.rules
```diff
@@ -0,0 +1,5 @@
+alert ip any any -> any any (msg:"isset rule need an option"; flowbits:isset; sid:1;)
+alert ip any any -> any any (msg:"isnotset rule need an option"; flowbits:isnotset; sid:2)
+alert ip any any -> any any (msg:"set rule need an option"; flowbits:set; sid:3;)
+alert ip any any -> any any (msg:"unset rule need an option"; flowbits:unset; sid:4;)
+alert ip any any -> any any (msg:"toggle rule need an option"; flowbits:toggle; sid:5;)
```
jufajardini, line 5: Let's also have a rule for the `flowbits:noalert` option, please :)

tests/rules/flowbits/test.rules
```diff
@@ -0,0 +1,5 @@
+alert ip any any -> any any (msg:"isset rule need an option"; flowbits:isset; sid:1;)
+alert ip any any -> any any (msg:"isnotset rule need an option"; flowbits:isnotset; sid:2)
```
jufajardini, line 2: In addition to the lack of a name for the flowbits variable, this rule would also result in a parse error because it lacks the `;` after the rule sid. Each field within the parenthesis must be followed by those. ;)

tests/rules/flowbits/test.rules
```diff
@@ -0,0 +1,5 @@
+alert ip any any -> any any (msg:"isset rule need an option"; flowbits:isset; sid:1;)
```
jufajardini, line 1: As seen in the flowbits documentation, this keyword needs two values, so all the rules written here will result in a parse error due to the lack of the second field.

(cf https://docs.suricata.io/en/latest/rules/flow-keywords.html#flowbits)

Tip: when an SV test fails with `FAILED: got exit code 1, expected 0` this indicates that Suricata finished with an error state, so to debug such cases you should check the `output` directory in the test folder and see what's in `stderr`.

tests/rules/flowbits/test.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 1
+      lists.packet.matches[0].name: "flowbits"
+      lists.packet.matches[0].flowbits.option: "isset"
```
jufajardini, line 15: The checks use `flowbits.option`, but the Suri code change has `flowbits.cmd`, which results in a check failure with SV.

jufajardini (changes requested):
Thanks for submitting this PR, as I said, it's much easier to review and offer feedback.

Please see inline comments. :)

In the commit message, can you please add a space between Task and ticket number? Thanks a lot!

tests/rules/flowbits/test.rules
hadiqaalamdar: I'm not sure what to pass as the second value? I have changed the code for DETECT-FLOWBITS case so that the name is being stored in `ored_variables` array, should I pass this array as the second value?

jufajardini: The second value is a variable name, and as such it's not predefined. Just like a code variable, it is user-defined. Does this clear up?

hadiqaalamdar:
Work continued on: https://github.com/OISF/suricata-verify/pull/1439

PR #1442, opened 2023-10-25
tests: add a test for missing runmode v2

comfort619:
Ticket: 5711

Test for missing suricata capture runmode hint
Incoporated previous feedback on  running test

Previous PR: https://github.com/OISF/suricata-verify/pull/1434
Suricata PR: https://github.com/OISF/suricata/pull/9692

### DIFF
--- a/tests/runmode_test/README.md
+++ b/tests/runmode_test/README.md
@@ -1,0 +1,10 @@
+# Suricata Missing Runmode Issue
+
+This directory contains test files and documentation for addressing the "Missing runmode" issue 5711 in Suricata.
+
+## Issue Description
+
+The issue relates to Suricata not providing adequate feedback when the runmode is missing, leading to confusion for users.
+
+
+
--- a/tests/runmode_test/suricata.yaml
+++ b/tests/runmode_test/suricata.yaml
@@ -1,0 +1,12 @@
+%YAML 1.1
+---
+
+runmodes:
+  - workers:
+      enabled: yes
+       
+  
+    
+
+
+
--- a/tests/runmode_test/test.yaml
+++ b/tests/runmode_test/test.yaml
@@ -1,0 +1,16 @@
+requires:
+  min-version: 7 
+  pcap: false
+
+exit-code: 1
+args:
+   - --dump-config
+
+checks:
+   - shell:
+       args: suricata -S /dev/null
+       expect: 1  
+
+         
+      
+                                           


tests/runmode_test/README.md
```diff
@@ -0,0 +1,10 @@
+# Suricata Missing Runmode Issue
+
+This directory contains test files and documentation for addressing the "Missing runmode" issue 5711 in Suricata.
+
+## Issue Description
+
+The issue relates to Suricata not providing adequate feedback when the runmode is missing, leading to confusion for users.
```
inashivb, line 7: nit: please add redmine ticket link here

tests/runmode_test/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+
+
```
inashivb, line 10: nit: unneeded whitespace

tests/runmode_test/suricata.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+runmodes:
+  - workers:
+      enabled: yes
+       
+  
+    
+
+
+
```
inashivb, line 12: nit: unneeded whitespace

tests/runmode_test/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+   - --dump-config
+
+checks:
+   - shell:
+       args: suricata -S /dev/null
+       expect: 1  
+
+         
+      
+
```
inashivb, line 16: nit: unneeded whitespace

tests/runmode_test/test.yaml
```diff
@@ -0,0 +1,16 @@
+requires:
+  min-version: 7 
+  pcap: false
+
+exit-code: 1
+args:
+   - --dump-config
```
inashivb, line 7: This is not needed for the test, it was an example given by Juliana. The args should be just like mentioned in the redmine ticket.

tests/runmode_test/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+
+checks:
+   - shell:
+       args: suricata -S /dev/null
+       expect: 1
```
inashivb, line 12: I don't understand this test. Could you please explain what were you trying to do here?
Note: This test does not seem to be testing what you've done in your Suricata PR, you should be looking for the error message in the Suricata logs/stderr. e.g. `tests/detect-bsize-01/test.yaml`
Hint: Recommended steps to follow:
1. Try to understand what `shell` checks are doing looking at other tests.
2. Check out the `output` dir created in your test after you run it. See what's inside each file.
3. Get back to fixing the test.

inashivb (changes requested):

tests/runmode_test/test.yaml
comfort619: Okay, I actually included this test to check whether Suricata behaves correctly when a specific run mode is missing. To do that, I ran Suricata with the '-S /dev/null' argument, which simulates the absence of the 'capture' run mode. I expected Suricata to return an error code (1) in this scenario, indicating that the run mode is missing.
But if I understand, I am expected to check for  the presence of the expected error message in Suricata logs.

comfort619: > I don't understand this test. Could you please explain what were you trying to do here? Note: This test does not seem to be testing what you've done in your Suricata PR, you should be looking for the error message in the Suricata logs/stderr. e.g. `tests/detect-bsize-01/test.yaml` Hint: Recommended steps to follow:
> 
> 1. Try to understand what `shell` checks are doing looking at other tests.
> 2. Check out the `output` dir created in your test after you run it. See what's inside each file.
> 3. Get back to fixing the test.

Okay, I actually included this test to check whether Suricata behaves correctly when a specific run mode is missing. To do that, I ran Suricata with the '-S /dev/null' argument, which simulates the absence of the 'capture' run mode. I expected Suricata to return an error code (1) in this scenario, indicating that the run mode is missing.
But if I understand, I am expected to check for  the presence of the expected error message in Suricata logs.

jufajardini: Thanks for going through your thought process :)

So, as I mentioned on the previous review, you don't have to invoke shell to run suricata in Suricata-verify tests. So, whatever arguments you want to pass for this test to run suricata with, those would passed in that `args` section where you've putt the `--dump-config` option, instead.

For the expected exit code, and how to check for an error message from the output - in case you don't want to check log outputs, you can check this test: https://github.com/OISF/suricata-verify/blob/master/tests/detect-strip_whitespace-01/test.yaml#L5C1-L5C1

Not that right above the first `args` section, there's an `exit-code: 1` line. That is how you tell suricata-verify that you expect Suri to exit with error code 1.

I hope this helps you move forward with this contribution! :)

comfort619: Okay, thanks. It helps

PR #1446, opened 2023-10-27
test: test for empty dns/eve formats

jasonish:
Test that when dns/eve "formats" is empty, it uses the default of all.

Bug: https://redmine.openinfosecfoundation.org/issues/6420


victorjulien:
Merged in #1448, thanks!

PR #1453, opened 2023-11-04
stats: test if the timed counters are calculated correctly in stats.log

I3run0:
Related to
Task OISF#6410

Redmine ticket link: https://redmine.openinfosecfoundation.org/issues/6410


tests/log-data-per-sec-interval-10/README.md
```diff
@@ -0,0 +1,4 @@
+# Test Description
+
+Test if the timed counters from the counters.c are calculated correctly in stats.log, 
+applying a stats interval of 10 secoonds
```
jufajardini, line 4: Could you also add a section indicating where the pcap comes from, and another for the redmine ticket? :)

tests/log-data-per-sec-interval-10/test.yaml
```diff
@@ -0,0 +1,24 @@
+requires:
+  min-version: 7.0.2
+  features:
+    - DEBUG
+  files:
+    - src/counters.c
+    - src/decode.c
```
jufajardini, line 7: I don't think this is necessary, for this test.

tests/log-data-per-sec-interval-10/test.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+pcap: input.pcap
```
jufajardini, line 9: This is used in case we want to indicate that the sv test doesn't require a pcap, or to inform Suricata-verify what's the pcap path - only if the test directory doesn't contain a pcap file.

This last case is what we must do if the test uses an existing pcap - we do that to avoid file duplication.

That explained, I've noticed that at least two of your tests have the same pcap. Could you please remove the pcap duplication for the other tests from this contribution? You can follow [this example](https://github.com/OISF/suricata-verify/blob/master/tests/base64-issue-5223-6/test.yaml#L1) for how to refer to a pcap from a different test.

tests/log-data-per-sec-interval-5/README.md
```diff
@@ -0,0 +1,4 @@
+# Test Description
+
+Test if the timed counters from the counters.c are calculated correctly in stats.log, 
+applying a stats interval of 5 secoonds
```
jufajardini, line 4: nit: typo - *seconds*

tests/log-data-per-sec-interval-10/README.md
jufajardini: For the descriptions in these tests, I think it's better if we use something more high-level.
So, something related to the pkts counters that are being tested, I guess?

jufajardini (changes requested):
Thanks for adding these tests!

Can you adjust the commit message to follow our guidelines? Use the same format you would for Suricata commit messages in terms of character limit, how to refer to ticket numbers etc ;)


Please check also the inline comments for other required changes.

tests/log-data-per-sec-interval-10/README.md
I3run0: Hi, 

I have added a random pcap file that is not special, so is it better use a file that already exists in the project?

I3run0: I'll think of a better description following your advice, thanks!

I3run0:
New pull request https://github.com/OISF/suricata-verify/pull/1477

PR #1464, opened 2023-11-15
tests: dns.answer.name, dns.query.name

jasonish:
Tickets:
- https://redmine.openinfosecfoundation.org/issues/6497
- https://redmine.openinfosecfoundation.org/issues/6496


tests/dns/dns-answer-name/test.rules
```diff
@@ -0,0 +1,3 @@
+alert dns any any -> any any (dns.answer.name; content:"suricata"; sid:1; rev:1;)
+alert dns any any -> any any (dns.answer.name; content:"suricata"; flow:to_client,established; sid:2; rev:1;)
+alert dns any any -> any any (dns.answer.name; content:"suricata"; flow:to_server,established; sid:3; rev:1;)
```
victorjulien, line 3: does this not match due to the dns keyword or due to "established" on 1st udp packet?

jasonish: This doesn't match as I don't have a crafted DNS request with an answer in it. However, it still wouldn't match due to the `established`, which I realized when I added the `dns.query.name` keyword afterwards, which doesn't use established as it prevents the request from matching, however works fine on the response.

jasonish: I've updated the test to use a pcap that has answers in the request. Fun fact: both Google and Cloudflare accept it as a request.

victorjulien (approved):

victorjulien:
Merged in #1544, thanks!

PR #1267, opened 2023-06-22
Adds a test about flow.pkts_toclient keyword

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6164

Modifies #1261 with more keywords tested

PR #1473, opened 2023-11-17
Check: start detect-flow-pkts at version 8

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6164

Follow up on #1267


victorjulien:
Addressed in #1471

PR #1487, opened 2023-11-20
Better support for nested directories - v1

jasonish:
First, if doing `run.py pattern` consider the directories to the test as part of the test name. For example you might have a directory `foobar` with tests named `one`, `two` and so on.  You'd want `run.py foobar` to run all those tests.

Second, handle more then one level of sub-directories when displaying the test name.

### DIFF
--- a/run.py
+++ b/run.py
@@ -786,10 +786,11 @@ class TestRunner:
 
         if not check_value["failure"] and not check_value["skipped"]:
             if not self.quiet:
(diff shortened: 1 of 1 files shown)


catenacyber:
Are you expecting a review Jason ?

jasonish:
Rebased at https://github.com/OISF/suricata-verify/pull/1577.

jasonish:
> Are you expecting a review Jason ?

Yeah, this being Python, working for me, and working for GitHub CI is not enough :)

PR #1490, opened 2023-11-23
ssh: do not test direction

catenacyber:
as we log packet direction instead of flow direction

## Ticket

Redmine ticket: should there be one ?


catenacyber:
Closing as unwanted cf https://github.com/OISF/suricata/pull/10300#issuecomment-1924468783

PR #1495, opened 2023-11-27
tests: add bug 1449 test

victorjulien:
Was private at the time but no need to keep it private.

https://redmine.openinfosecfoundation.org/issues/1449


tests/bug-1449-01/test.yaml
```diff
@@ -0,0 +1,17 @@
...
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 2220017
+  - filter:
+      count: 1
+      match:
+        event_type: smtp
+        smtp.helo: bug.client
+        email.status: PARSE_ERROR
```
catenacyber, line 16: Why do we want to enforce failure when we can succeed with the rust version ? (without overflowing)

victorjulien: point here is to describe existing behavior. We won't update 6 or 7 parsers to rust

catenacyber: So, should the check be done with lt-version: 8 ?

victorjulien: I guess only when behavior for 8 changes

catenacyber (approved):
Ok for me.

The commit message could be improved (mention smtp and mime) instead of just the redmine id

victorjulien:
Merged in #1495, thanks!

PR #1496, opened 2023-11-27
tests: update ips-state-1 test - v2

jufajardini:
This test indicated that there were FP drops for http and that another check was failing, but currently the test passes.

Previous PR: https://github.com/OISF/suricata-verify/pull/1227

Changes from previous PR:
- rebase
- add more checks
-- make test fail again, by checking for `flow.action: drop` for `tls` flow
- update README to indicate current test behavior

Question: do we expect to see `flow.action` set to `drop` here?

### DIFF
--- a/tests/ips-state-1/README.md
+++ b/tests/ips-state-1/README.md
@@ -1,13 +1,9 @@
 ## PCAP
 
-This PCAP contains 3 flows.  2 are http and one is TLS. The HTTP flows should
+This PCAP contains 3 flows. 2 are http and one is TLS. The HTTP flows should
 be full passed with no alerts, while the TLS flow should be dropped.
 
 ## Current Observations
 
-- HTTP response packets are being logged as dropped, however the transaction is
-  logged suggesting the drop is only in logging only, but not actually
-  occurring.
-
-- All the TLS packets apear to be getting dropped, but `flow.action` is never
-  set to true.
+- Test seems to indicate that Suricata mostly behaves as expected. BUT, although
+  we see TLS logged as dropped, the actual flow.action is never set to drop...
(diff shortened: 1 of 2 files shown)


catenacyber:
> Question: do we expect to see flow.action set to drop here?

Looks reasonable.

Disclaimer : I have never used this flow.action field
Can it be set to `drop` ?
Could you point me to some doc ?

jufajardini:
> > Question: do we expect to see flow.action set to drop here?
> 
> Looks reasonable.
> 
> Disclaimer : I have never used this flow.action field Can it be set to `drop` ? 

It can, check for instance https://github.com/OISF/suricata-verify/blob/master/tests/exception-policy-midstream-02/test.yaml#L22

> Could you point me to some doc ?

I'm afraid we don't have anything directly, we mention it when explaining Exception Policies, but it's rather indirect: https://docs.suricata.io/en/latest/configuration/exception-policies.html#auto

catenacyber:
Thanks.

What means `flow.action` ?
Is there another field like `flow.state` that can be set to drop ?

jufajardini:
> Thanks.
> 
> What means `flow.action` ? Is there another field like `flow.state` that can be set to drop ?

From what I understand, `flow.action` is... the action that the engine applies to the flow, based on rules, exception policies, thresholds... `flow.state` seems to represent something slightly different to me, a bit closer to connection states? `verdict` can be `verdict.action: drop`, but that's linked to packets.

catenacyber:
So, my feeling from someone who does not know this part is that `flow.action` should indeed be set to `drop`

jufajardini:
> So, my feeling from someone who does not know this part is that `flow.action` should indeed be set to `drop`

Thanks for helping in figuring out this one!
@victorjulien Would you agree?

catenacyber:
Can I help further here ?

jufajardini:
> Can I help further here ?

Thanks, I don't know exactly what to do with this check. I've created a new PR with more checks, and kepts the flow.action check, still unsure if this is a possible bug to investigate or not...

jufajardini:
Follow by: https://github.com/OISF/suricata-verify/pull/1781

PR #1505, opened 2023-12-01
test: test requires keyword - v1

jasonish:
Also includes readme update for files to exists as a requirement.

### DIFF
--- a/tests/requires/test.rules
+++ b/tests/requires/test.rules
@@ -1,0 +1,15 @@
+# Rule for Suricata >= 7 and < 8.
+alert http any any -> any any (msg:"TEST Suricata >= 7 and < 8"; content:"uid=0"; requires: version >= 7 < 8; sid:7; rev:1;)
+
+# Rule for Suricata >= 7.0.3 but less than 8... Or >= 8.0.1
+alert http any any -> any any (content:"uid=0"; requires: version >= 7.0.3 < 8 | >= 8.0.1; sid:9; rev:1;)
+
+# Rule for Suricata >= 8.
+alert http any any -> any any (msg:"TEST Suricata >= 8"; content:"uid=0"; requires: version >= 8.0.0; sid:8; rev:1;)
+
+# These rules have something invalid about them, but do follow the general rule
+# structure, so should be eliminated by the requires statement.
+alert vxlan any any -> any any (requires: version >= 10; sid:1;)
+alert udp any any -> any any (vxlan_vni:10; requires: version >= 10; sid:2;)
+alert http any any => any any (requires: version >= 10; sid:3;)
+alert tcp any any -> any any (frame:smtp.not_supported; requires: version >= 10; sid:4;)
--- a/tests/requires/test.yaml
+++ b/tests/requires/test.yaml
@@ -1,0 +1,44 @@
+requires:
+  files:
+    - rust/src/detect/requires.rs
+
+pcap: ../eve-metadata/testmyids.pcap
+
+checks:
+
+  # Check that for Suricata >= 8.0.0 we have one alert and that alert is sid 8.
+  - filter:
+      requires:
+        min-version: 8
+      count: 1
+      match:
+        event_type: alert
+  - filter:
+      requires:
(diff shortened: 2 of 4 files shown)


victorjulien:
Can you add a test with the example rules I shared in the previous PR?

tests/requires/test.yaml
```diff
@@ -0,0 +1,40 @@
...
+        min-version: 8
+      count: 1
+      match:
+        event_type: alert
+  - filter:
+      requires:
+        min-version: 8
+      count: 1
+      match:
+        alert.signature_id: 8
```
inashivb, line 21: These could be one filter. Same for below filters. Why were they written separately?

tests/requires/test.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+  # Check that for Suricata >= 8.0.0 we have one alert and that alert is sid 8.
```
inashivb, line 23: Typo in comment.

tests/requires/test.yaml
jasonish: I don't think so. We want to confirm 2 things:
- There is only 1 alert, regardless of SID
- That alert is SID 8

If these checks were to be combined, we'd be testing that there is 1 or more alerts, but only one has SID 8. But there could still be more alerts with other SIDs.

inashivb: ah. ok. thanks for explaining!

jasonish:
> Can you add a test with the example rules I shared in the previous PR?

Done.

victorjulien (approved):

tests/requires/test.rules
```diff
@@ -0,0 +1,15 @@
...
+
+# Rule for Suricata >= 7.0.3 but less than 8... Or >= 8.0.1
+alert http any any -> any any (content:"uid=0"; requires: version >= 7.0.3 < 8 | >= 8.0.1; sid:9; rev:1;)
+
+# Rule for Suricata >= 8.
+alert http any any -> any any (msg:"TEST Suricata >= 8"; content:"uid=0"; requires: version >= 8.0.0; sid:8; rev:1;)
+
+# These rules have something invalid about them, but do follow the general rule
+# structure, so should be eliminated by the requires statement.
+alert vxlan any any -> any any (requires: version >= 10; sid:1;)
```
catenacyber, line 12: I would put version 42 rather than 10 which is within our lifetimes I hope

tests/requires/test.rules
```diff
@@ -0,0 +1,15 @@
+# Rule for Suricata >= 7 and < 8.
+alert http any any -> any any (msg:"TEST Suricata >= 7 and < 8"; content:"uid=0"; requires: version >= 7 < 8; sid:7; rev:1;)
+
+# Rule for Suricata >= 7.0.3 but less than 8... Or >= 8.0.1
+alert http any any -> any any (content:"uid=0"; requires: version >= 7.0.3 < 8 | >= 8.0.1; sid:9; rev:1;)
```
catenacyber, line 5: Wow we allow complex "or" !

tests/requires/test.rules
```diff
@@ -0,0 +1,15 @@
...
+
+# Rule for Suricata >= 8.
+alert http any any -> any any (msg:"TEST Suricata >= 8"; content:"uid=0"; requires: version >= 8.0.0; sid:8; rev:1;)
+
+# These rules have something invalid about them, but do follow the general rule
+# structure, so should be eliminated by the requires statement.
+alert vxlan any any -> any any (requires: version >= 10; sid:1;)
+alert udp any any -> any any (vxlan_vni:10; requires: version >= 10; sid:2;)
+alert http any any => any any (requires: version >= 10; sid:3;)
+alert tcp any any -> any any (frame:smtp.not_supported; requires: version >= 10; sid:4;)
```
catenacyber, line 15: Could we also test `requires: version >= 10, superpower: notyet` ? (superpower being an unknown key yet for the `requires` keyword

victorjulien:
Merged in #1558, thanks!

PR #1498, opened 2023-11-27
Ssh newkeys 6578 v2

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6578

Suricata PR required https://github.com/OISF/suricata/pull/9903

PR #1507, opened 2023-12-05
Ssh newkeys 6578 v3

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6578

Suricata PR required https://github.com/OISF/suricata/pull/9903

#1498 with additional commit to fix test exception-policy-simulated-flow-memcap

@jufajardini do I miss something in the point of this test ?


jufajardini:
> #1498 with additional commit to fix test exception-policy-simulated-flow-memcap
> 
> @jufajardini do I miss something in the point of this test ?

Not as far as I can see. The important bit for us is to ensure that there is a drop if the memcap is hit, and since those checks are there, and passing, it looks still good to me.

catenacyber:
#1508 (ie the first part) should get merged first

catenacyber:
Rebased in https://github.com/OISF/suricata-verify/pull/1587

PR #1441, opened 2023-10-25
tests: adding rule type check for flowbits v4

hadiqaalamdar:
Task #6309
Related to
Issue: #6309

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6309
Previous PR: https://github.com/OISF/suricata-verify/pull/1439

Suricata PR: https://github.com/OISF/suricata/pull/9691

Output from console:
```
===> flowbits: Sub test #1: FAIL : expected 1 matches; got 0 for filter {'filename': 'rules.json', 'count': 1, 'match': {'id': 1, 'lists.packet.matches[0].name': 'flowbits', 'lists.packet.matches[0].flowbits.action': 'noalert'}}
===> flowbits: Sub test #8: FAIL : expected 1 matches; got 0 for filter {'filename': 'rules.json', 'count': 1, 'match': {'id': 8, 'lists.postmatch.matches[0].name': 'flowbits', 'lists.postmatch.matches[0].flowbits.cmd': 'isset', 'lists.postmatch.matches[0].flowbits.ored_flowbits': 2, 'lists.postmatch.matches[0].flowbits.ored_variables[0]': 'fb1', 'lists.postmatch.matches[0].flowbits.ored_variables[1]': 'fb2'}}
===> flowbits: Sub test #9: FAIL : expected 1 matches; got 0 for filter {'filename': 'rules.json', 'count': 1, 'match': {'id': 9, 'lists.packet.matches[0].name': 'flowbits', 'lists.packet.matches[0].flowbits.ored_flowbits': 1, 'lists.packet.matches[0].flowbits.cmd': 'isnotset', 'lists.packet.matches[0].flowbits.ored_variables[0]': 'fb1', 'lists.packet.matches[0].flowbits.ored_variables[1]': 'fb2'}}
```

PR #1512, opened 2023-12-05
tests: add rule type check for flowbits v5

hadiqaalamdar:
Task #6309
Related to
Issue: #6309

Previous PR: https://github.com/OISF/suricata-verify/pull/1441

Suricata PR: https://github.com/OISF/suricata/pull/9971

Output from console:
```

```


tests/rules/flowbits/test.yaml
```diff
@@ -0,0 +1,100 @@
+requires:
+    min-version: 7.0.0
```
jufajardini, line 2: For now, let's keep `min-version: 8.0`, as we're not sure if we will bring this work to suricata 7, yet :)

jufajardini (changes requested):
Nice work, most tests are passing, and the output seems to be very close to what we're expecting it to look like!

I've noticed that on the previous Suricata PR, Shivani had requested the addition of two other types of rules (cf https://github.com/OISF/suricata/pull/9691#issuecomment-1782340931), could you please ensure we have those cases covered here, too? :)

I have left a comment review on the Suricata PR about my guess on why is the last check for the tests here failing.

tests/rules/flowbits/test.yaml
hadiqaalamdar: got it.

PR #1514, opened 2023-12-06
tests/pgsql: update password log disabled msgs - v2

jufajardini:
Removing the white spaces from this log output, as these can cause issues with grepping commands querying log results, and also doesn't show a consistent behavior among different environments.

Previous PR: https://github.com/OISF/suricata-verify/pull/1511

Following Jason's reflection on why to log the disabled password log output at all, remove checks for those.


inashivb (approved):
LGTM. Thanks! :)

victorjulien:
Merged in #1517, thanks!

PR #1519, opened 2023-12-06
Websockets 2695 v1

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/2695

Requires suricata PR to come


catenacyber:
replaced by #1528

PR #1427, opened 2023-10-16
Filestore 6388 v3

catenacyber:
## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6388
https://redmine.openinfosecfoundation.org/issues/6390
https://redmine.openinfosecfoundation.org/issues/6392

Requires https://github.com/OISF/suricata/pull/9586

https://github.com/OISF/suricata-verify/pull/1416 with shell check for filestore

PR #1522, opened 2023-12-07
Adds a test about filestore

catenacyber:
## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6392

Requires Suricata PR to come

https://github.com/OISF/suricata-verify/pull/1427 with only test that currently fails master

min-version is set to 8
Backport to 7 should be clean
But backport to 6 would be completely different

I propose to do S-V PRs to lower the min-version to pass this test as Suricata PR get merged


catenacyber:
Replaced by #1524

PR #1524, opened 2023-12-07
Adds a test about filestore

catenacyber:
## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6390

Requires Suricata PR to come

https://github.com/OISF/suricata-verify/pull/1522 without shell check using `not` command that is not everywhere

min-version is set to 8
Backport to 7 should be clean


catenacyber:
Rebased in #1863

PR #1525, opened 2023-12-07
detect-engine-state: add tests

0xEniola:
Task: 6146


tests/engine-state/detect-engine-state-01/test.yaml
```diff
@@ -0,0 +1,51 @@
+# *** Add configuration here ***
```
inashivb, line 1: Please remove this

inashivb, line 1: Please remove this

0xEniola: Sure!

0xEniola:
> Hi @0xEniola!
> Nice job figuring out createst.py and scapy! :)
> 1. Could you please also mention the name of the complex unit test that you have converted?
> 2. Fix the README.md to reflect what you're trying to do in the test

For 1; I'm not really sure what the test is for, although, Test02 says “multiple pipelined http transactions” & Test01 & 02 are similar.

So I guess it'll be the same.

jufajardini:
> > Hi @0xEniola!
> > Nice job figuring out createst.py and scapy! :)
> > 
> > 1. Could you please also mention the name of the complex unit test that you have converted?
> > 2. Fix the README.md to reflect what you're trying to do in the test
> 
> For 1; I'm not really sure what the test is for, although, Test02 says “multiple pipelined http transactions” & Test01 & 02 are similar.
> 
> So I guess it'll be the same.

Yeah, some of our unit tests are rather lacking when it comes to descriptions :P
It takes some time for one to grasp all that happens, IMHO, another good measure that they're more complex than a unit test should be. 🤔 Doesn't mean we can't give it a try! :P

Sometimes, just trying to describe the more obvious things we see can be a good start. For instance, test 02 says multiple pipelined HTTP transactions, and we see two different HTTP methods, and that we're parsing the content as separate buffers. (this last part for both tests). We see that we only expect to see alerts after a certain extent of content has been parsed (as the first alert checks in both tests assert that there are _no_ alerts).

Besides that, I think that Shivani also means indicating, in this type of task, which unittests are being covered/converted, as we need this info to better review and compare.

Does this help?

0xEniola:
> > > Hi @0xEniola!
> > > Nice job figuring out createst.py and scapy! :)
> > > 
> > > 1. Could you please also mention the name of the complex unit test that you have converted?
> > > 2. Fix the README.md to reflect what you're trying to do in the test
> > 
> > For 1; I'm not really sure what the test is for, although, Test02 says “multiple pipelined http transactions” & Test01 & 02 are similar.
> > 
> > So I guess it'll be the same.
> 
> Yeah, some of our unit tests are rather lacking when it comes to descriptions :P
> It takes some time for one to grasp all that happens, IMHO, another good measure that they're more complex than a unit test should be. 🤔 Doesn't mean we can't give it a try! :P
> 
> Sometimes, just trying to describe the more obvious things we see can be a good start. For instance, test 02 says multiple pipelined HTTP transactions, and we see two different HTTP methods, and that we're parsing the content as separate buffers. (this last part for both tests). We see that we only expect to see alerts after a certain extent of content has been parsed (as the first alert checks in both tests assert that there are _no_ alerts).
> 
> Besides that, I think that Shivani also means indicating, in this type of task, which unittests are being covered/converted, as we need this info to better review and compare.
> 
> Does this help?
> 
> 

I'll try to read the tests more to understand it better, but no lies that it's quite complex.

And, yea; I now understand what Shivani meant, I didn't read it clearly before.

Plus, I didn't bother doing all the necessary edits because it was a draft PR.

jufajardini:
> Plus, I didn't bother doing all the necessary edits because it was a draft PR.

I figured, that's why I hadn't added a review ;P

0xEniola:
What about the alert event that createtest does not add?

inashivb:
> What about the alert event that createtest does not add?

Does the `eve.json` have it? If you run the test, there should be an `output` dir in the test dir that should hold eve.json. If eve has it and createst is not adding it, it's a bug with createst. Else, your test needs rework.

0xEniola:
> > What about the alert event that createtest does not add?
> 
> Does the `eve.json` have it? If you run the test, there should be an `output` dir in the test dir that should hold eve.json. If eve has it and createst is not adding it, it's a bug with createst. Else, your test needs rework.

There's an output in `eve.json`, but alert there says 0.

0xEniola:
OK! What I got to understand was; when the packet content is fragmented as in this PR I submitted, the alert event is not added, because based on `test.rules`, Suri only throws the alert when both content are in a single packet.

But since the contents have been fragmented, it doesn't throw the alert event. I think that's also what the unittest tried to achieve.

PR #1532, opened 2023-12-11
tests: add rule type check for flowbits v8

hadiqaalamdar:
Task #6309

Previous PR: https://github.com/OISF/suricata-verify/pull/1529

Suricata PR: https://github.com/OISF/suricata/pull/10026

Output from console:
```
===> rules/flowbits: OK

PASSED:  3
FAILED:  0
SKIPPED: 6
```
Output from stdout file:
```
Notice: suricata: This is Suricata version 8.0.0-dev (9a2432696 2023-12-11) running in USER mode [LogVersion:suricata.c:1147]
Warning: threshold-config: Error opening file: "/usr/local/etc/suricata//threshold.config": Permission denied [SCThresholdConfInitContext:util-threshold-config.c:177]
Warning: detect-flowbits: flowbit 'fb3' is checked but not set. Checked in 4 and 1 other sigs [DetectFlowbitsAnalyze:detect-flowbits.c:596]
Warning: detect-flowbits: flowbit 'fb4' is checked but not set. Checked in 5 and 0 other sigs [DetectFlowbitsAnalyze:detect-flowbits.c:596]
Warning: detect-flowbits: flowbit 'fb5' is checked but not set. Checked in 12 and 0 other sigs [DetectFlowbitsAnalyze:detect-flowbits.c:596]
```


inashivb (changes requested):
I believe this is missing @jufajardini 's review update from the last PR. ref: https://github.com/OISF/suricata-verify/pull/1529#discussion_r1420988638

Please fix that. From the first look, it's looking good otherwise :)

hadiqaalamdar:
> I believe this is missing @jufajardini 's review update from the last PR. ref: [#1529 (comment)](https://github.com/OISF/suricata-verify/pull/1529#discussion_r1420988638)
> 
> Please fix that. From the first look, it's looking good otherwise :)

Sorry, I just saw the comment. I was wondering if we need a rule for `noalert` now if we've already removed that field from detect-flowbits files?

jufajardini:
> > I believe this is missing @jufajardini 's review update from the last PR. ref: [#1529 (comment)](https://github.com/OISF/suricata-verify/pull/1529#discussion_r1420988638)
> > Please fix that. From the first look, it's looking good otherwise :)
> 
> Sorry, I just saw the comment. I was wondering if we need a rule for `noalert` now if we've already removed that field from detect-flowbits files?

I think we could keep it. My reasoning for this is that with this `engine-analysis` work we are also documenting and showcasing things, and therefore there is value in showing what will it look like if you use `flowbits:noalert`.

Adding the check to the `test.yaml` file will help show where to find information about that, too.

hadiqaalamdar:
> I think we could keep it. My reasoning for this is that with this `engine-analysis` work we are also documenting and showcasing things, and therefore there is value in showing what will it look like if you use `flowbits:noalert`.
> 
> Adding the check to the `test.yaml` file will help show where to find information about that, too.

Okay, I'm not sure how to alter the yaml file such that the `noalert` rule doesn't create an error.

jufajardini:
> > I think we could keep it. My reasoning for this is that with this `engine-analysis` work we are also documenting and showcasing things, and therefore there is value in showing what will it look like if you use `flowbits:noalert`.
> > Adding the check to the `test.yaml` file will help show where to find information about that, too.
> 
> Okay, I'm not sure how to alter the yaml file such that the `noalert` rule doesn't create an error.

That's ok, we're figuring this one out as we progress :)

From the eve log from the previous PR (https://github.com/OISF/suricata-verify/pull/1529#discussion_r1421008397), I _think_ that we can:
i) have a rule that actually sets a flowbit, but also calls `flowbit:noalert` (just to ensure we're using a rule that makes some sense :P
ii) besides the check for the set flowbit, we can have something like:
```json
flag[4]: "noalert"
```
I've tried a local test, and this seems to work.
What do you think?

hadiqaalamdar:
New PR: https://github.com/OISF/suricata-verify/pull/1542

PR #1479, opened 2023-11-19
test: adds test for header_lowercase transform

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6290

Requires https://github.com/OISF/suricata/pull/9760 and next one

#1458 targetting only 8

PR #1533, opened 2023-12-12
detect/transform/header_lowercase: support from 7.0.3

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6613

https://github.com/OISF/suricata-verify/pull/1479 with lowering version support.

Do I get it right to put 7.0.3 which is not out yet ?

### DIFF
--- a/tests/transform-header-lowercase/test.yaml
+++ b/tests/transform-header-lowercase/test.yaml
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8.0.0
+  min-version: 7.0.3
 
 pcap: ../http2-range/http2-range.pcap
 


victorjulien:
I don't think we try to maintain compatibility with older versions in a branch, so you can just use 7 or 7.0 here

victorjulien:
Did the test get run with the current setting?

catenacyber:
> I don't think we try to maintain compatibility with older versions in a branch, so you can just use 7 or 7.0 here

So, should I do a new version of this PR with only 7 ?

> Did the test get run with the current setting?

The test runs and fails current main-7.0.x but passes with https://github.com/OISF/suricata/pull/10031
Is this not what is expected ?

I felt like this should be labelled `requires Suricata PR` instead of `requires backport` as the master Suricata PR was merged, and Suricata PR for main-7 is ready

victorjulien (approved):
I think it works correctly this way.

victorjulien:
Merged in #1537, thanks!

PR #1535, opened 2023-12-12
Adds test for negated content and absent buffer

catenacyber:
## Ticket

Requires next version of https://github.com/OISF/suricata/pull/9944

#1527 with new rule with both positive and negative content matching (that should not math on NULL buffers)


catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1912

PR #1538, opened 2023-12-12
sip: add tests for sip over tcp v6

glongo:
## Ticket

changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/3351


tests/sip-tcp-body-frames/test.rules
```diff
@@ -0,0 +1,11 @@
+alert sip any any -> any any (flow:to_server; frame:pdu; content:"REGISTER"; startswith; sid:2;)
+alert sip any any -> any any (flow:to_client; frame:pdu; content:"SIP/2.0 200 OK|0D 0A|"; startswith; sid:11;)
```
catenacyber, line 2: 🤔 maybe we should automate the run of S-V with simulated TCP fragmentation

glongo: TCP data (pkt 4 and 6) in sip-tcp-method/sip-tcp.pcap is already fragmented.

catenacyber: Why does not a PDU frame get created with only packet 4 ?

catenacyber: So, a frame gets created on packet 4 with size 400 (which is wrong)
But no tx gets created.
Then, the right frame gets created on packet 6...

victorjulien: So is this a bug?

glongo: A PDU frame is created only when a packet is parsed correctly, so the behavior described above does not happen anymore.

catenacyber: I think there was a bug in Suricata PR cf https://github.com/OISF/suricata/pull/10037#discussion_r1425882598

Discussion continues now in https://github.com/OISF/suricata/pull/10058#discussion_r1431038422

victorjulien (approved):

victorjulien:
Merged in #1672, thanks!

PR #1504, opened 2023-12-01
Add test about enip stats with _udp prefix

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6304

Suricata PR coming

PR #1546, opened 2023-12-13
Add test about enip stats with _udp prefix

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6304

Suricata PR https://github.com/OISF/suricata/pull/9953 required

#1504 with more explicit testing

### DIFF
--- a/tests/enip-stats-udp/test.yaml
+++ b/tests/enip-stats-udp/test.yaml
@@ -1,0 +1,28 @@
+requires:
+  min-version: 8
+
+pcap: ../enip-alert/enip_test1.pcap
+
+# disables checksum verification
+args:
+- -k none --set stream.midstream=true
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        app_proto: enip
+  - filter:
+      count: 1
+      match:
+        event_type: stats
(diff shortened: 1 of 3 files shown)


tests/enip-stats-udp/test.yaml
```diff
@@ -0,0 +1,28 @@
...
+args:
+- -k none --set stream.midstream=true
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        app_proto: enip
+  - filter:
```
victorjulien, line 16: I don't understand this. If we have a flow with enip, shouldn't we also have a non-0 stats entry?

catenacyber: Yes, but that is another bug.

There are many cases where the stats do not reflect the number of flows

`jq 'select(.event_type=="flow" and .app_proto=="ftp") | .app_proto'  log/eve.json | wc -l`
And
`jq 'select(.event_type=="stats") | .stats."app_layer".flow.ftp' log/eve.json`
do not give the same number

cf https://github.com/OISF/suricata/pull/8328

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1547

PR #1556, opened 2023-12-18
add test for bug 6617

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6617

### DIFF
--- a/tests/bug-6617/test.rules
+++ b/tests/bug-6617/test.rules
@@ -1,0 +1,2 @@
+alert http any any -> any any (msg:"Filestore toserver"; filestore:to_server,flow; sid:1; rev:1;)
+alert http any any -> any any (msg:"Filestore toclient"; filestore:to_client,flow; sid:2; rev:1;)
--- a/tests/bug-6617/test.yaml
+++ b/tests/bug-6617/test.yaml
@@ -1,0 +1,17 @@
+pcap: ../filestore-v2.1-forced/suricata-update-pdf.pcap
+
+args:
+- -k none
+
+checks:
+- filter:
+    count: 0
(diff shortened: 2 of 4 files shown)


tests/bug-6617/test.rules
```diff
@@ -0,0 +1,2 @@
+alert http any any -> any any (msg:"Filestore toserver"; filestore:to_server,flow; sid:1; rev:1;)
+alert http any any -> any any (msg:"Filestore toclient"; filestore:to_client,flow; sid:2; rev:1;)
```
catenacyber, line 2: The bug happens if we have only one of these, right ?

tests/bug-6617/test.yaml
```diff
@@ -0,0 +1,17 @@
...
+    count: 0
+    match:
+      alert.signature: Filestore toserver
+      alert.signature_id: 1
+
+- filter:
+    count: 1
+    match:
+      alert.signature: Filestore toclient
+      alert.signature_id: 2
```
catenacyber, line 17: The test should check the stored files (so file info events with the stored property)

catenacyber: More precisely, the bug is that files will be stored whatever their direction

inashivb: hmmm I think I understood that but misunderstood `filestore: to_server,flow` keyword. I thought it'll only alert if it is able to store a file in that dir.

catenacyber: If I understand correctly, `filestore` keyword is postmatch action

PR #1559, opened 2023-12-19
requires: check stat; unknown requires keyword - v2

jasonish:
For the requires keyword, check the new state, `rules_skipped`. Also add a test
for an unknown requires keyword to make sure the rule doesn't error out.

Feature: https://redmine.openinfosecfoundation.org/issues/6637


victorjulien (approved):

victorjulien:
Merged in #1570, thanks!

PR #1563, opened 2023-12-21
test: add test for dns.rcode

hadiqaalamdar:
Feature #6621

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6621

Suricata PR: https://github.com/OISF/suricata/pull/10087

Output from stdout file:
```
Notice: suricata: This is Suricata version 8.0.0-dev (5cc872fa1 2023-12-19) running in USER mode [LogVersion:suricata.c:1147]
Warning: detect: 1 rule files specified, but no rules were loaded! [SigLoadSignatures:detect-engine-loader.c:359]

```

### DIFF
--- a/tests/dns/dns-rcode/README.md
+++ b/tests/dns/dns-rcode/README.md
@@ -1,0 +1,5 @@
+Test the `dns.rcode` sticky buffer.
+
+The PCAP here was a request created with Scapy to include answers in
+the request. However the response is from a real DNS server with the
+provided request.
--- a/tests/dns/dns-rcode/test.rules
+++ b/tests/dns/dns-rcode/test.rules
@@ -1,0 +1,8 @@
+# Should alert in both directions as no flow is provided.
+alert dns any any -> any any (dns.rcode; content:"oisf"; sid:1; rev:1;)
+
+# Should only alert in the request direction.
+alert dns any any -> any any (dns.rcode; content:"oisf"; flow:to_server; sid:2; rev:1;)
+
+# Should only alert in the response direction.
+alert dns any any -> any any (dns.rcode; content:"oisf"; flow:to_client; sid:3; rev:1;)
--- a/tests/dns/dns-rcode/test.yaml
+++ b/tests/dns/dns-rcode/test.yaml
@@ -1,0 +1,43 @@
+requires:
+  min-version: 8
+
+checks:
+  - filter:
+      count: 2
+      match:
+        alert.signature_id: 1
+        app_proto: dns
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 1
+        direction: to_client
+        app_proto: dns
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 1
+        direction: to_server
+        app_proto: dns
+
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 2
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 2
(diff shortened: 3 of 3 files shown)


jasonish:
This test is missing a pcap. Most of our existing DNS pcaps will have rcode values of 0 so are not that interesting. But `./tests/dns-eve-v2-udp-nxdomain-soa/dns-udp-nxdomain-soa.pcap` does have a DNS response with a non-zero error code.

catenacyber:
> This test is missing a pcap. Most of our existing DNS pcaps will have rcode values of 0 so are not that interesting. But `./tests/dns-eve-v2-udp-nxdomain-soa/dns-udp-nxdomain-soa.pcap` does have a DNS response with a non-zero error code.

You should use `pcap:` in test.yaml to tell which pcap you want to reuse

tests/dns/dns-rcode/test.rules
```diff
@@ -0,0 +1,8 @@
+# Should alert in both directions as no flow is provided.
+alert dns any any -> any any (dns.rcode; content:"oisf"; sid:1; rev:1;)
```
catenacyber, line 2: should it not be something like `dns.rcode:12` ?

catenacyber: and no content

tests/dns/dns-rcode/test.yaml
```diff
@@ -0,0 +1,43 @@
...
+      count: 1
+      match:
+        alert.signature_id: 2
+        direction: to_server
+        app_proto: dns
+
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 3
```
catenacyber, line 37: You can have only one filter per signature id, right ?

tests/dns/dns-rcode/README.md
```diff
@@ -0,0 +1,5 @@
+Test the `dns.rcode` sticky buffer.
+
+The PCAP here was a request created with Scapy to include answers in
+the request. However the response is from a real DNS server with the
+provided request.
```
jufajardini, line 5: Please add a reference to the redmine ticket :)

If you do re-use a pcap from a different test, you can indicate that here, too.

tests/dns/dns-rcode/test.rules
jufajardini: Since it's an integer keyword ;)

tests/dns/dns-rcode/test.yaml
```diff
@@ -0,0 +1,43 @@
+requires:
+  min-version: 8
+
+checks:
+  - filter:
+      count: 2
+      match:
+        alert.signature_id: 1
+        app_proto: dns
```
jufajardini, line 9: it's important to add a check for the rcode field on the checks, too :)

jufajardini:
> This test is missing a pcap. Most of our existing DNS pcaps will have rcode values of 0 so are not that interesting. But `./tests/dns-eve-v2-udp-nxdomain-soa/dns-udp-nxdomain-soa.pcap` does have a DNS response with a non-zero error code.

On the other hand, maybe we could have two tests, and use one of the pcaps with`rcode` of zero to test the negated content option? 🤔

jufajardini (changes requested):
Added a few more comments to help you keep moving! :)

hadiqaalamdar:
New PR: https://github.com/OISF/suricata-verify/pull/1567

PR #1565, opened 2023-12-25
detect-engine-state: tests - Draft

0xEniola:
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6146


tests/engine-state/detect-engine-state-03/suricata.yaml
```diff
@@ -0,0 +1,14 @@
+%YAML 1.1
+---
+
+outputs:
+  - eve-log:
+      enabled: yes
+      types:
+        - files
+        - stats
```
inashivb, line 9: It would be useful to log at least `http`, `anomaly` and `flow` events alongside these. If the file events are not coming through, we'll get some idea.

0xEniola: You mean to add http, anomaly and flow to suricata.yaml or to createst?

PR #1566, opened 2023-12-26
detect-engine-state: tests - draft

0xEniola:
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6146


inashivb (changes requested):
Hi, Eniola! Two things from me for your next rev even in draft:
1. The pcap file seems to have a lot more than is needed for the test. Could you please restrict the capture to the http communication like you did previously so its easier?
2. There should be a test.yaml now in the test.

0xEniola:
> Hi, Eniola! Two things from me for your next rev even in draft:
> 1. The pcap file seems to have a lot more than is needed for the test. Could you please restrict the capture to the http communication like you did previously so its easier?
> 2. There should be a test.yaml now in the test.

So I should restrict it to only the HTTP communication.

OK then.

PR #1575, opened 2024-01-07
detect-engine-state: add tests - v2

0xEniola:
Link to redmine ticket: https://redmine.openinfosecfoundation.org/issues/6146

Conversion:

detect-engine-state-01: DeStateSigTest01 - DeStateSigTest02
detect-engine-state-02: DeStateSigTest03 - DeStateSigTest06
detect-engine-state-03: DeStateSigTest07
detect-engine-state-04: DeStateSigTest08
detect-engine-state-05: DeStateSigTest09 - DeStateSigTest10

Completed task from https://github.com/OISF/suricata-verify/pull/807.

Previous PR: #1555

- Incorporated feedbacks from last PR.
- Rebuilt the pcap files for all tests.
- Corrected the rules for some tests.
- Added Suricata.yaml to some tests.
- Added filestore tests to the test.yaml files of some of the tests.


tests/engine-state/detect-engine-state-03/input.t
```diff
@@ -0,0 +1,15 @@
+>>>
+POST /upload.cgi HTTP/1.1
+Host: www.server.lan
+Content-Type: multipart/form-data; boundary=---------------------------277531038314945
+Content-Length: 215
+
```
jufajardini, line 6: This looks the same as for detect-engine-state-02, is that correct?

If that's the case, then test 3 could just pass the pcap from test 2 as argument, instead of us having duplicated pcaps. (see https://github.com/OISF/suricata-verify/blob/master/tests/tls/tls-subject/test.yaml#L1 for an example on how to do that :) )

jufajardini (changes requested):
Kudos on this work, it's looking good!

Two things to adjust, from my end:
- I'm not sure we want to have the `htptopcap.py` script here - definitely not in each of the test folders, at least.  Maybe add it to the `engine-state` folder?
- Could you adjust the commit message a bit? Maybe... 
```
detect-engine-state: add assorted tests

Task: 6146
Add previously Suricata unittests ass Suricata-verify tests.
```

tests/engine-state/detect-engine-state-03/input.t
0xEniola: No! test 3 makes use of two buffers, test 2 makes use of one, that's why they are in different test folders.

jufajardini: Ok, right, I checked everything else but that aspect.

PR #1577, opened 2024-01-09
 Better support for nested directories - v2

jasonish:
Rebase of https://github.com/OISF/suricata-verify/pull/1487.

First, if doing `run.py pattern` consider the directories to the test as part of the test name. For example you might have a directory `foobar` with tests named `one`, `two` and so on.  You'd want `run.py foobar` to run all those tests.

Second, handle more then one level of sub-directories when displaying the test name.

### DIFF
--- a/run.py
+++ b/run.py
@@ -1,6 +1,6 @@
 #! /usr/bin/env python3
 #
-# Copyright (C) 2017-2022 Open Information Security Foundation
+# Copyright (C) 2017-2024 Open Information Security Foundation
 #
 # Permission is hereby granted, free of charge, to any person
 # obtaining a copy of this software and associated documentation files
@@ -1165,11 +1165,12 @@ def main():
         if not args.patterns:
             tests.append(dirpath)
         else:
+            test_name = dirpath[len(tdir) + 1:]
             for pattern in args.patterns:
                 if args.exact:
-                    if pattern == basename:
+                    if pattern == test_name:
                         tests.append(dirpath)
-                elif basename.find(pattern) > -1:
+                elif test_name.find(pattern) > -1:
                     tests.append(dirpath)
 
     # Sort alphabetically.
--- a/run.py
+++ b/run.py
@@ -793,10 +793,11 @@ class TestRunner:
 
             if not check_value["failure"] and not check_value["skipped"]:
                 if not self.quiet:
-                    if os.path.basename(os.path.dirname(self.directory)) != "tests":
-                        path_name = os.path.join(os.path.basename(os.path.dirname(self.directory)), self.name)
+                    test_name_offset = self.directory.find("tests/")
+                    if test_name_offset > -1:
+                        path_name = self.directory[test_name_offset + len("tests/"):]
                     else:
-                        path_name = (os.path.basename(self.directory))
+                        path_name = os.path.basename(self.directory)
                     print("===> %s: OK%s" % (path_name, " (%dx)" % count if count > 1 else ""))
             elif not check_value["failure"]:
                 if not self.quiet:


run.py
```diff
@@ -1165,11 +1166,12 @@ def main():
         if not args.patterns:
             tests.append(dirpath)
         else:
+            test_name = dirpath[len(tdir) + 1:]
             for pattern in args.patterns:
                 if args.exact:
-                    if pattern == basename:
+                    if pattern == test_name:
```
catenacyber, line 1172: So, now I need to say `dns/dns-rcode` instead of `dns-rcode` ?

run.py
```diff
@@ -793,10 +793,11 @@ def run(self, outdir):
 
             if not check_value["failure"] and not check_value["skipped"]:
                 if not self.quiet:
-                    if os.path.basename(os.path.dirname(self.directory)) != "tests":
-                        path_name = os.path.join(os.path.basename(os.path.dirname(self.directory)), self.name)
+                    test_name_offset = self.directory.find("tests/")
```
catenacyber, line 796: We cannot use `tests` in subdirectories then 🤔

run.py
```diff
@@ -793,10 +793,11 @@ def run(self, outdir):
...
+                    if test_name_offset > -1:
+                        path_name = self.directory[test_name_offset + len("tests/"):]
                     else:
-                        path_name = (os.path.basename(self.directory))
+                        path_name = os.path.basename(self.directory)
```
catenacyber, line 800: Does not this case need to be updated as well not to use basename ?

run.py
jasonish: > We cannot use `tests` in subdirectories then 🤔

Why not?

Moving `tests/dns/*` into `tests/dns/tests/...` seems to work fine:
```
[jason@desktop …/dev/suricata/master]  master ❯ ~/oisf/dev/suricata-verify/master/run.py tests/dns
    Finished release [optimized] target(s) in 0.01s
Number of concurrent jobs: 8
===> dns/tests/dns-eve-empty-format: OK
===> dns/tests/dns-frames: OK
===> dns/tests/dns-query-name: OK
===> dns/tests/dns-invalid-opcode: OK
===> dns/tests/dns-answer-name: OK
```

run.py
jasonish: Perhaps.  But this case is hit when you are pointing Suricata at some directory of private tests.  I didn't want to make any assumptions as to what that directory might look like, and the "top" directory pointed to by the command line is lost by the time we get here.

But perhaps it makes sense to just display the full path then?

```
[jason@desktop …/dev/suricata/master]  master ❯ ~/oisf/dev/suricata-verify/master/run.py --testdir ~/scratch
    Finished release [optimized] target(s) in 0.01s
Number of concurrent jobs: 8
===> /home/jason/scratch/dns-invalid-opcode: OK

PASSED:  1
FAILED:  0
SKIPPED: 0
```

run.py
jasonish: How so?
```
[jason@desktop …/dev/suricata/master]  master ❯ ~/oisf/dev/suricata-verify/master/run.py dns-frames
    Finished release [optimized] target(s) in 0.01s
Number of concurrent jobs: 8
===> dns/dns-frames: OK

PASSED:  1
FAILED:  0
SKIPPED: 0
```

run.py
victorjulien: full path sounds good to me

run.py
catenacyber: You did not use `args.exact` here

catenacyber: Also This line does not fit with the commit message :-p

run.py
jasonish: Will use full-path.

run.py
catenacyber: > Does anyone actually use this feature?

I do not.

I guess this line change deserves its own commit, so that the chosen behavior is documented...

jufajardini:
This is most welcome! I think it would deserve a readme update.

PR #1597, opened 2024-01-21

coledishington:

Bug: #6260

Redmine ticket:


catenacyber:
Looks good to me with the suricata PR

jufajardini:
I see that IPS mode is mentioned in the README, but didn't see a test that simulated ips mode. Shouldn't there be one, or isn't that relevant for the changes?

coledishington:
> I see that IPS mode is mentioned in the README, but didn't see a test that simulated ips mode. Shouldn't there be one, or isn't that relevant for the changes?

IPS simulation shouldn't make a difference as this feature is related to stitching traffic together when one side is detected before tunnel encapsulation and the other side is detected before tunnel decapsulation. e.g. For an ipv6 tunnel
request:  IPv4]ICMP] -> |IPS| -> IPv6]IPv4]ICMP]
reply:               <- |IPS| <- IPv6]IPv4]ICMP]

This is relevant when the IDS/IPS is also the tunnel terminator.

@jufajardini As both IDS and IPS mode would work in this scenario, I am happy to enable --simulate-ips if it makes more sense.

jufajardini:
> > I see that IPS mode is mentioned in the README, but didn't see a test that simulated ips mode. Shouldn't there be one, or isn't that relevant for the changes?
> 
> IPS simulation shouldn't make a difference as this feature is related to stitching traffic together when one side is detected before tunnel encapsulation and the other side is detected before tunnel decapsulation. e.g. For an ipv6 tunnel request: IPv4]ICMP] -> |IPS| -> IPv6]IPv4]ICMP] reply: <- |IPS| <- IPv6]IPv4]ICMP]
> 
> This is relevant when the IDS/IPS is also the tunnel terminator.
> 
> @jufajardini As both IDS and IPS mode would work in this scenario, I am happy to enable --simulate-ips if it makes more sense.

[Sorry for the very late answer] considering that it should work for both, I think that ideally we would have tests for both scenarios :P

coledishington:
> > > I see that IPS mode is mentioned in the README, but didn't see a test that simulated ips mode. Shouldn't there be one, or isn't that relevant for the changes?
> > 
> > 
> > IPS simulation shouldn't make a difference as this feature is related to stitching traffic together when one side is detected before tunnel encapsulation and the other side is detected before tunnel decapsulation. e.g. For an ipv6 tunnel request: IPv4]ICMP] -> |IPS| -> IPv6]IPv4]ICMP] reply: <- |IPS| <- IPv6]IPv4]ICMP]
> > This is relevant when the IDS/IPS is also the tunnel terminator.
> > @jufajardini As both IDS and IPS mode would work in this scenario, I am happy to enable --simulate-ips if it makes more sense.
> 
> [Sorry for the very late answer] considering that it should work for both, I think that ideally we would have tests for both scenarios :P

Addressed your comments in https://github.com/OISF/suricata-verify/pull/2081.

PR #1598, opened 2024-01-22
requires: update skipped rules counter

inashivb:
@jasonish this was fairly confusing to me.
1. If the `requires` keyword has bad language, why is it not an error?
2. If `requires` keyword has bad language, we skip the rule with a warning but the rule would still be parsed generating another error if the rule lang is bad elsewhere. Why do we allow this? [Case in point: `sid: 102` rule for `requires` test]

### DIFF
--- a/tests/requires/test.yaml
+++ b/tests/requires/test.yaml
@@ -47,4 +47,4 @@ checks:
       count: 1
       match:
         event_type: stats
-        stats.detect.engines[0].rules_skipped: 6
+        stats.detect.engines[0].rules_skipped: 7


jasonish:
> 1. If the `requires` keyword has bad language, why is it not an error?
> 
>     2. If `requires` keyword has bad language, we skip the rule with a warning but the rule would still be parsed generating another error if the rule lang is bad elsewhere. Why do we allow this? [Case in point: `sid: 102` rule for `requires` test]

The first pass of the rule parser just check for good form, and sid 102 has good form.  The protocol is not known, nor is the `foo bar` in requires, but the format is correct. This allows us to extend the requires later.  For this rule, the `foo` requirement field is not known, so it is considered not to full-fill the requirements, so will not be loaded.

inashivb:
> The first pass of the rule parser just check for good form, and sid 102 has good form. The protocol is not known, nor is the `foo bar` in requires, but the format is correct. This allows us to extend the requires later. For this rule, the `foo` requirement field is not known, so it is considered not to full-fill the requirements, so will not be loaded.

Ok about everything related to `requires`.
Questions:
1. if the protocol is not known and it **is** an error, why do we skip that rule based on its good form?
Reason I'm asking: If we display an "error parsing signature ...", I'd expect some kind of error count to be updated (or a message like that suppressed), why should the "skipped" count be updated? This is still a bit unclear to me..
2. Do the changes in this PR look alright to you?

jasonish:
> > The first pass of the rule parser just check for good form, and sid 102 has good form. The protocol is not known, nor is the `foo bar` in requires, but the format is correct. This allows us to extend the requires later. For this rule, the `foo` requirement field is not known, so it is considered not to full-fill the requirements, so will not be loaded.
> 
> Ok about everything related to `requires`. Questions:
> 
>     1. if the protocol is not known and it **is** an error, why do we skip that rule based on its good form?
>        Reason I'm asking: If we display an "error parsing signature ...", I'd expect some kind of error count to be updated (or a message like that suppressed), why should the "skipped" count be updated? This is still a bit unclear to me..
>        ~2. Do the changes in this PR look alright to you?~ nvm hadn't seen your latest PR :)

Given the rule:
```
alert bad-proto any any -> any any (requires: version >= 9; sid:999; rev:0;)
```
Here we skip the rule if Suricata is less than version 9.  But if Suricata was at least 9 and `bad-proto` was not known, you'd get the error as usual as we will fully evaluate this rule.

Does that answer your question?

inashivb:
Sorry for dragging this and not explaining properly. 😅  Let me try to explain my question again.
> 
> Given the rule:
> 
> ```
> alert bad-proto any any -> any any (requires: version >= 9; sid:999; rev:0;)
> ```
> 
> Here we skip the rule if Suricata is less than version 9. But if Suricata was at least 9 and `bad-proto` was not known, you'd get the error as usual as we will fully evaluate this rule.
> 

I tried this rule, it leads to just the skipping which makes sense to me as the version requirements don't meet so we don't evaluate this further. Output:
```
[283029 - Suricata-Main] 2024-01-23 10:11:37 Info: detect-requires: Suricata did not meet the rule requirements: Suricata version less than required: version >= 9
[283029 - Suricata-Main] 2024-01-23 10:11:37 Info: detect: Skipping signature due to missing requirements: alert bad-proto any any -> any any (requires: version >= 9; sid:999; rev:0;) from file /home/ard/Orgs/oisf/public/suricata-verify/tests/DELETE-Jason-requires/test.rules at line 1
[283029 - Suricata-Main] 2024-01-23 10:11:37 Warning: detect: 1 rule files specified, but no rules were loaded!
```

But, for a rule like
```
alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)
```
we get the following output:
```
[259972 - Suricata-Main] 2024-01-22 15:12:21 Warning: requires: Unknown requires keyword: foo
[259972 - Suricata-Main] 2024-01-22 15:12:21 Error: detect-parse: protocol "asdf" cannot be used in a signature.  Either detection for this protocol is not yet supported OR detection has been disabled for protocol through the yaml option app-layer.protocols.asdf.detection-enabled
[259972 - Suricata-Main] 2024-01-22 15:12:21 Error: detect: error parsing signature "alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)" from file /home/ard/Orgs/oisf/public/suricata-verify/tests/requires/test.rules at line 17
[259972 - Suricata-Main] 2024-01-22 15:12:21 Info: detect: Skipping signature due to missing requirements: alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;) from file /home/ard/Orgs/oisf/public/suricata-verify/tests/requires/test.rules at line 17
```
There's a warning **and** an error for the same rule.
Since there is an error on the rule, I'd expect `rules_failed` counter to be 1. But, instead, what's stated in `Info` message takes precedence and we increment `rules_skipped`.
Why do we choose to do this?

jasonish:
> `alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)`

With just this rule, and Suricata git master I do get the rules failed counter incremented and not rules_skipped:
```json
    "detect": {
      "engines": [
        {
          "id": 0,
          "last_reload": "2024-01-23T10:40:18.127892-0600",
          "rules_loaded": 0,
          "rules_failed": 1,
          "rules_skipped": 0
        }
      ],
      "alert": 0,
      "alert_queue_overflow": 0,
      "alerts_suppressed": 0
    },
```

I was wrong about unknown requires keywords being treated as unful-filled features.  They just emit a warning now, but don't alter the flow of execution.  So you'd get all the rule loading errors you'd expect if it wasn't there.

inashivb:
> > `alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)`
> 
> With just this rule, and Suricata git master I do get the rules failed counter incremented and not rules_skipped:
> 
> ```json
>     "detect": {
>       "engines": [
>         {
>           "id": 0,
>           "last_reload": "2024-01-23T10:40:18.127892-0600",
>           "rules_loaded": 0,
>           "rules_failed": 1,
>           "rules_skipped": 0
>         }
>       ],
>       "alert": 0,
>       "alert_queue_overflow": 0,
>       "alerts_suppressed": 0
>     },
> ```
> 

ah. You're absolutely right. I never tried loading **just** that rule. Now, its even more interesting.
Because, the above rule standalone fails and increments `rules_failed` but the same rule as a part of the other `requires` keyword rules increments `rules_skipped` exactly as demonstrated in the `requires` test.
Just to be clear,
```
# Rule for Suricata >= 7 and < 8.
alert http any any -> any any (msg:"TEST Suricata >= 7 and < 8"; content:"uid=0"; requires: version >= 7 < 8; sid:7; rev:1;)

# Rule for Suricata >= 7.0.3 but less than 8... Or >= 8.0.1
alert http any any -> any any (content:"uid=0"; requires: version >= 7.0.3 < 8 | >= 8.0.1; sid:9; rev:1;)

# Rule for Suricata >= 8.
alert http any any -> any any (msg:"TEST Suricata >= 8"; content:"uid=0"; requires: version >= 8.0.0; sid:8; rev:1;)

# These rules have something invalid about them, but do follow the general rule
# structure, so should be eliminated by the requires statement.
alert vxlan any any -> any any (requires: version >= 10; sid:1;)
alert udp any any -> any any (vxlan_vni:10; requires: version >= 10; sid:2;)
alert http any any => any any (requires: version >= 10; sid:3;)
alert tcp any any -> any any (frame:smtp.not_supported; requires: version >= 10; sid:4;)

alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)
```
gives the following in Suricata output:
```
[299008 - Suricata-Main] 2024-01-24 04:47:36 Warning: requires: Unknown requires keyword: foo
[299008 - Suricata-Main] 2024-01-24 04:47:36 Error: detect-parse: protocol "asdf" cannot be used in a signature.  Either detection for this protocol is not yet supported OR detection has been disabled for protocol through the yaml option app-layer.protocols.asdf.detection-enabled
[299008 - Suricata-Main] 2024-01-24 04:47:36 Error: detect: error parsing signature "alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)" from file /home/ard/Orgs/oisf/public/suricata-verify/tests/DELETE-only-requires-2/test.rules at line 17
[299008 - Suricata-Main] 2024-01-24 04:47:36 Info: detect: Skipping signature due to missing requirements: alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;) from file /home/ard/Orgs/oisf/public/suricata-verify/tests/DELETE-only-requires-2/test.rules at line 17
```
and, the following in eve.json (as was recently updated in the test):
```json
{
    "id": 0,
    "last_reload": "2024-01-24T04:47:36.604925+0000",
    "rules_loaded": 1,
    "rules_failed": 0,
    "rules_skipped": 7
  }
```

This looks inconsistent behavior to me. What do you think?

jasonish:
> This looks inconsistent behavior to me. What do you think?

Yeah, this is odd how it works correctly in isolation, but not with the other rules. Taking a look.

jasonish:
> > This looks inconsistent behavior to me. What do you think?
> 
> Yeah, this is odd how it works correctly in isolation, but not with the other rules. Taking a look.

Thanks for the sharp eye: https://redmine.openinfosecfoundation.org/issues/6710

inashivb:
> Thanks for the sharp eye: https://redmine.openinfosecfoundation.org/issues/6710

Awesome! Thanks for fixing! :)

PR #1603, opened 2024-01-23
Adds test about invalid bidirectional rules

catenacyber:
## Ticket

#1591 with more tests for matching on first side and not on second side


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/1668

PR #1606, opened 2024-01-24

coledishington:

Feature: #6293

Redmine ticket:


coledishington:
Memcap won't always be hit when flow-reuse is on. Changed in https://github.com/OISF/suricata-verify/pull/1607

PR #1607, opened 2024-01-24

coledishington:

Feature: #6293

Redmine ticket:


coledishington:
Suricata PR is https://github.com/OISF/suricata/pull/10232

tests/flow-force-reuse/flow-force-reuse-off/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+args:
+- --set flow.memcap-policy=drop-flow
+- --set flow.prealloc=0
+- --set flow.memcap=3kb
+- --set flow.hash-size=32
+- --set flow.force-reuse=false
+
+checks:
+  - shell:
+      args: grep 'flow.memcap ' stats.log | wc -l
```
victorjulien, line 12: prefer eve.json stats checks here

tests/flow-force-reuse/flow-force-reuse-on/test.yaml
```diff
@@ -0,0 +1,13 @@
...
+args:
+- --set flow.memcap-policy=drop-flow
+- --set flow.prealloc=0
+- --set flow.memcap=3kb
+- --set flow.hash-size=32
+- --set flow.force-reuse=true
+
+checks:
+  - shell:
+      args: grep -c 'flow[.]get_used ' stats.log
```
victorjulien, line 12: eve.json

victorjulien (changes requested):
see inline

tests/flow-force-reuse/flow-force-reuse-off/test.yaml
coledishington: The number of counters on flow.memcap and flow.get_used isn't always the same, so the grep is just checking if there is any. Is this comment to use the match syntax of test.yaml or continue using grep but grep eve.json instead?

Thanks

coledishington:
Updated for PR in https://github.com/OISF/suricata-verify/pull/1614

PR #1610, opened 2024-01-24
tests/requires: test failure case

jasonish:
Break the requires test into 2 tests.  One that runs to success so we can verify the output.

A second that is expected to fail due to fatal init errors.

Ticket: https://redmine.openinfosecfoundation.org/issues/6710


victorjulien (approved):

victorjulien:
Merged in #1612, thanks!

PR #1012, opened 2022-11-21
tests: add tests for negated protocol matching

catenacyber:
https://redmine.openinfosecfoundation.org/issues/4921

#615 still failing test

PR #1613, opened 2024-01-25
tests: add tests for negated protocol matching

catenacyber:
## Ticket

https://redmine.openinfosecfoundation.org/issues/4921

#1012 with rewritten rule to match suricata incoming PR


jufajardini (commented):
Considering the README description and the docs for this, could we have a rule with something like `app-layer-protocol:failed; app-layer-protocol:toserver` ?

catenacyber:
> Considering the README description and the docs for this, could we have a rule with something like `app-layer-protocol:failed; app-layer-protocol:toserver` ?

Do you mean rather `app-layer-protocol:failed,toserver;` ?

catenacyber:
Changed the pcap to not use `SUBSCRIBE` SIP method

jufajardini:
> > Considering the README description and the docs for this, could we have a rule with something like `app-layer-protocol:failed; app-layer-protocol:toserver` ?
> 
> Do you mean rather `app-layer-protocol:failed,toserver;` ?

Yes, that, sorry.

catenacyber:
But this rule would not match :

Here, we have a pcap with one flow, where do not recognize protocol to server, but we recognize http toclient, and we thus classify the whole as http...

Maybe you are on to something about protocol detection, rather than the keyword...
So, what do you expect ?

jufajardini:
> But this rule would not match :
> 
> Here, we have a pcap with one flow, where do not recognize protocol to server, but we recognize http toclient, and we thus classify the whole as http...
> 
> Maybe you are on to something about protocol detection, rather than the keyword... So, what do you expect ?

It was more about trying to understand how the keyword works and what is allowed or not with it...

catenacyber:
Do you want unit tests for keyword parser ?

jufajardini:
> Do you want unit tests for keyword parser ?

You're right, maybe this isn't the right test for those. XD

catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1823

PR #1178, opened 2023-04-17
tests: add exception/policy stats counters - v2

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/1160

Add exception policy counters to the existing exception-policy tests

Includes commit that should hopefully be merged before this one (the simulated flow memcap tests).

Changes from previous PR:
- removed test for flowmemcap, since I wasn't able to craft one that works.

PR #1617, opened 2024-01-26
tests: add exception/policy stats counters - v3

jufajardini:
Edit the existing exception policy tests to check for the new exception policy stats counters.

Add one new test.

Ticket #5816

Previous PR: https://github.com/OISF/suricata-verify/pull/1178

Sharing as a draft because this will trigger a buffer overflow caused by test `exception-policy-applayer-01` when enabling threads and deltas for stats. Want to have it like that, so I'm reminded that this has to be fixed, still.

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5816

### DIFF
--- a/tests/exception-policy-applayer-02/README.md
+++ b/tests/exception-policy-applayer-02/README.md
@@ -1,0 +1,15 @@
+# Test
+
+Showcase exception policy stats counters for application layer protocol errors,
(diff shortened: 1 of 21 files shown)


jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1671

PR #1620, opened 2024-01-29
tests/imap-detection: add simple test for protocol detection.

mmaatuq:
## Ticket #2886

Redmine ticket: [imap protocol detection is incomplete](https://redmine.openinfosecfoundation.org/issues/2886)


inashivb (changes requested):
Hi @mmaatuq !
Thank you for your first contribution to the project. :)
The test looks good me. Minor changes requested from me:
- Please make the commit message short. Something like: `imap: add test for protocol detection`
- Get rid of `.` at the end of the commit message.
- Make sure to use the same email in both the projects to sign the commits. This should be the one that you used to sign the contribution agreement. (I think they're in sync but your Suricata commit's signed-off-by line is diff. We don't need the signed-off-by line but don't have any reservations against it.)

mmaatuq:
> Hi @mmaatuq ! Thank you for your first contribution to the project. :) The test looks good me. Minor changes requested from me:
> 
>     * Please make the commit message short. Something like: `imap: add test for protocol detection`
> 
>     * Get rid of `.` at the end of the commit message.
> 
>     * Make sure to use the same email in both the projects to sign the commits. This should be the one that you used to sign the contribution agreement. (I think they're in sync but your Suricata commit's signed-off-by line is diff. We don't need the signed-off-by line but don't have any reservations against it.)

Hi @inashivb, 
thanks for your notes, they are incorporated now.

tests/imap-detection/test.yaml
```diff
@@ -0,0 +1,36 @@
+requires:
+  files:
+    - src/app-layer-imap.c
```
catenacyber, line 3: I do not think we should this requirement.

So, CI will be red until suricata PR gets merged...

PR #1435, opened 2023-10-22
tests: add rule to check for tcp_seq - v1

0xEniola:
Test for rule type for tcp-seq keyword

Related to
Issue: #6353

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6353

Previous Suricata PR: https://github.com/OISF/suricata/pull/9676

Suricata PR: https://github.com/OISF/suricata/pull/9677

PR #1621, opened 2024-01-30
tests: add rule to check for tcp_seq - v2

0xEniola:
Test for rule type for tcp-seq keyword

Related to
Issue: 6353

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6353

Suricata PR: https://github.com/OISF/suricata/pull/10285

Previous PR: https://github.com/OISF/suricata-verify/pull/1435

### DIFF
--- a/tests/rules/tcp-seq-keyword/test.yaml
+++ b/tests/rules/tcp-seq-keyword/test.yaml
@@ -1,0 +1,21 @@
+requires:
+    min-version: 7.0.0
+    pcap: false
+
+args:
+    - --engine-analysis
+
+checks:
+- filter:
+    filename: rules.json
+    count: 1
+    match:
+      id: 1
+      lists.packet.matches[0].name: "tcp.seq"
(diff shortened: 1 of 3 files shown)


tests/rules/tcp-seq-keyword/test.yaml
```diff
@@ -0,0 +1,21 @@
+requires:
+    min-version: 7.0.0
```
jufajardini, line 2: Just confirmed: none of these tasks to add details to `engine-analysis` output will be available for 7, so all PR for SV tests related to similar changes should have `min-version: 8.0`

jufajardini (changes requested):
Thanks for adding the README file, we still have minor updates needed (for this test and the similar PR you submitted :) )

tests/rules/tcp-seq-keyword/test.yaml
0xEniola: OK! Understood.

Thank you.

PR #1618, opened 2024-01-27
tests: add rule to check for tcp/ack - v4

0xEniola:
Test for rule type for tcp-ack keyword.

Related to
Issue: 6354

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6354

Suricata PR: https://github.com/OISF/suricata/pull/

Previous PR: https://github.com/OISF/suricata-verify/pull/1432

PR #1622, opened 2024-01-30
tests: add rule to check for tcp/ack - v5

0xEniola:
Test for rule type for tcp-ack keyword.

Related to
Issue: 6354

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6354

Suricata PR: https://github.com/OISF/suricata/pull/10266

Previous PR: https://github.com/OISF/suricata-verify/pull/1618

### DIFF
--- a/tests/rules/tcp_ack/README.md
+++ b/tests/rules/tcp_ack/README.md
@@ -1,0 +1,2 @@
+## Description
+Test for tcp-ack keyword rule; includes test.yaml and test.rules file.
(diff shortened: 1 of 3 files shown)


jufajardini (changes requested):
Thanks for this updated PR!
As I've commented on another review, this change won't make it to 7, so please update `min-version` requirement. The test description in the README file should be updated, too, to indicate that this test is for the `engine-analysis` output details.

0xEniola:
> Thanks for this updated PR! As I've commented on another review, this change won't make it to 7, so please update `min-version` requirement. The test description in the README file should be updated, too, to indicate that this test is for the `engine-analysis` output details.

OK, to change the`min-version` to 8, right?

PR #1630, opened 2024-02-06
tcp: add a SYN packet test to verify correct flow output v1

lukashino:
Adds a test for tcp syn flow output.

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6733


lukashino:
Follow up in #1659

PR #1631, opened 2024-02-06
ssh: do not enforce pcap_cnt

catenacyber:
## Ticket

Redmine ticket: None

Prerequisite for https://github.com/OISF/suricata/pull/10307 and next

### DIFF
--- a/tests/bug-4903/bug-4903-04/test.yaml
+++ b/tests/bug-4903/bug-4903-04/test.yaml
@@ -89,7 +89,6 @@ checks:
       flow.bytes_toserver: 336
       flow.pkts_toclient: 3
       flow.pkts_toserver: 4
-      pcap_cnt: 7
       proto: TCP
       src_ip: 192.168.100.1
       src_port: 10003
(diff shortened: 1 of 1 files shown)


jufajardini (approved):
Looks good to me, thanks for adding an explanation in the commit message!

jasonish:
Should we then be checking that `pcap_cnt` doesn't exist on the event? Not sure if S-V handles this case, but its probably worth doing than just removing the check.

catenacyber:
I think the important test is that we get the alert.
Getting it on packet 7, or pseudo packet timing out the flow is less important.

If we really want to test this, are we sure we want to consider packet 7 as invalid and not run tx detection on it ?

jasonish:
> If we really want to test this, are we sure we want to consider packet 7 as invalid and not run tx detection on it ?

I don't know specifically.  But S-V is about detecting changes in behaviour as well.  If it used to be 7, and some fix made the field go away, I think that should be encoded in the test, to detect if it ever comes back unintentionally.

catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1640

catenacyber:
> > If we really want to test this, are we sure we want to consider packet 7 as invalid and not run tx detection on it ?
> 
> I don't know specifically. But S-V is about detecting changes in behaviour as well. If it used to be 7, and some fix made the field go away, I think that should be encoded in the test, to detect if it ever comes back unintentionally.

Did something in next version of the PR

PR #1564, opened 2023-12-23
add JA4 test cases -- v5

satta:
Previous PR: #1426

This PR introduces test cases for TLS and QUIC JA4 calculation.
Requires Suricata PR https://github.com/OISF/suricata/pull/10095

Changes:
* Rebase against current master to match up expectations with current Suricata master behaviour.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6379

PR #1634, opened 2024-02-08
add JA4 test cases -- v6

satta:
Previous PR: #1564 

This PR introduces test cases for TLS and QUIC JA4 calculation.
Requires Suricata PR https://github.com/OISF/suricata/pull/10095

Changes to previous PR:
* Rebase against current master to match up expectations with current Suricata master behaviour.
* Add "requires: feature" tests for regular cases as well as for cases where JA3/JA4 are not compiled in or disabled.

## Ticket


tests/ja4-quic/test.yaml
```diff
@@ -0,0 +1,16 @@
+requires:
+  min-version: 7.0.0
```
catenacyber, line 2: Should be 8 ?

tests/ja4-rules-invalid/test.rules
```diff
@@ -0,0 +1 @@
+alert smb any any -> any any (msg:"JA4 Broken Test 1"; ja4.hash; content: "q"; sid:1;)
```
catenacyber, line 1: It is broken because of `smb` right ?

tests/ja4-rules-requires-off/test.yaml
```diff
@@ -0,0 +1,19 @@
+requires:
+  min-version: 7.0.3
+  files:
+    - rust/src/ja4.rs
+  script:
+    - grep 'ProvidesFeature(FEATURE_JA4);' src/app-layer-ssl.c > /dev/null
```
catenacyber, line 6: I do not think this grep is needed, if you have the next one

tests/ja4-rules-requires-off/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+    - ./src/suricata --build-info | grep "JA4 support" | grep no > /dev/null
```
catenacyber, line 7: Why not use `--dump-features` ?

tests/ja4-rules-requires/test.yaml
```diff
@@ -0,0 +1,18 @@
+requires:
+  min-version: 7.0.3
+  files:
+    - rust/src/ja4.rs
+  script:
+    - grep 'ProvidesFeature(FEATURE_JA4);' src/app-layer-ssl.c > /dev/null
```
catenacyber, line 6: Why not use --dump-features here ? What is the difference with the previous test ?

tests/tls/tls-ja3s/test.yaml
```diff
@@ -6,6 +6,7 @@ requires:
     - HAVE_LIBJANSSON
     - HAVE_NSS
     - HAVE_LUA
+    - HAVE_JA3
```
catenacyber, line 9: Could be its own prerequisite commit

tests/ja4-rules-invalid/test.rules
satta: Right, wrong protocol.

tests/ja4-quic/test.yaml
satta: True.

tests/ja4-rules-requires-off/test.yaml
satta: Yeah, right. You wouldn't even have "JA4 support" in the  `--build-info` output if there was no JA4 support in the code, even if is not enabled.

tests/tls/tls-ja3s-requires-off/test.yaml
```diff
@@ -0,0 +1,23 @@
+pcap: ../../tls/tls-certs-alert/input.pcap
+
+requires:
+  min-version: 7.0.3
+  features:
+    - HAVE_LIBJANSSON
+    - HAVE_NSS
```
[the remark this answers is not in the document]
satta: Totally. Will remove :)

tests/ja4-rules-requires/test.yaml
satta: In this case we want to test if the `requires` functionality works when JA4 support has been enabled at compile time.

tests/tls/tls-ja3s/test.yaml
satta: I think I'll separate all JA3 related changes out into a separate commit; might be cleaner.

satta:
Next: #1682

tests/ja4-rules-requires/test.yaml
catenacyber: This looks fragile, like if I rename `ProvidesFeature` to `ProvidingFeature`, this test will get skipped...

tests/ja4-rules-requires-off/test.yaml
catenacyber: ok, thanks, deserves a comment ;-)

PR #1641, opened 2024-02-12
http: adds test with chunked as a token

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6415


tests/http-chunked/test.yaml
```diff
@@ -0,0 +1,119 @@
+args:
+  - -k none --set packet-alert-max=32
```
catenacyber, line 2: This is always hard to debug

victorjulien:
Why does it fail on 6.0.x?

catenacyber:
> Why does it fail on 6.0.x?

Because `http.request_header` keyword was introduced in 7, doing another version

catenacyber:
Replaced by #1642

PR #1645, opened 2024-02-12
check-eve: open files with utf-8 encoding

jasonish:
Not needed in modern versions of Python, but required for older
versions like 3.6.


victorjulien (approved):

victorjulien:
Merged in #1646, thanks!

PR #1653, opened 2024-02-19
add rule grouping tests

inashivb:


victorjulien:
Think it would be nice to have a test with a "joingr" at the end as well.

inashivb:
> Think it would be nice to have a test with a "joingr" at the end as well.

Done in https://github.com/OISF/suricata-verify/pull/1657 thank yo!

PR #1658, opened 2024-02-20
add rule grouping tests

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/1657

Changes since v3:
- add check to ensure only the expected groups are created and no more


tests/rule-grouping/rule-grouping-1/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Test to demonstrate the port grouping and SGH distribution for small range
+overlaps and single points with "any".
+
+## PCAP
+
+None
+
+## Related issues
+
```
jufajardini, line 11: Did you leave ticket link out because this is still being discussed?

inashivb: No, because we don't have one but I think we should and that is still to be discussed. But, nonetheless, this PR should have been draft as it is incomplete. Thanks. :)

PR #1673, opened 2024-02-28
test-dump-counters: Test dump-counters via unix-command socket

awelzel:
## Ticket

Test for: https://github.com/OISF/suricata/pull/10468

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6732

PR #1676, opened 2024-03-01
test-dump-counters: Test dump-counters via unix-command socket - v3

awelzel:
Replaces https://github.com/OISF/suricata-verify/pull/1673

Changes since v2:

* Rewrote as shell script invoking suricatasc instead of using the suricatasc Python module
* Set rule-files.0=/dev/null to prevent loading of rules
* Set a BPF filter to make capture on `lo` unlikely

Test for https://github.com/OISF/suricata/pull/10468

## Ticket

Ticket: 6732

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6732

### DIFF
--- a/tests/test-dump-counters/test.sh
+++ b/tests/test-dump-counters/test.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+#
+# This test runs Suricata as background process listening on the lo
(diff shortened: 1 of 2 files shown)


awelzel:
Replaced by https://github.com/OISF/suricata-verify/pull/1683

PR #1682, opened 2024-03-04
JA3 and JA4 test adjustments and additions  - v7

satta:
Previous PR: #1634 

This PR introduces test cases for TLS and QUIC JA4 calculation, as well as some adjustments for JA3.

Changes to previous PR:
* Rebase against current master.
* Remove JA4s related data from reference output.
* Use version 8 in `requires.min-version`.
* Remove HAVE_NSS and HAVE_LIBJANSSON where unnecessary.
* Separate JA3 and JA4 related changes in to single commits.
* Clean up requires a bit.

## Ticket


tests/tls/tls-ja3s-requires/test.yaml
```diff
@@ -0,0 +1,22 @@
+pcap: ../../tls/tls-certs-alert/input.pcap
+
+requires:
+  min-version: 7.0.3
+  features:
+    - HAVE_LUA
+    - HAVE_JA3
+  script:
+    - grep 'ProvidesFeature(FEATURE_JA3);' src/app-layer-ssl.c > /dev/null
```
catenacyber, line 9: Not sure about this precondition...

Why is it needed ?
The SV PR will only be merged after the Suricata one anyways, right ?

tests/ja4-quic/suricata.yaml
```diff
@@ -0,0 +1,29 @@
...
+            extended: yes
+        - quic
+
+app-layer:
+  protocols:
+    tls:
+      enabled: yes
+      detection-ports:
+        dp: 443
+      ja4-fingerprints: yes
```
catenacyber, line 26: Maybe we can just set this line instead of having a full suricata.yaml ? (with args in test.yaml)

tests/ja4-quic/test.yaml
```diff
@@ -0,0 +1,16 @@
+requires:
+  min-version: 8.0.0
+  files:
+    - rust/src/ja4.rs
```
catenacyber, line 4: Do not need to check the file, if you require the feature, right ?
Thinking about if the file gets moved to rust/src/crypto/ja4.rs for instance

tests/ja4-rules-disabled/test.yaml
```diff
@@ -0,0 +1,9 @@
+requires:
+  min-version: 8.0.0
+  files:
+    - rust/src/ja4.rs
+
+args:
+  - -k none
+
+exit-code: 1
```
catenacyber, line 9: A comment explaining the purpose of the test is welcome (or a README.md)

tests/ja4-rules/test.rules
```diff
@@ -0,0 +1,4 @@
+alert quic any any -> any any (msg:"JA4 QUIC Test 1"; ja4.hash; content: "q13d0310h3_55b375c5d22e_cd85d2d88918"; sid:1;)
+alert tls any any -> any any (msg:"JA4 TLS Test 1"; ja4.hash; content: "t13d1516h2_8daaf6152771_e5627efa2ab1"; sid:2;)
+alert quic any any -> any any (msg:"JA4 QUIC Test 2"; ja4.hash; content: "q13d0310h3_55b375c5d22e_cd85d2d88918X"; sid:3;)
+alert tls any any -> any any (msg:"JA4 TLS Test 2"; ja4.hash; content: "t13d1516h2_8daaf6152771_e5627efa2ab1X"; sid:4;)
```
catenacyber, line 4: Could we use ` requires: feature ja4` ? So that we have a positive example with `requires`

tests/tls/tls-ja3s-requires/test.yaml
satta: Yes, but I also wanted the tests to also succeed in the PR here, where they are not necessarily tested against the branch of Suricata that this branch is targeted at. Some tests might not work correctly with the Suricata version used in the S-V CI.

tests/ja4-quic/suricata.yaml
satta: True. Changed.

tests/ja4-rules/test.rules
satta: Sure!

tests/ja4-quic/test.yaml
satta: True, changed.

tests/ja4-rules-disabled/test.yaml
satta: Done.

satta:
Next PR: #1761

PR #1684, opened 2024-03-05

coledishington:

This patch is using grep on stats.log, rather than usual eve.json, to test for existence of flow.get_used or flow.get_used_eval rather than a specific number.

Feature: #6293

Redmine ticket:


coledishington:
Flow memcap needed to be increased due to Mac's flow bucket size, changed in https://github.com/OISF/suricata-verify/pull/1688

PR #1664, opened 2024-02-22
test: add test for dns.rcode v6

hadiqaalamdar:
Feature #6621

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6621

Previous PR: https://github.com/OISF/suricata-verify/pull/1654

Suricata PR: https://github.com/OISF/suricata/pull/10458

Changes made:
- improved the tests for rcode which are all passing.

Output after running `cat output/eve.json | jq 'select(.dns)`:
```
{
  "timestamp": "2017-01-27T16:03:18.623093+0000",
  "flow_id": 1831739251032567,
  "pcap_cnt": 1,
  "event_type": "dns",
  "src_ip": "10.16.1.11",
  "src_port": 59465,
  "dest_ip": "8.8.4.4",
  "dest_port": 53,
  "proto": "UDP",
  "pkt_src": "wire/pcap",
  "dns": {
    "type": "query",
    "id": 33429,
    "rrname": "dne.oisf.net",
    "rrtype": "A",
    "tx_id": 0,
    "opcode": 0
  }
}
{
  "timestamp": "2017-01-27T16:03:18.709160+0000",
  "flow_id": 1831739251032567,
  "pcap_cnt": 2,
  "event_type": "alert",
  "src_ip": "8.8.4.4",
  "src_port": 53,
  "dest_ip": "10.16.1.11",
  "dest_port": 59465,
  "proto": "UDP",
  "pkt_src": "wire/pcap",
  "tx_id": 1,
  "alert": {
    "action": "allowed",
    "gid": 1,
    "signature_id": 1,
    "rev": 1,
    "signature": "",
    "category": "",
    "severity": 3
  },
  "dns": {
    "answer": {
      "version": 2,
      "type": "answer",
      "id": 33429,
      "flags": "8183",
      "qr": true,
      "rd": true,
      "ra": true,
      "opcode": 0,
      "rrname": "dne.oisf.net",
      "rrtype": "A",
      "rcode": "NXDOMAIN",
      "authorities": [
        {
          "rrname": "oisf.net",
          "rrtype": "SOA",
          "ttl": 899,
          "soa": {
            "mname": "ns-110.awsdns-13.com",
            "rname": "awsdns-hostmaster.amazon.com",
            "serial": 1,
            "refresh": 7200,
            "retry": 900,
            "expire": 1209600,
            "minimum": 86400
          }
        }
      ]
    }
  },
  "app_proto": "dns",
  "direction": "to_client",
  "flow": {
[...]

PR #1687, opened 2024-03-05
test: add test for dns.rcode - v7

jufajardini:
Feature #6621

Previous PR: #1664 

Changes from Previous PR:
- edit pcap HEX to edit the `rcode` header section of the request message
- add a rule that matches on the `to_server` direction
- edit README to explain where the pcap comes from

Suricata PR: https://github.com/OISF/suricata/pull/10458

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6621


inashivb:
Thanks for doing this!

>     * edit pcap HEX to edit the `rcode` header section of the request message

Why was this required?

jufajardini:
> Thanks for doing this!
> 
> > ```
> > * edit pcap HEX to edit the `rcode` header section of the request message
> > ```
> 
> Why was this required?

Because `dns.rcode` can be used in both directions, but as a response code, it is not commonly seen in the request messages, so I added it directly, so we could test it. Should I improve on the text?

victorjulien (approved):

PR #1688, opened 2024-03-05

coledishington:

This patch is using grep on stats.log, rather than usual eve.json, to test for existence of flow.get_used or flow.get_used_eval rather than a specific number.

Feature: #6293

Redmine ticket:


tests/flow-force-reuse/flow-force-reuse-on/test.yaml
```diff
@@ -0,0 +1,13 @@
...
+args:
+- --set flow.memcap-policy=drop-flow
+- --set flow.prealloc=0
+- --set flow.memcap=4kb
+- --set flow.hash-size=32
+- --set flow.force-reuse=true
+
+checks:
+  - shell:
+      args: grep -q 'flow.get_used \|flow.get_used_eval ' stats.log && printf '1\n' || printf '0\n'
```
victorjulien, line 12: I prefer using the eve.json based checks

victorjulien (changes requested):
This will need to pass on main-7.0.x and master-6.0.x to be considered for merge.

Please also see the inline comment.

tests/flow-force-reuse/flow-force-reuse-on/test.yaml
coledishington: This patch is using grep on stats.log, rather than usual eve.json, to test for existence of flow.get_used or flow.get_used_eval rather than a specific number.
Is there a way to do this with eve.json?

Thanks

victorjulien: would this work? https://github.com/OISF/suricata-verify/blob/master/README.md?plain=1#L148

coledishington:
Addressed comments in https://github.com/OISF/suricata-verify/pull/1699 . Thanks

PR #1699, opened 2024-03-12

coledishington:

Feature: #6293

Redmine ticket:


victorjulien (approved):

victorjulien:
Holding off for now, as I'm seeing the tests fail on my arm64 CI runners (private setup). Any idea why that may fail?
Only info I have is:
```
===> flow-force-reuse-on: Sub test #1: FAIL : expected 0 matches; got 1 for filter {'count': 0, 'event_type': 'stats', 'match': {'stats.flow.get_used': 0}}
```

victorjulien (changes requested):
setting "request changes" state to avoid my tooling to see it as accepted still

coledishington:
> Holding off for now, as I'm seeing the tests fail on my arm64 CI runners (private setup). Any idea why that may fail? Only info I have is:
> 
> ```
> ===> flow-force-reuse-on: Sub test #1: FAIL : expected 0 matches; got 1 for filter {'count': 0, 'event_type': 'stats', 'match': {'stats.flow.get_used': 0}}
> ```

It looks like all of the increments of that stat should be controlled by the return value of FlowGetUsedFlow(), which should always return NULL if flow_config.force_reuse is off.

Without extra logs it is hard to tell but my guess would be it is related to memcap and/or hash-size, although I can't see any differences for arm64 compared to the others. Running it with SCLogDebug() logs should show if memcap and hash-size look correct.

Thanks

victorjulien:
SV now has the capability to run tests only on a specific OS and ARCH (see https://github.com/OISF/suricata-verify/commit/a144407e830fe1a69b0111c06750f76b9c280e79, https://github.com/OISF/suricata-verify/commit/435c06997cf5f4f893e48082839de3d711458815, example https://github.com/OISF/suricata-verify/commit/118e4f82d2022ede97a24638bae8789a5aed69bf)

coledishington:
@victorjulien Yep, I am happy to make it only run on only the passing archs, but I am still confused why the change acted differently on arm64. We have platforms that are arm64 that run the feature (not the test), so I think it is a test issue. Does the github SV, or similar,  have arm64 nodes that I can get logs from?

victorjulien:
A quick look at this suggests that some data structs are slightly bigger on arm64 than on amd64, at least on my Ubuntu 22.04 test:
pthread_mutex_t (part of Flow, FlowBucket) is 40 bytes on amd64, 48 bytes on arm64. This would lead to different memuse.

catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/2021

PR #1702, opened 2024-03-13
http: adds check for request line missing protocol

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6856


victorjulien (changes requested):
Need it to pass on 6 and 7, not wanting them to be in a broken state. Esp so close to the release

catenacyber:
Indeed, see https://github.com/OISF/suricata-verify/pull/1704

PR #1704, opened 2024-03-14
http: adds check for request line missing protocol

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6856

#1702 with check only added from version 8


catenacyber:
> Can you also add a test for the 2221055 rule, that it matches?

Done in https://github.com/OISF/suricata-verify/pull/1705

PR #1719, opened 2024-03-21
Bug 5220/v7

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5220

Previous PR: https://github.com/OISF/suricata-verify/pull/1703

Changes since v6:

### DIFF
--- a/tests/bug-5220/bug-5220-2/README.md
+++ b/tests/bug-5220/bug-5220-2/README.md
@@ -1,0 +1,8 @@
+# Test Description
+This test demonstrates an invalid combination of base64_data with fast_pattern.
+
+## PCAP
+None
+
+## Related issues
+https://redmine.openinfosecfoundation.org/issues/5220
--- a/tests/bug-5220/bug-5220-4/README.md
+++ b/tests/bug-5220/bug-5220-4/README.md
@@ -1,0 +1,10 @@
+# Test Description
+This test demonstrates a invalid use of fast_pattern along with base64_data
+where base64_data is succeeded by a buffer keyword and the content right
+after that is fast_pattern.
+
+## PCAP
+None
+
+## Related issues
+https://redmine.openinfosecfoundation.org/issues/5220
--- a/tests/bug-5220/bug-5220-6/test.yaml
+++ b/tests/bug-5220/bug-5220-6/test.yaml
@@ -1,0 +1,8 @@
+requires:
+  pcap: false
+  min-version: 7
+
+args:
+  - --engine-analysis
+
+exit-code: 0
--- a/tests/bug-5220/bug-5220-1/README.md
+++ b/tests/bug-5220/bug-5220-1/README.md
@@ -1,0 +1,9 @@
+# Test Description
+This test demonstrates that fast_pattern along with base64_data
+should fail with Suricata instead of silently passing through.
+
+## PCAP
+None
+
+## Related issues
+https://redmine.openinfosecfoundation.org/issues/5220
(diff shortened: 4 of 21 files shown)


tests/bug-5220/bug-5220-2/README.md
```diff
@@ -0,0 +1,8 @@
+# Test Description
+This test demonstrates an invalid combination of base64_data with fast_pattern.
```
catenacyber, line 2: Looks like the same as first test to me. What is the added value ?

tests/bug-5220/bug-5220-4/README.md
```diff
@@ -0,0 +1,10 @@
+# Test Description
+This test demonstrates a invalid use of fast_pattern along with base64_data
+where base64_data is succeeded by a buffer keyword and the content right
+after that is fast_pattern.
```
catenacyber, line 4: README et rule do not seem to match

tests/bug-5220/bug-5220-6/test.yaml
```diff
@@ -0,0 +1,8 @@
+requires:
+  pcap: false
+  min-version: 7
+
+args:
+  - --engine-analysis
+
+exit-code: 0
```
catenacyber, line 8: Should we test the output for what is used as fast pattern in the end ?

tests/bug-5220/bug-5220-2/README.md
inashivb: Not much with the way `base64_data` is setup but I tried to create different combinations that are allowed rn w.r.t. the `base64_data` and `fast_pattern` keywords to see if anything would happen differently.
Do you think this test should be removed?

tests/bug-5220/bug-5220-4/README.md
inashivb: Why?
`base64_decode:bytes 28; base64_data; bsize:=21; content:"something"; fast_pattern;`

> base64_data is succeeded by a buffer keyword and the content right
after that is fast_pattern

tests/bug-5220/bug-5220-6/test.yaml
inashivb: Will add that. Thank you!

tests/bug-5220/bug-5220-4/README.md
catenacyber: What do you mean by buffer keyword ? I would expect `http.uri` or such

tests/bug-5220/bug-5220-2/README.md
catenacyber: Maybe rather more in the README to explain the difference : multiple `base64_data` with one which is fast_pattern... or such

And why it matters : I would expect rule parser to stop after fast_pattern and not care about the end of the rule...

tests/bug-5220/bug-5220-4/README.md
inashivb: I'm sorry. Because `bsize` is for buffer size, I called it a buffer keyword. Should have written payload keyword instead. Will fix. Thank you very much! 🙇🏽‍♀️

tests/bug-5220/bug-5220-2/README.md
inashivb: You're right about the why part. I think I added all these different combinations for myself to understand things better and walk code in diff scenarios. I'll fix the README and also add that the rule parser stops after the fast_pattern. Thank you!

PR #1723, opened 2024-03-21
tests/imap-detection: add simple test for protocol detection - v2

mmaatuq:
## Ticket #2886

Redmine ticket: [imap protocol detection is incomplete](https://redmine.openinfosecfoundation.org/issues/2886)
Relates to this [PR](https://github.com/OISF/suricata/pull/10661)
changes from PR #1620 : rebased to latest master branch.


tests/imap-detection/test.yaml
```diff
@@ -0,0 +1,36 @@
+requires:
+  files:
+    - src/app-layer-imap.c
```
catenacyber, line 3: I do not think we should have this requirement.

So, CI will be red until suricata PR gets merged...

And so, a renaming of this file would still run the test in the future...

PR #1737, opened 2024-03-30
smtp: adds test with data command being rejected

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6906


catenacyber:
Continued in #1840

PR #1745, opened 2024-04-04
tests: add checks for hiding zero counter stats - v1

jufajardini:
Task #5976

Add a simple test to showcase this. Don't know if we want more tests or more checks, or one with and one without, for comparison purposes.

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5976


jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1748

PR #1748, opened 2024-04-08
tests: add checks for hiding zero counter stats - v2

jufajardini:
Task #5976

Add a simple test to showcase this. Don't know if we want more tests or more checks, or one with and one without, for comparison purposes.

Previous PR: #1745

Changes from previous PR:
- update suricata.yaml config option

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/5976

### DIFF
--- a/tests/feature-5976-zero-stats-counters-hidden/test.yaml
+++ b/tests/feature-5976-zero-stats-counters-hidden/test.yaml
@@ -1,0 +1,42 @@
+requires:
+  features:
+    - DEBUG
+pcap: ../tls/tls-certs-alert/input.pcap
+args:
+- --simulate-ips
+- -k none
+# pretend pretend error in the first data
+- --simulate-applayer-error-at-offset-ts=0
+- --set app-layer.error-policy=drop-flow
+checks:
+  - filter:
+      min-version: 7
+      count: 1
+      match:
+        event_type: stats
+        stats.ips.drop_reason.applayer_error: 1
+  - filter:
+      min-version: 8
+      count: 0
+      match:
+        event_type: stats
+        stats.ips.blocked: 0
+  - filter:
+      min-version: 8
+      count: 0
+      match:
+        event_type: stats
+        stats.ips.rejected: 0
+  - filter:
+      min-version: 8
+      count: 0
+      match:
+        event_type: stats
+        stats.ips.drop_reason.decode_error: 0
+  - filter:
+      min-version: 8
+      count: 1
+      match:
+        event_type: stats
+        stats.ips.accepted: 3
+        stats.decoder.pkts: 32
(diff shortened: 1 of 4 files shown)


jufajardini:
Suricata PR: https://github.com/OISF/suricata/pull/10787

victorjulien:
Why does this test already pass in master?

tests/feature-5976-zero-stats-counters-hidden/test.yaml
```diff
@@ -0,0 +1,42 @@
...
+      count: 0
+      match:
+        event_type: stats
+        stats.ips.rejected: 0
+  - filter:
+      min-version: 8
+      count: 0
+      match:
+        event_type: stats
+        stats.ips.drop_reason.decode_error: 0
```
jasonish, line 35: We probably need some tests that use the `not-has-key` check to verify that stats we know to be 0 don't exist.

jasonish, line 35: We probably need some tests that use the not-has-key check to verify that stats we know to be 0 don't exist.

jasonish (changes requested):

tests/feature-5976-zero-stats-counters-hidden/test.yaml
jufajardini: Thanks, I had forgotten about this usage, again >__<

jufajardini:
> Why does this test already pass in master?

Possibly I'm not checking the existence of the zero-ed counters the best way? I'll see what I get locally, and make sure to make those fail with master beforehand.

jufajardini:
> > Why does this test already pass in master?
> 
> Possibly I'm not checking the existence of the zero-ed counters the best way? I'll see what I get locally, and make sure to make those fail with master beforehand.

Hmm... these fail locally.
I have a vague memory of a discuss around that somewhere. I'll try to retrieve it, to see if there's any lesson to be learned. Besides Jason's note, too.

jasonish:
> > Why does this test already pass in master?
> 
> Aha. So that's because this test requires `--enable-debug` and the checks don't have that option, so the test is sneakly skipped. Would it make sense to have some sort of warning if the added test was skipped for the PR, since not all SV PR are accompanied by a Suri PR?

I don't think I follow.  Why is `--enable-debug` required?  Is `--enabled-debug` required for the suppressing of 0 valued counters as well?

jufajardini:
> > > Why does this test already pass in master?
> > 
> > 
> > Aha. So that's because this test requires `--enable-debug` and the checks don't have that option, so the test is sneakly skipped. Would it make sense to have some sort of warning if the added test was skipped for the PR, since not all SV PR are accompanied by a Suri PR?
> 
> I don't think I follow. Why is `--enable-debug` required? Is `--enable-debug` required for the suppressing of 0 valued counters as well?

No, it's more a result of the test I choose to use, which I'll change. `debug` was needed for simulating an app-layer error, to see exception-policies in action - and this was the test that inspired this one. I'll create one that doesn't require it.

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1753

PR #1756, opened 2024-04-12
tests: add defrag datalink tests

victorjulien:
Bug: 6887. https://redmine.openinfosecfoundation.org/issues/6887
Bug: 6932. https://redmine.openinfosecfoundation.org/issues/6943

Includes a disabled test for https://redmine.openinfosecfoundation.org/issues/6942


catenacyber:
> Includes a disabled test for https://redmine.openinfosecfoundation.org/issues/6942

Why ?
Should not this test be included when you get a PR to fix this ?

Otherwise, it looks like you should set min-version to 8 for the first round ;-)

PR #1757, opened 2024-04-12
http: test FP for http.response_body

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6948


catenacyber:
I guess I should create a new PR with min-version: 8, right ?

PR #1770, opened 2024-04-15
tests/pgsql: add checks and test for bug 6092 - v1

jufajardini:
Ensure that pgsql metadata flags (for now, just setting whether passwords should be logged or not) are properly processed by Suri and logging functions.

Related to
Bug #6092

Updated existing tests where there could be logged a password message, but we disable that in the configuration. Add two tests specifically to test this when passwords enabled or disabled.

The Suri PR doesn't change the output behavior, just ensures that we pass pgsql config flags as metadata, therefore, these tests should pass with master, too.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6092

### DIFF
--- a/tests/pgsql/pgsql-5000-query-results/test.yaml
+++ b/tests/pgsql/pgsql-5000-query-results/test.yaml
@@ -64,6 +64,7 @@ checks:
       dest_port: 5432
       event_type: pgsql
       pcap_cnt: 29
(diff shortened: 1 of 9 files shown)


victorjulien (approved):

victorjulien:
Merged in #1771, thanks!

PR #1781, opened 2024-04-17
tests: update ips-state-1 test - v3

jufajardini:
This test indicated that there were FP drops for http and that another check was failing, but currently the are no more FP for http.

Updated the checks to reflect this.

flow.action still not set to drop with the tls drops...

Previous PR: #1496

Changes from previous PR:
- rebase
- add a few more checks - for drop events, for all existing alerts, for drop reason...
- improve README description - kept the previous comment about the TLS behavior

## Ticket

Redmine ticket:
none, so far. Should I create one, as in, do we think there's a possible bug here?

### DIFF
--- a/tests/ips-state-1/README.md
+++ b/tests/ips-state-1/README.md
@@ -1,13 +1,11 @@
 ## PCAP
 
-This PCAP contains 3 flows.  2 are http and one is TLS. The HTTP flows should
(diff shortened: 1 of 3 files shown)


catenacyber:
> Should I create one, as in, do we think there's a possible bug here?

I would say yes

PR #1788, opened 2024-04-18
Threshold flow/v5

victorjulien:
#1701 rebased


tests/threshold/threshold-rule-flow/test.yaml
```diff
@@ -0,0 +1,24 @@
...
+      match:
+        event_type: alert
+        alert.signature_id: 2
+  - filter:
+      count: 2
+      match:
+        event_type: alert
+        alert.signature_id: 3
+  - filter:
+      count: 1
```
catenacyber, line 21: Why is there only one alert ?
Could you add a comment about `threshold:type both` intention ?

tests/threshold/threshold-rule-flow/README.md
```diff
@@ -0,0 +1,3 @@
+PCAP
+====
+PCAP from https://wiki.wireshark.org/Internet_Control_Message_Protocol
```
catenacyber, line 3: Could you add the purpose of this test ?

tests/threshold/detection_filter-rule-flow/test.yaml
```diff
@@ -0,0 +1,34 @@
+requires:
+  min-version: 8
+
```
catenacyber, line 3: Cannot you reuse the pcap with `pcap:` instead of duplicating it ?

tests/threshold/detection_filter-rule-hostsrc/test.rules
```diff
@@ -0,0 +1,6 @@
+alert icmp any any -> any any (itype:8; sid:1;)
+alert icmp any any -> any any (itype:8; detection_filter:track by_src, count 1, seconds 3600; sid:2;)
+alert icmp any any -> any any (itype:8; detection_filter:track by_src, count 2, seconds 3600; sid:3;)
+alert icmp any any -> any any (itype:8; detection_filter:track by_src, count 4, seconds 3600; sid:4;)
+alert icmp any any -> any any (itype:8; detection_filter:track by_src, count 7, seconds 3600; sid:5;)
+alert icmp any any -> any any (itype:8; detection_filter:track by_src, count 8, seconds 3600; sid:6;)
```
catenacyber, line 6: Why this test ?
As there is only one flow, we do not see the difference between ` detection_filter:track by_src` and ` detection_filter:track by_flow` done in tests/threshold/detection_filter-rule-flow

tests/threshold/threshold-config-rate-filter-alert-flow/input.rules
```diff
@@ -0,0 +1 @@
+drop tcp any any -> any any (dsize:0; sid: 1000001;)
```
catenacyber, line 1: What does this rule mean to do ?

tests/threshold/threshold-rule-flow/test.yaml
victorjulien: No, this is not the place to document features. Check the thresholding docs to learn how the thresholding mechanics work.

tests/threshold/threshold-rule-flow/README.md
victorjulien: Not really seeing the point. Test name says it all.

tests/threshold/detection_filter-rule-flow/test.yaml
victorjulien: yeah, doing that

tests/threshold/detection_filter-rule-hostsrc/test.rules
victorjulien: it tests a different code path

tests/threshold/threshold-config-rate-filter-alert-flow/input.rules
victorjulien: it's a generic per packet drop rule that we convert to alert based on thresholding

victorjulien:
replaced by #1822

tests/threshold/detection_filter-rule-hostsrc/test.rules
catenacyber: Yes, but we do not prove that since we get the same behavior

tests/threshold/threshold-rule-flow/test.yaml
catenacyber: The numbers still do not look obvious to me... Why did you pick 3 ?

tests/threshold/threshold-rule-flow/README.md
catenacyber: I think there is a generic template for readme now, with also redmine ticket involved... cc @jufajardini ?

tests/threshold/threshold-config-rate-filter-alert-flow/input.rules
catenacyber: So, we drop the first N packets, then we alert ?

tests/threshold/threshold-rule-flow/README.md
jufajardini: Indeed, when one uses createst, it generates a readme file that expects: Description, Pcap, and Ticket/related issues). I personally find it useful to add something more descriptive in case people use SV tests to understand how something works, or as examples.

tests/threshold/threshold-config-rate-filter-alert-flow/input.rules
jufajardini: This sort of question I try to answer in the readme, for instance :P

PR #1794, opened 2024-04-22
tests: update ips-state-1 test - v3

jufajardini:
This test indicated that there were FP drops for http and that another check was failing, but currently the are no more FP for HTTP.  Updated the checks to reflect this. flow.action still not set to drop with the tls drops...

Related to
Bug #6976

Previous PR: #1781 

Changes from last PR:
- rebased
- added ticket reference
- add `lt-version` for checks specific to 6.0.x, as the number of generated alerts for TLS is different - not sure if this is expected, or another bug. I seem to remember some work around packets done by Philippe which I imagine could impact this, but can't recall more than this...

## Ticket

Redmine ticket:

### DIFF
--- a/tests/ips-state-1/README.md
+++ b/tests/ips-state-1/README.md
@@ -1,13 +1,15 @@
 ## PCAP
 
-This PCAP contains 3 flows.  2 are http and one is TLS. The HTTP flows should
+This PCAP contains 3 flows. 2 are http and one is TLS. The HTTP flows should
 be full passed with no alerts, while the TLS flow should be dropped.
 
 ## Current Observations
 
-- HTTP response packets are being logged as dropped, however the transaction is
-  logged suggesting the drop is only in logging only, but not actually
-  occurring.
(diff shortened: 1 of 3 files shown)


catenacyber:
Why are all the branches failing ?
Should there be some `min-version: 8` ?

jufajardini:
> Why are all the branches failing ? Should there be some `min-version: 8` ?

I left it run with all Suri versions, because we have `flow.action` as a log field, so thought it's worth checking. The test for `flow.action: pass` works, so I think it makes sense that the check for `flow.action: drop` should, too... 🤔

catenacyber (changes requested):
Need to be green for version 6 and 7 even if red for master, right ?

jufajardini:
> Need to be green for version 6 and 7 even if red for master, right ?

Why, if 6 and 7 also have `flow.action`? Or more like in terms of incremental fixes?

jufajardini:
Replaced by https://github.com/OISF/suricata-verify/pull/1973

PR #1822, opened 2024-05-07
Threshold flow/v6

victorjulien:
#1788 rebased with a pcap deduplication


tests/threshold/detection_filter-rule-flow/test.yaml
```diff
@@ -0,0 +1,36 @@
...
+pcap: ../threshold-rule-flow/icmp.pcap
+
+checks:
+  - filter:
+      count: 8
+      match:
+        event_type: alert
+        alert.signature_id: 1
+  - filter:
+      count: 7
```
catenacyber, line 13: I do not understand why it is not 8, `count 1` should be a nop, right ?

tests/threshold/threshold-config-threshold-both-flow/test.yaml
```diff
@@ -0,0 +1,21 @@
+pcap: ../../http-all-headers/input.pcap
+
+requires:
+  min-version: 8
+
+args:
+- --set threshold-file=${TEST_DIR}/threshold.config
+- --simulate-ips
+
+retry: 3
```
catenacyber, line 10: What is this retry ?
Flaky test ?

tests/threshold/threshold-config-rate-filter-alert-flow/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+        event_type: alert
+        alert.signature_id: 1000001
+  - filter:
+      count: 1
+      match:
+        event_type: drop
+  - filter:
+      count: 1
+      match:
+        event_type: http
```
catenacyber, line 25: Why does this http event matter ?

tests/threshold/threshold-config-rate-filter-alert-flow/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+checks:
+  - filter:
+      count: 19
+      match:
+        event_type: alert
+        alert.signature_id: 1000001
+  - filter:
+      count: 1
+      match:
+        event_type: drop
```
catenacyber, line 21: Would like a comment or README stating something like

The first packet (syn without payload) triggers the rule with its original action being a drop
Then, all 19 "empty" packets (Wireshark filter `tcp && !(len(tcp.payload) > 0)`) (including the first one) will trigger the new action defined by the rate filter so alert (and no drop)

tests/threshold/threshold-config-rate-filter-alert-flow/threshold.config
```diff
@@ -0,0 +1 @@
+rate_filter gen_id 1, sig_id 1000001, track by_flow, count 1, seconds 60, new_action alert, timeout 1000
```
catenacyber, line 1: Would like a comment stating that `seconds 60` does not matter, could be `seconds 1` as we count only one event

tests/threshold/threshold-config-threshold-limit-flow/threshold.config
```diff
@@ -0,0 +1 @@
+threshold gen_id 1, sig_id 1000001, type limit, track by_flow, count 1, seconds 60
```
catenacyber, line 1: Could we test another value for count ? Like 10 out of the 19

tests/threshold/threshold-config-threshold-threshold-flow/threshold.config
```diff
@@ -0,0 +1 @@
+threshold gen_id 1, sig_id 1000001, type threshold, track by_flow, count 5, seconds 60
```
catenacyber, line 1: Test name `threshold-config-threshold-threshold-flow` has a majority of `threshold` words :-p

Could some comment state that 19 // 5 == 3 (+ remainder 4 which did not reach the threshold yet)

victorjulien:
Not planning to update this further, cleanups and improvements can be done in follow up work.

catenacyber:
> Not planning to update this further, cleanups and improvements can be done in follow up work.

Do you want me to propose a new version, or wait before this one is merged ?

victorjulien:
We can move ahead with this now, I think.

tests/threshold/threshold-config-threshold-both-flow/test.yaml
jasonish: The retry was first added for tests that are sensitive to timing and may need a few goes to test properly, so not really flaky, but kinda. If a test uses it, it should probably state why, or you can assume its sensitive to timing.

tests/threshold/detection_filter-rule-flow/test.yaml
victorjulien: "The detection_filter keyword can be used to alert on every match after a threshold has been reached."
So matching starts after threshold is reached. The logic is not new btw, just also implemented for `by_flow`.

tests/threshold/threshold-config-rate-filter-alert-flow/test.yaml
victorjulien: traffic flows as expected

tests/threshold/threshold-config-rate-filter-alert-flow/test.yaml
victorjulien: Not seeing the point. Test is simple enough.

tests/threshold/threshold-config-rate-filter-alert-flow/threshold.config
victorjulien: updated the test to use count 3

tests/threshold/threshold-config-threshold-limit-flow/threshold.config
victorjulien: done in another test now

victorjulien:
replaced by #1891

tests/threshold/threshold-config-rate-filter-alert-flow/threshold.config
catenacyber: even better

tests/threshold/threshold-config-threshold-limit-flow/threshold.config
catenacyber: Seen in tests/threshold/threshold-rule-flow/test.rules ok

PR #1824, opened 2024-05-07
test: test for lua rules erroring out - v1

jasonish:
test: lua-rule-errors

Test for correct behvaior when a Lua rule crashes.

On crash, a Lua script should not result in a match when negation is
applied. Also test for:
- error stat
- one time logging per thread

Ticket: https://redmine.openinfosecfoundation.org/issues/6940

Expected to fail on current branches.


jasonish:
Closing for https://github.com/OISF/suricata-verify/pull/1826.

PR #1825, opened 2024-05-08
tests/ja4: Enable ja4 tests in 7.0.6 and later

jlucovsky:
Issue: 7010

Enable the JA4 tests for Suricata 7.0.6 and later.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7010


tests/ja4-rules-requires-off/test.yaml
```diff
@@ -1,10 +1,11 @@
 requires:
-  min-version: 8.0.0
+  min-version: 7.0.6
   script:
     - ./src/suricata --build-info | grep "JA4 support" | grep no > /dev/null
 
 args:
   - -k none
+  - --set logging.default-log-level=info
```
catenacyber, line 8: Why this addition ?

tests/ja4-rules/test.yaml
```diff
@@ -27,3 +28,6 @@ checks:
       match:
         event_type: alert
         alert.signature: JA4 TLS Test 1
+  - shell:
```
catenacyber, line 31: Should this new check be in its own commit ?

jlucovsky: Yes. I'll prepare a new branch/pr

tests/ja4-rules-requires-off/test.yaml
catenacyber: Is it Info in 7, and something else in 8 ?

jlucovsky: It was missing from the original test; the same level is used for 7 and 8.

jlucovsky:
Continued in #1830

tests/ja4-rules-requires-off/test.yaml
catenacyber: So why do we need it ?
Tests were passing for 8...

PR #1826, opened 2024-05-08
test: test for lua rules erroring out - v2

jasonish:
Test for correct behvaior when a Lua rule crashes.

On crash, a Lua script should not result in a match when negation is
applied. Also test for:

- error stat
- one time logging per thread
- 
Ticket: https://redmine.openinfosecfoundation.org/issues/6940

Expected to fail on current branches.

New version pushed to break some caching.


tests/lua/lua-bad-script/test.yaml
```diff
@@ -0,0 +1,44 @@
...
+      count: 8
+      match:
+        alert.signature_id: 2
+
+  - filter:
+      count: 0
+      match:
+        alert.signature_id: 3
+
+  # PCAP has 8 packets, which means 8 Lua failures.
```
catenacyber, line 31: Could you reuse an existing pcap instead of putting a new one ?

tests/lua/lua-bad-script/test.yaml
```diff
@@ -0,0 +1,44 @@
...
+        event_type: stats
+        stats.detect.lua.errors: 8
+
+  # The Lua failure should only be logged once.
+  - filter:
+      filename: suricata.json
+      count: 1
+      match:
+        event_type: engine
+        engine.message.__find: Lua script failed to run successfully
```
catenacyber, line 44: Why not the full message ?

jasonish: "Lua script failed to run successfully: /home/jason/oisf/dev/suricata-verify/lua-rule-errors/tests/lua/lua-bad-script/error.lua:6: attempt to call a nil value (global 'foobar')"

victorjulien (approved):

victorjulien:
Merged in #1835, thanks!

PR #1827, opened 2024-05-09
    run.py: allow a test to require a feature not being enabled

jasonish:
Example:
```
requires:
  features:
    - -HAVE_LUA
```
would require that Suricata does not have Lua support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,9 @@ requires:
     # Restrict the test to builds with HAVE_LUA.
     - HAVE_LUA
 
+    # Only run a test when a feature is NOT present.
+    - -HAVE_JA4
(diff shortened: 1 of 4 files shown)


jlucovsky:
LGTM.

victorjulien:
Looks good overall. Wondering if `!HAVE_LUA` would work as well. Since that is such a common pattern in suri, it feels more appropriate.

catenacyber:
> Looks good overall. Wondering if `!HAVE_LUA` would work as well. Since that is such a common pattern in suri, it feels more appropriate.

The double dash bugged my eyes, so +1 for `!`

PR #1828, opened 2024-05-10
mqtt: add tests for MQTT log limiting

satta:
This PR adds some tests related to the upcoming MQTT message log limiting functionality.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6984


tests/mqtt-limit-msg-1/test.yaml
```diff
@@ -0,0 +1,58 @@
+requires:
+  features:
+    - HAVE_LIBJANSSON
```
catenacyber, line 3: not needed anymore

tests/mqtt-limit-msg-1/test.yaml
```diff
@@ -0,0 +1,58 @@
...
+  script:
+    - grep 'msg-log-limit' suricata.yaml.in > /dev/null
+
+args:
+  - -k none
+
+checks:
+
+  - filter:
+      count: 1
```
catenacyber, line 15: Why this check ?

tests/mqtt-limit-msg-fail/suricata.yaml
```diff
@@ -0,0 +1,16 @@
...
+---
+
+outputs:
+  - eve-log:
+      enabled: yes
+      filetype: regular
+      filename: eve.json
+      types:
+        - mqtt:
+            msg-log-limit: XXX
```
catenacyber, line 11: I am not sure this deserves a SV test, but fait enough for me

tests/mqtt-limit-msg-1/test.yaml
jufajardini: Ideally, yes, containing: a brief description, Pcap, and Ticket/related issues.

I personally find it useful to add something more descriptive in case people use SV tests to understand how something works, or as examples. It is also where I try to answer questions that might arise about the checks that are made by the test, or the setup used.

satta:
> I personally prefer to see these tests red before the PR gets merged

I see. Interesting, my approach so far was to only run the tests if the tested feature is available in the tested suricata version/code tree. Would you always like to have tests from s-v PRs only succeed if the corresponding suricata PR is merged?

tests/mqtt-limit-msg-1/test.yaml
satta: done

tests/mqtt-limit-msg-1/test.yaml
satta: This first check is just to make sure that nothing in the `connect` log event is changed if log limiting is selected. But technically it is not testing the actual log limiting functionality. I can remove it if it helps with clarity and understandability of the test.

satta: OK, will also add a `README.md` for these tests.

tests/mqtt-limit-msg-fail/suricata.yaml
satta: This is to test if Suricata actually exits at startup with an error if this is not a  valid value. I think this is better done here than in an unit test.

satta:
New PR upcoming.

catenacyber:
> Would you always like to have tests from s-v PRs only succeed if the corresponding suricata PR is merged?

Personnaly yes, so we see that the suricata PR brings an improvement.

satta:
> > Would you always like to have tests from s-v PRs only succeed if the corresponding suricata PR is merged?
> 
> Personnaly yes, so we see that the suricata PR brings an improvement.

I see, good to know. I always expected these branches here to be a bit polymorphic: 

* when run here with a Suricata version without support for the new features to be tested, the tests should be skipped -> test green here
* when run in the context of a Suricata PR _with_ support for the new features to be tested, run the tests and hopefully -> test green in Suricata PR

I assumed that's what the requirement definition functionality in the `test.yaml` was for.

catenacyber:
This is a detail, but if we want to dive into it :

There are 2 things here :
- you are right for the `requirement`, for example, that is meant to run only tests if I have my suricata compiled with lib magic for instance
- for PRs, I like to see that we go from red to green, as it is more visible than going from skipped to passed (where I need to recompile and check locally)

satta:
Got it, thanks!

PR #1837, opened 2024-05-19
detect-proto: add a test for DNS detected as DCERPC by PM

ilya-bakhtin:
a test for https://github.com/OISF/suricata/pull/11100


catenacyber:
There should be a min-version: 8 for first the first SV PR...
Then, there would be a second SV PR to decrease this from 8 to 7, with the corresponding Suricata backport

catenacyber (changes requested):
see comment above

catenacyber:
#1989 is the new version, right ?

PR #1843, opened 2024-05-21
Lz4/v3

victorjulien:
Various pcap logging tests.

PR #1844, opened 2024-05-21
Lz4/v4

victorjulien:
replaces #1843

### DIFF
--- a/tests/pcap-log-lz4-01/test.yaml
+++ b/tests/pcap-log-lz4-01/test.yaml
@@ -1,0 +1,14 @@
+requires:
+  features:
+    - liblz4
+  min-version: 8
+      
+pcap: ../bug-2482-01/proxyCONNECT_443.pcap
+
+args:
+  - --runmode=single
+
+checks:
+    - shell:
+        args: ls -l log.pcap.15233891*.lz4 | wc -l | xargs
+        expect: 2
--- a/tests/pcap-log-lz4-05-tunnel/test.yaml
+++ b/tests/pcap-log-lz4-05-tunnel/test.yaml
@@ -1,0 +1,14 @@
+requires:
+  features:
+    - liblz4
+  min-version: 8
+      
+pcap: ../vxlan-decoder-01/input.pcap
+
+args:
+  - --runmode=single
+
+checks:
+    - shell:
+        args: ls -l log.pcap.1525778007.671337.lz4 | wc -l | xargs
+        expect: 1
--- a/tests/pcap-log-uncompressed-01/test.yaml
+++ b/tests/pcap-log-uncompressed-01/test.yaml
@@ -1,0 +1,12 @@
+requires:
+  min-version: 8
+      
+pcap: ../bug-2482-01/proxyCONNECT_443.pcap
+
+args:
+  - --runmode=single
+
+checks:
+    - shell:
+        args: find . -type f -name 'log.pcap.1523389*.*' | wc -l | xargs
+        expect: 3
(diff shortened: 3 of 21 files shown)


tests/pcap-log-uncompressed-01/test.yaml
```diff
@@ -0,0 +1,12 @@
...
+  min-version: 8
+      
+pcap: ../bug-2482-01/proxyCONNECT_443.pcap
+
+args:
+  - --runmode=single
+
+checks:
+    - shell:
+        args: find . -type f -name 'log.pcap.1523389*.*' | wc -l | xargs
```
catenacyber, line 11: What does `1523389` mean here ?

Would `file-compare` be a good check here rather than resorting to shell ?

tests/pcap-log-lz4-05-tunnel/test.yaml
```diff
@@ -0,0 +1,14 @@
...
+  min-version: 8
+      
+pcap: ../vxlan-decoder-01/input.pcap
+
+args:
+  - --runmode=single
+
+checks:
+    - shell:
+        args: ls -l log.pcap.1525778007.671337.lz4 | wc -l | xargs
```
catenacyber, line 13: Is this `log.pcap.1525778007.671337.lz4` supposed to always have the same content ?

This test would pass if the file exists and is empty for instance...

victorjulien: content as in packets, yes, content as in exact bytes: no. Platform differences can make them different (e.g. endianess)

tests/pcap-log-uncompressed-01/test.yaml
catenacyber: Is there not a config format string that you can use to ignore the timestamp ?

tests/pcap-log-lz4-05-tunnel/test.yaml
victorjulien: or perhaps run tshark on it, it supports the pcap.lz4 out of the box. In any case this can be done as a follow up effort.

tests/pcap-log-uncompressed-01/test.yaml
victorjulien: I explicitly added it to test it

victorjulien: which exposed https://redmine.openinfosecfoundation.org/issues/7034

catenacyber: Oh... 👏

catenacyber (approved):
Thanks for the answers

victorjulien:
Force push was: rebase to master, but more importantly a fix to a `grep` command using a GNUish that fails on my OpenBSD runner.

tests/pcap-log-lz4-01/test.yaml
```diff
@@ -0,0 +1,14 @@
...
+  min-version: 8
+      
+pcap: ../bug-2482-01/proxyCONNECT_443.pcap
+
+args:
+  - --runmode=single
+
+checks:
+    - shell:
+        args: ls -l log.pcap.15233891*.lz4 | wc -l | xargs
```
catenacyber, line 13: Why do we need xargs ?

catenacyber (approved):
Still good

tests/pcap-log-lz4-01/test.yaml
victorjulien: What I've seen it do is strip leading whitespaces `           1` -> `1`. Don't know if it is needed.

victorjulien:
Merged in #1848, thanks!

tests/pcap-log-lz4-01/test.yaml
inashivb: `-l` should ideally count lines I think so as long as the spaces leading or trailing are on the same line, it should work.

PR #1847, opened 2024-05-22
smtp/mime: update test to match new behavior

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3487

The suricata PR modifies the behavior, so both suricata and SV PR should be merged at the same time...


catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1851

PR #1853, opened 2024-05-24
tests: linktype_name test

jlucovsky:
Issue: 6954

Ensure that the linktype_name is included in the alerts.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6954

Suricata PR: https://github.com/OISF/suricata/pull/11207

### DIFF
--- a/tests/linktype_name/test.yaml
+++ b/tests/linktype_name/test.yaml
@@ -1,0 +1,14 @@
+requires:
+  min-version: 8
+
+pcap:  ../bug-2482-01/proxyCONNECT_443.pcap
+
+args:
+- -k none --set outputs.1.eve-log.types.0.alert.packet=yes
+
+checks:
+  - filter:
+      count: 86
+      match:
+        event_type: alert
+        packet_info.linktype_name: RAW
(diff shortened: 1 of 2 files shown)


tests/linktype_name/test.yaml
```diff
@@ -0,0 +1,12 @@
+requires:
+  min-version: 8
+
+args:
+- -k none --set outputs.1.eve-log.types.0.alert.packet=yes --set vars.address-groups.EXTERNAL_NET=any
```
catenacyber, line 5: Could we just change the rule to use `any` instead of `EXTERNAL_NET` ?

tests/linktype_name/test.yaml
```diff
@@ -0,0 +1,12 @@
+requires:
+  min-version: 8
+
```
catenacyber, line 3: Is there not a pcap to reuse instead of putting a new one ?

tests/linktype_name/test.yaml
jlucovsky: WIll do.

tests/linktype_name/test.yaml
jlucovsky: Good point; I'll make that adjustment.

catenacyber:
Fair enough for me, a README.md could be added to improve this ;-)

victorjulien:
Can we add a few more tests? I think it's fine to use existing tests. There are quite a few different linktypes in our SV suite I think.

jlucovsky:
Continued in #1993

PR #1855, opened 2024-05-25
lua: tests for lua sandbox - v2

jasonish:
- **run.py: add __startswith and __endswith for string matching**
- **run.py: use single mode if -j1 and on Linux**
- **test: add lua test for blocked lua library**
- **test: lua scripts exceeding the instruction limit**


jasonish:
Replace by https://github.com/OISF/suricata-verify/pull/1859.

PR #1858, opened 2024-05-27
tests: add tcp split handshake tests

victorjulien:


victorjulien:
Yes, this was addressed in the code many years ago, and surprisingly it still worked :)

tests/README.md
```diff
@@ -0,0 +1,3 @@
+Based on split handshake research by Tod Beardsley
+
+https://www.macrothink.org/journal/index.php/npa/article/view/285/807
```
catenacyber, line 3: Should be in a subdirectory

tests/tcp-split-handshake-01-4whs/test.rules
```diff
@@ -0,0 +1,2 @@
+alert tcp any any -> any any (content:"EVIL"; sid:1;)
+alert tcp-stream any any -> any any (content:"EVIL"; sid:2;)
```
catenacyber, line 2: Is there an expected difference in behavior for these 2 sigs ?

victorjulien: no, added `tcp-pkt` variant to make sure they all match

victorjulien:
replaced by #1959

PR #1836, opened 2024-05-19
mqtt: add tests for MQTT log limiting - v2

satta:
Previous PR: #1828 

Changes compared to previous PR:
* Add `README.md`s for all tests.
* Slim down test checks to be more descriptive for the reader; reduce to the aspects to be checked.
* Reduce duplication of pcap files by referencing across test cases.
* Remove some requirements: obsolete checks such as libjansson but also checks that would cause the test to be skipped because the associated Suricata PR wasn't merged yet (the latter at reviewer's request)

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6984

PR #1880, opened 2024-06-02
mqtt: add tests for MQTT log limiting -- v3

satta:
Previous PR: #1836 

Changes compared to previous PR:
* Rebase against current master.
* Rename tests to `mqtt-limit-log-*` to reflect the fact we're not only restricting messages anymore.
* Add test for topic array limiting.
* Remove unneeded requirements, add minimum version 8 requirement.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6984

### DIFF
--- a/tests/mqtt-limit-log-3/test.yaml
+++ b/tests/mqtt-limit-log-3/test.yaml
@@ -1,0 +1,19 @@
+pcap: ../mqtt5-sub-userpass/input.pcap
+
+requires:
+  min-version: 8
+
+args:
+  - -k none
+
+checks:
+
+  - filter:
+      count: 1
+      match:
+        event_type: mqtt
+        mqtt.subscribe.qos: 1
+        mqtt.subscribe.retain: false
+        mqtt.subscribe.dup: false
+        mqtt.subscribe.topics: [ {topic: topicX, qos: 0}, {topic: "topi[truncated 2 additional bytes]", qos: 0} ]
+        mqtt.subscribe.message_id: 1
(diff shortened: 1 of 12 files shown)


tests/mqtt-limit-log-3/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+checks:
+
+  - filter:
+      count: 1
+      match:
+        event_type: mqtt
+        mqtt.subscribe.qos: 1
+        mqtt.subscribe.retain: false
+        mqtt.subscribe.dup: false
+        mqtt.subscribe.topics: [ {topic: topicX, qos: 0}, {topic: "topi[truncated 2 additional bytes]", qos: 0} ]
```
catenacyber, line 18: Nit : Should we test `mqtt.subscribe.topics[1]` ? and `mqtt.subscribe.topics.__len: 2`

catenacyber (approved):
These SV tests look good, thanks Sascha

tests/mqtt-limit-log-3/test.yaml
satta: Doesn't line 18 test for exact equality of the result? The code in `FilterCheck::match()` seems to indicate so. Then I don't think there is a reason to add these two tests, because both the existence of `mqtt.subscribe.topics[1]` as well as the array length is already covered in the line above.

satta:
Thanks for taking a look!

tests/mqtt-limit-log-3/test.yaml
catenacyber: Yes it works.
This is a nit style remark...

victorjulien (approved):

PR #1889, opened 2024-06-05
Adds test about bidirectional rules

catenacyber:
## Ticket

#1786 with more impossible signature using  stream only


catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1910

PR #1897, opened 2024-06-06
tests/dns: add tests for task 7018 - v1

jufajardini:
Add tests for earlier raw stream reassembly for dns

Also related to
Bug #7004

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7018


jufajardini:
Failed to add version-check to protect 6.0.x

jufajardini:
> Failed to add version-check to protect 6.0.x

It seems that what's missing for 6.0.x for tests `task-7018-ips-stream-rule` and its ids counter-part are the fields related to TCP max_regions. Are those somehow also connected to stream reassembly, somehow?

jufajardini:
My observation was incomplete. The dns-keyword tests were passing for that test because they're 8.0.x.

Followed by: https://github.com/OISF/suricata-verify/pull/1902

PR #1907, opened 2024-06-10
dns: update for v3 dns request logging - v3

jasonish:
Previous PR: #1888 

Ticket: https://redmine.openinfosecfoundation.org/issues/6281


catenacyber:
That is a big diff !

Guess we can make a passing PR with only the removal commit first ?

I see more of it like `bug-2482-01` has checks for version 4 and version 5

PR #1906, opened 2024-06-10
datasets: add tests for string memcap

inashivb:
## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3910

PR #1909, opened 2024-06-11
Datasets memcap tests/v2

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3910

Previous PR: https://github.com/OISF/suricata-verify/pull/1906

Changes since v1:
- added a run.py key for os check
- fixed README for tests

### DIFF
--- a/run.py
+++ b/run.py
@@ -368,6 +368,9 @@ def check_requires(requires, suricata_config: SuricataConfig):
         elif key == "lambda":
             if not eval(requires["lambda"]):
                 raise UnsatisfiedRequirementError(requires["lambda"])
+        elif key == "os":
+            if not sys.platform.startswith(requires["os"]):
+                raise UnsatisfiedRequirementError(requires["os"])
         else:
             raise Exception("unknown requires types: %s" % (key))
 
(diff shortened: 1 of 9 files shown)


run.py
```diff
@@ -368,6 +368,9 @@ def check_requires(requires, suricata_config: SuricataConfig):
         elif key == "lambda":
             if not eval(requires["lambda"]):
                 raise UnsatisfiedRequirementError(requires["lambda"])
+        elif key == "os":
```
inashivb, line 371: @jasonish I want to check the OS that a test is being run on and skip if it isn't the expected one. Do you think this is a good way of testing that? Is there a better way?

jasonish: I guess `platform.system()` is known to have a little more consistent output.  So I'd use that.. `os` is good for the configuration key (remember to add to readme).  May as well do a lower case conversion before comparing as well, as `platform.system()` returns "Linux", "Windows", etc.

inashivb: Thank you! Done in the following PR.

PR #1912, opened 2024-06-11
Adds tests for negated content and absent keyword

catenacyber:
## Ticket

#1535 with new version
Also some test for frames

Need to understand why I cannot get it to work with only an absent frame...

Still need test for http.request_body


catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1914

PR #1913, opened 2024-06-12
Datasets memcap tests/v3

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3910

Previous PR: https://github.com/OISF/suricata-verify/pull/1909

Changes since v2:
- use `platform.system()` to determine the OS, update Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,10 @@ requires:
   # release, but 4.0.3 would only match 4.0.3.
   version: 4.0
 
+  # Test is only for the listed OS. For example, the following would make
+  # a test run only on Linux.
+  os: linux
+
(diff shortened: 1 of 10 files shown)


victorjulien (approved):

victorjulien (changes requested):
Fails on ARM
image

inashivb:
> Fails on ARM image

Based on this, I can deduce that on `armhf` the size of structs is again small as they accept the rule (i.e. exit code 0), not sure about `arm64` will check when I have access.
What does this mean for the memory tests; should there be an architecture `requires` arg too? @jasonish

PR #1914, opened 2024-06-12
Adds tests for negated content and absent keyword

catenacyber:
## Ticket

#1912 with more test for http.request_body dans files and multi buffer (http.request_header)


catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/1942

PR #1910, opened 2024-06-11
Adds test about bidirectional rules

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5665

#1889 with another test suggested by Brandon Murphy

PR #1922, opened 2024-06-18
Adds test about bidirectional rules

catenacyber:
## Ticket

#1910 with more tests
- negative for frames
- positive for tls with ja3 and ja3s (worked out of the box)


victorjulien:
Would like to see more test cases, for different protocols, as well as `--engine-analysis` output tests.

catenacyber:
> Would like to see more test cases, for different protocols

Which ones besides tls and http ?

> as well as `--engine-analysis` output tests.

Doing that

catenacyber:
Next in https://github.com/OISF/suricata-verify/pull/2083

PR #1936, opened 2024-06-22
applayer: add test for bug 7044

inashivb:
## Ticket

changes to the ticket status:

Note: This shows truncate fn is not needed. It is already taken care of.


inashivb:
@catenacyber what do you think of this test?

tests/truncate-applayer-test/test.yaml
```diff
@@ -0,0 +1,14 @@
+args:
+  - -k none --set stream.reassembly.depth=8kb
+
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 1
```
catenacyber, line 9: Could we check pcap_cnt ?
That this does not happen on flow timeout

catenacyber: Also check for dcerpc metadata now that they are available in master  cf https://github.com/OISF/suricata/pull/11151#issuecomment-2182390221

catenacyber: Oh and more important, we should test for a dcerpc event, not an alert !!

tests/truncate-applayer-test/test.yaml
```diff
@@ -0,0 +1,14 @@
...
+- filter:
+    count: 0    # because of the depth constraints
```
catenacyber, line 11: We could also have a second test without the `stream.reassembly.depth=8kb` to show that we get the alert

tests/truncate-applayer-test/test.yaml
inashivb: The alert happens at an exact dcerpc event. I had modified the stub_size settings in rule to have the exact buffer size where stream depth changes start to take place.
I will also add the dcerpc event though as it makes it easier to correlate the alert then. :)
Is there another reason that I'm missing perhaps for which you want to see the dcerpc event?

catenacyber: dcerpc event logging and alert are 2 different code paths.
We should test dcerpc event logging (as it is more bidirectional than alert)
cf https://github.com/OISF/suricata/pull/11151#issuecomment-2182390221

inashivb: ah I had missed the part explaining logger path in your comment. 🤦🏽‍♀️ 
Thank you! Taking one more look around the code paths and truncate usages.

PR #1930, opened 2024-06-20
output: adds checks for payload_length field

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7098

PR #1929, opened 2024-06-20
dcerpc: check for app-layer metadata in alert

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6090

PR #1863, opened 2024-05-28
Adds a test about filestore

catenacyber:
## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6390

https://github.com/OISF/suricata-verify/pull/1524 without references to Suricata 6 which will not be fixed

PR #1670, opened 2024-02-27
tests/transform: from_base64 test

jlucovsky:
from_base64 transform test

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket: 6487

PR #1937, opened 2024-06-22
next/sv/499/20240622/v1

victorjulien:
Staging:
 - #1863
 - #1929
 - #1930
 - #1670


tests/from_base64-02/test.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - filter:
+      count: 0
+      match:
+         event_type: alert
+         alert.signature_id: 2
+  - filter:
+      count: 0
+      match:
+         event_type: alert
+         alert.signature_id: 2
```
inashivb, line 24: To be followed up perhaps. @jlucovsky does this need to be `3`?

inashivb (approved):
LGTM 🚀

tests/from_base64-02/test.yaml
jlucovsky: Yes -- good eye. It should be `3`.

PR #1951, opened 2024-07-03
tests: showcase unexpected tcp rule behavior

jufajardini:
Showcase different scenarios for Suricata behavior in IPS mode with TCP rules and TLS traffic, with PASS or DROP actions.

Decided to share this as a WIP, as creating these for `case a` took me much longer than anticipated.

## Ticket

changes to the ticket status:

Redmine ticket:
- none yet


victorjulien:
Lets remove the unnecessary verbosity from the test.yaml, the goal is to understand a specific issue.

victorjulien:
Would also like to see a pcap that uses different IP addresses than localhost. Makes things much easier to reason about if the src/dst are different.

catenacyber:
Why does the test pass if it is unexpected behavior ? What is the expected behavior ?

jufajardini:
Closing this in favour of: https://github.com/OISF/suricata-verify/pull/1952 as it's more aligned with what we want to see for these tests.

jufajardini:
> Why does the test pass if it is unexpected behavior ? What is the expected behavior ?

There is a suspicion that the behavior of the TCP rule not passing is not what the rule language seems to express (cf https://github.com/OISF/suricata-verify/pull/1952/files#r1663786361)

PR #1952, opened 2024-07-03
tests: ips action test cases

victorjulien:
I think the most interesting thing here is the difference between 6 and 7/master for the "b" case.

Also, we can debate if "a-01" should apply the pass to the flow. I'll put an inline comment too.

### DIFF
--- a/tests/ips-state-actions-test-a-01/test.rules
+++ b/tests/ips-state-actions-test-a-01/test.rules
@@ -1,0 +1,5 @@
+pass udp any any -> any 6081 (sid:1;)
+pass tcp 2a03:b0c0:0002:00d0:0000:0000:0bd3:4001 any -> 2606:2800:0220:0001:0248:1893:25c8:1946 443 (msg:"PASS_CUSTOM_RULE TCP port:443 to support traffic"; flow:established; sid:201000044;)
+
+pass tcp 2a03:b0c0:0002:00d0::/64 any <> any [[80,443]] (msg:"PASS_HTTP_NOT_ESTABLISHED TCP allow http/https traffic to the established state to allow further inspection"; flow:not_established; sid:201000012;)
+drop ip any any -> any any (msg:"drop_strict action"; flow:stateless; sid:2;)
--- a/tests/ips-state-actions-test-a-01/test.yaml
+++ b/tests/ips-state-actions-test-a-01/test.yaml
@@ -1,0 +1,25 @@
+pcap: ../tls/tls-certs-alert/input.pcap
+
+args:
+  - --set stream.midstream=false
+  - -k none
+
+checks:
+  # drop is applied after app-layer parsing, so getting tls despite dropping
+  - filter:
+      count: 1
+      match:
+        event_type: tls
+  - filter:
+      count: 13
+      match:
+        event_type: drop
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+  - filter:
+      count: 0
+      match:
+        event_type: flow
+        flow.action: "drop"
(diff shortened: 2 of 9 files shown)


tests/ips-state-actions-test-a-01/test.rules
```diff
@@ -0,0 +1,5 @@
+pass udp any any -> any 6081 (sid:1;)
+pass tcp 2a03:b0c0:0002:00d0:0000:0000:0bd3:4001 any -> 2606:2800:0220:0001:0248:1893:25c8:1946 443 (msg:"PASS_CUSTOM_RULE TCP port:443 to support traffic"; flow:established; sid:201000044;)
```
victorjulien, line 2: This does not apply the `pass` to the flow, but should it? The flow starts as `not_established`, but after transitioning to `established` it stays that way, even if TCP state moves to "closed".

tests/ips-state-actions-test-b-01/test.yaml
```diff
@@ -0,0 +1,30 @@
...
+  - --set stream.midstream=false
+  - -k none
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: tls
+  - filter:
+      min-version: 7
```
victorjulien, line 13: This is a behavior change between 6 and 7+master. I think it makes sense, as the app_proto can change.

tests/ips-state-actions-test-a-01/test.yaml
```diff
@@ -0,0 +1,25 @@
+pcap: ../tls/tls-certs-alert/input.pcap
+
+args:
+  - --set stream.midstream=false
+  - -k none
+
```
jufajardini, line 6: isn't it important to run these as `IPS`?

victorjulien: they are, because ips is in the test name

jufajardini: Wait, if the test says ips-state SV will run them as IPS? 🤯

catenacyber:
Is this PR a draft to discuss current vs expected behavior ?

tests/ips-state-actions-test-a-01/test.rules
inashivb: That would indeed be coherent with the rule language as we read it.

tests/ips-state-actions-test-a-02/test.rules
```diff
@@ -0,0 +1,5 @@
+pass udp any any -> any 6081 (sid:1;)
+pass tcp 2a03:b0c0:0002:00d0:0000:0000:0bd3:4001 any <> 2606:2800:0220:0001:0248:1893:25c8:1946 443 (msg:"PASS_CUSTOM_RULE TCP port:443 to support traffic"; flow:established; sid:201000044;)
+
+pass tcp 2a03:b0c0:0002:00d0::/64 any <> any [[80,443]] (msg:"PASS_HTTP_NOT_ESTABLISHED TCP allow http/https traffic to the established state to allow further inspection"; flow:not_established; sid:201000012;)
```
jufajardini, line 4: Just to support better seen what's going on, couldn't we add `alert` to these pass rules?

tests/ips-state-actions-test-a-01/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+        event_type: drop
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+  - filter:
+      count: 0
+      match:
+        event_type: flow
+        flow.action: "drop"
```
jufajardini, line 25: Unrelated to this PR, just to see if we agree on this: shouldn't we see `flow.action` in the logs, here? I don't see it, at all.

tests/ips-state-actions-test-a-02/test.rules
jufajardini: Similarly, if we do, enable `verdict` for the alert events, too?

catenacyber:
Should this PR be a draft ?

victorjulien:
Unclear why I did this PR, will close it for now

PR #1953, opened 2024-07-03
applayer: add tests for bug 7044

inashivb:
## Ticket

changes to the ticket status:

Previous PR: https://github.com/OISF/suricata-verify/pull/1946

Changes since v2:
- a min-version check was added
Reason: Due to `truncate` callback in 6, a `REQUEST` was logged as a local variable `req_done` was set to `true`, however, the request logged in such a case was empty. So, it isn't much different from the behavior in `master` or `7.0.x` where the request is logged as "lost" bc the request never had anything and there was no callback to make it available for logging.


tests/truncate-applayer-test-01/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+requires:
+  min-version: 7
+
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 1
```
catenacyber, line 12: maybe add pcap_cnt ?

tests/truncate-applayer-test-02/README.md
```diff
@@ -0,0 +1,12 @@
+# Test Description
+
+This test demonstrates that the pcap is properly parsed and assessed in case
+there was no stream reassembly depth constraints.
```
catenacyber, line 4: there is the default one ;-)

catenacyber (approved):
Details can be added, but looks good to me to check current behavior and that truncate removal does not affect it

tests/truncate-applayer-test-02/README.md
inashivb: indeed. Wil change the words. Thanks :)

PR #1955, opened 2024-07-03
README/help: add info on IPS mode tests creation - v1

jufajardini:
It is possible to create a test that runs in IPS mode by just adding `ips` to its name. But that is not documented.

Task #7039

## Ticket

changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7039

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ Or to run a single test:
 
 - Create a directory that is the name of the new test.
 
+  If you want a test to run in IPS mode, add `ips` to the test name
+  this will make `--simulate-ips` command-line argument be passed when
+  the test is run.
+
 - Copy a single pcap file into the test directory. It must end in
   ".pcap".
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -200,6 +204,8 @@ directory.
 
 positional arguments:
   <test-name>           Name of the test folder
+                        Add `ips` to the test name if you want it to run in IPS mode
+                        the test is run.
```
inashivb, line 208: nits:
1. I'm realizing this README output is quite mature than the actual help output of the script. Maybe we should make it consistent. wdyt?
2. Help outputs are generally supposed to be one liners w minimal info. Should such additional info stay in the docs?

jufajardini: > nits:
> 
>     1. I'm realizing this README output is quite mature than the actual help output of the script. Maybe we should make it consistent. wdyt?
> 

Agreed!

>     2. Help outputs are generally supposed to be one liners w minimal info. Should such additional info stay in the docs?

I still think that this info should be outputted when we have the help, as I can imagine people would often first refer to the `help` output before going to the documentation. I can try to make it more concise, somehow, then.

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1965

PR #1959, opened 2024-07-04
tests: add tcp split handshake tests

victorjulien:
Address comments of #1858.


tests/split-handshake.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
```
catenacyber, line 1: should be in a subdirectory ?

victorjulien: fixed

victorjulien:
Merged in #1962, thanks!

PR #1975, opened 2024-07-10
tests: showcase flow.action lack of update - v1

jufajardini:
It seems that in certain cases as seen in this test, flow.action isn't updated, even if, say, all packets from the flow are dropped.

Maybe this is due to the rule not being applied directly to the flow, but to each packet individually. But considering we are using a flow keyword, it seems that the engine should pass over the drop action to `flow.action`, at least in the flow event.

Bug #6976

- This was started as part of https://github.com/OISF/suricata-verify/pull/1794, but Philippe suggested that I keet the changes separate, for ease of tracking the possible bug

## Ticket

changes to the ticket status:

Redmine ticket:

### DIFF
--- a/tests/flow-action-update-ips/README.md
+++ b/tests/flow-action-update-ips/README.md
@@ -1,0 +1,17 @@
+# Test
+
+This is a "fork" of `ips-state-1`, to showcase the `flow.action` update bug.
+
+## PCAP
+
+From `ips-state-1` test.
+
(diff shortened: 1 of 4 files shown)


victorjulien:
Please add an `--engine-analysis` test for the rules.

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/1987

PR #1985, opened 2024-07-19
rfb: adds checks for rules with secresult being an integer keyword

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6723

PR #1986, opened 2024-07-19
Rfb rust keywords 6723 v2

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6723

#1985 with separate test to have `min-version: 8`


jufajardini (commented):
Looks good to me, thanks for adding a test for the integer keyword usage.

The only thing to note is whether we'll have a separate ticket for these to-rust changes for RFB or not (in case we may need to update any ticket references here).

victorjulien:
Merged in #2003, thanks!

PR #1992, opened 2024-07-26
Detect prefilter events 6728 v1

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6728

### DIFF
--- a/tests/decode-too-small/test.rules
+++ b/tests/decode-too-small/test.rules
@@ -1,3 +1,4 @@
 alert tcp any any -> any any (msg:"TCP packet too small"; decode-event:tcp.pkt_too_small; sid:1;)
+alert tcp any any -> any any (msg:"TCP packet too small"; decode-event:tcp.pkt_too_small; prefilter; requires: version >= 8; sid:1; rev: 2;)
 alert udp any any -> any any (msg:"UDP packet too small"; decode-event:udp.hlen_too_small; sid:2;)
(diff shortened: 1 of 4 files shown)


tests/decode-too-small/test.rules
```diff
@@ -1,3 +1,4 @@
 alert tcp any any -> any any (msg:"TCP packet too small"; decode-event:tcp.pkt_too_small; sid:1;)
+alert tcp any any -> any any (msg:"TCP packet too small"; decode-event:tcp.pkt_too_small; prefilter; requires: version >= 8; sid:1; rev: 2;)
```
victorjulien, line 2: sid 2 is already in this file

victorjulien: ah no it's sid 1 rev 2

victorjulien: any reason to use this pattern? Might as well have 2 rules side by side, to test both prefilter and non-prefilter path

catenacyber: > any reason to use this pattern? 

This way, it works on both suricata 7 and suricata 8

If you have another rule side by side, you need the whole test to have min-version 8 as suricata 7 does not support prefilter on decode-event 

And this is also a good example/showcase of the `requires` keyword : you load the better rule if suricata version allows it

catenacyber:
Continued in https://github.com/OISF/suricata-verify/pull/2001

PR #1995, opened 2024-07-31
tests: add test for 7187

victorjulien:
Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7187


catenacyber:
Strange pcap, is not call_id supposed to be incremented by the client ?

inashivb:
> Strange pcap, is not call_id supposed to be incremented by the client ?

While that's what's mentioned in the RFC and that's how correct request/response pairs should be mapped, I can find several pcaps with repeated call IDs.. within s-v and outside. I don't think I understand this dcerpc behavior properly.

catenacyber (approved):
Test looks good to me

victorjulien:
Merged in #2036, thanks!

PR #1911, opened 2024-06-11
tests: add test to verify unmatched signatures does not write to dataset

norg:
Related to https://redmine.openinfosecfoundation.org/issues/5576

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5576

Requires the fix from https://github.com/OISF/suricata/pull/8123

PR #2000, opened 2024-08-01
tests: add test to verify unmatched signatures does not write to dataset

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5576

#1911 with adding a positive test, that data gets written to dataset when there is a full match


catenacyber:
Next in https://github.com/OISF/suricata-verify/pull/2093

PR #1991, opened 2024-07-24
pgsql: add tests with alert metadata, adjust pcap_cnt - v2

jufajardini:
Check for transaction metadata in PGSQL alerts.
Add `engine-analysis` tests for the used rules, as well, to better describe them and compare with expected behavior.

Related to
Task #7000

https://github.com/OISF/suricata-verify/pull/1990 without force-pushes

## Suricata PR

https://github.com/OISF/suricata/pull/11547

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7113
https://redmine.openinfosecfoundation.org/issues/7000

PR #2004, opened 2024-08-07
pgsql: add tests with alert metadata - v3

jufajardini:
Check for transaction metadata in PGSQL alerts.
Add `engine-analysis` tests for the used rules, as well, to better describe them and compare with expected behavior.

Related to
Task #7000

Previous PR #1991

### Suricata PR
https://github.com/OISF/suricata/pull/11618

## Ticket

changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7113
https://redmine.openinfosecfoundation.org/issues/7000


tests/pgsql/pgsql-7000-ips/README.md
```diff
@@ -0,0 +1,21 @@
...
+Showcase engine behavior with simple pgsql stream rules and alert metadata, in
+IPS mode.
+
+## Behavior
+
+The 2 rules will match on the PGSQL queries for `SELECT` (sid 1) and `DELETE FROM`,
+(sid 2). However, the combination of IPS mode matching on the traffic as soon as
+the engine sees it with the continuously available stream buffer leads to way
+more matches than could be initially expected. The `payload_printable` EVE field
+was left to illustrate this.
```
jufajardini, line 12: to be decided whether we want to have this test to fail, or keep it as is, and add a bug ticket to track this?

tests/rules/pgsql-7000/test.rules
```diff
@@ -0,0 +1,2 @@
+alert pgsql any any -> any any (msg:"Pgsql table select"; content:"select"; nocase; sid:1;)
+alert pgsql any any -> any any (msg:"Pgsql table delete"; content:"delete from"; nocase; sid:2;)
```
catenacyber, line 2: Why do we test these 2 rules ? They look similar

tests/pgsql/pgsql-7000-ips/test.yaml
```diff
@@ -0,0 +1,533 @@
...
+    count: 1
+    match:
+      alert.signature: Pgsql table delete
+      alert.signature_id: 2
+      app_proto: pgsql
+      direction: to_server
+      event_type: alert
+      payload_printable: ......./...O....user.ju-Test.database.Test.application_name.psql.client_encoding.UTF8..p...6SCRAM-SHA-256.... n,,n=,r=ROtF8e2Fme8+eORLNHTwkZaKp...lc=biws,r=ROtF8e2Fme8+eORLNHTwkZaKtpbEaXYJOnd3qt6QNCsAv0wj,p=I4V0zdtQqrxum6B+QzprHHC0nBD+mVtBWpc+arfXa+c=Q....BEGIN;.Q....DELETE FROM new_table WHERE NAME='Remus';.
+      pcap_cnt: 22
+      pgsql.request.simple_query: DELETE FROM new_table WHERE NAME='Londubat';
```
catenacyber, line 36: Why do we have `Londubat` in  pgsql.request.simple_query ? and `Remus` in `payload_printable` ?

tests/pgsql/pgsql-7000-ips/README.md
catenacyber: Why does it fail ?

jufajardini: Both for master and with the currently proposed changes we have 37 alerts in IPS mode, which is more than expected - in comparison, we have 8 alerts in IDS mode.
A good illustration of the problem that there is there is because we're getting alerts for the 2 rules (which inspect different contents) for the same packet.
From master to the proposed changes, one of the differences we have that makes the tests fail is the tx_id. 
But my question was more: should we have tests that register what's the expected behavior (the correct number of alerts), versus what we're actually seeing...

tests/rules/pgsql-7000/test.rules
jufajardini: Wanted to see different transactions being detected.

tests/pgsql/pgsql-7000-ips/test.yaml
jufajardini: This is part of the possible bug that I mentioned in the other comment: in IPS mode, possibly because the engine keeps a bigger stream window (I suppose?), we get alerts triggering in weird situations.

tests/rules/pgsql-7000/test.rules
jufajardini: (And then used the same rules from the test in this test)

jufajardini:
Replaced with https://github.com/OISF/suricata-verify/pull/2014

PR #2007, opened 2024-08-13
tests: showcase flow.action lack of update - v3

jufajardini:
It seems that in certain cases as seen in this test, flow.action isn't updated, even if, say, all packets from the flow are dropped.

Maybe this is due to the rule not being applied directly to the flow, but to each packet individually. But considering we are using a flow keyword, it seems that the engine should pass over the drop action to flow.action, at least in the flow event.

Bug #6976

Previous PR: #1987 

Updates:
- rebase
- keep checks for version 8, for now - once bug is fixed and we merge these tests, then we can update them, if we backport the fix.

## Ticket

changes to the ticket status:

Redmine ticket:

### DIFF
--- a/tests/flow-action-update-ips-01/README.md
+++ b/tests/flow-action-update-ips-01/README.md
@@ -1,0 +1,17 @@
+# Test
+
+This is a "fork" of `ips-state-1`, to showcase the `flow.action` update bug.
+
+## PCAP
+
+From `ips-state-1` test.
+
+This PCAP contains 3 flows. 2 are http and one is TLS. The HTTP flows should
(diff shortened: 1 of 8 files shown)


catenacyber:
Should we deprecate the master-6 builders now that Suricata 6 is EOL ?

jufajardini:
> Should we deprecate the master-6 builders now that Suricata 6 is EOL ?

Good question. Considering we don't have builders for the others, I'm guessing that makes sense... 🤔

jufajardini:
suricata 6 checks removed with: https://github.com/OISF/suricata-verify/pull/2052

PR #2009, opened 2024-08-13
github-ci: remove 6.0.x builds

jasonish:
Suricata 6.0 is now EOL.


jasonish:
Updated here, with even more removals: https://github.com/OISF/suricata-verify/pull/2013

PR #2011, opened 2024-08-14
http2: add test for frames

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5743

### DIFF
--- a/tests/eve-payload-07-http-gap/suricata.yaml
+++ b/tests/eve-payload-07-http-gap/suricata.yaml
@@ -7,6 +7,9 @@ outputs:
       filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
       filename: eve.json
       types:
+        - frame:
+            # this should not affect detection (but it did)
+            enabled: yes
         - alert:
             payload: yes             # enable dumping payload in Base64
             payload-buffer-size: 4kb # max size of payload buffer to output in eve-log
--- a/tests/http2-frames/test.rules
+++ b/tests/http2-frames/test.rules
@@ -1,0 +1,1 @@
+alert http2 any any -> any any (frame:http2.hdr; content:"|00 00 04 08 00|"; flow:to_server; sid:1;)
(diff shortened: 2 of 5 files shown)


tests/http2-frames/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+  min-version: 8
+
+# disables checksum verification
+args:
+  - -k none --set stream.midstream=true
+
+pcap:  ../http2-keywords2/input.pcap
+
+checks:
+# checks for http.uri keyword : 1 for HTTP1, 1 for mimicked HTTP2 response, so 2 for whole HTTP
```
inashivb, line 11: leftover from copy paste?

tests/http2-frames/test.rules
```diff
@@ -0,0 +1 @@
+alert http2 any any -> any any (frame:http2.hdr; content:"|00 00 04 08 00|"; flow:to_server; sid:1;)
```
inashivb, line 1: Why don't we test all frame types?

tests/eve-payload-07-http-gap/suricata.yaml
```diff
@@ -7,6 +7,9 @@ outputs:
       filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
       filename: eve.json
       types:
+        - frame:
+            # this should not affect detection (but it did)
```
inashivb, line 11: This really shouldn't. This is output config section..
Q: Is the fix really then https://github.com/OISF/suricata/pull/11717/commits/ec69ad0a7a6d1918839b00c75314af68aa013765 ?
It seems to me that stream frame should be enabled at all times. wdyt?

tests/http2-frames/test.yaml
catenacyber: indeed removing

tests/eve-payload-07-http-gap/suricata.yaml
catenacyber: > This really shouldn't. 

Indeed, it should not affect detection but it does now in master...
The fix is indeed https://github.com/OISF/suricata/commit/ec69ad0a7a6d1918839b00c75314af68aa013765

> It seems to me that stream frame should be enabled at all times. wdyt?

Suricata is smart enough that frames are truly created only if they get used in rules or output...
So, no rules with frame:http2.stream, and no output, means these frames do not get created...

tests/http2-frames/test.rules
catenacyber: :-p

Because I tested the one I needed

Adding the other ones

victorjulien (approved):

victorjulien:
Merged in #2036, thanks!

PR #2012, opened 2024-08-15
tests: Updates for 6555

jlucovsky:
This commit provides updates needed for issue 6555. Previously, the gap handling was restricted to master; 6555 adds those changes to main-7.0.x

Most of the changes are to extend the version; the eve-payload-07-http-gap tests adds version-based checks as a new output value payload_length is not available in main-7.0.x

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6155

### DIFF
--- a/tests/eve-payload-07-http-gap/test.yaml
+++ b/tests/eve-payload-07-http-gap/test.yaml
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  min-version: 7.0.7
 
 pcap: ../http-gap-beyond-body/input.pcap
 
@@ -14,6 +14,7 @@ checks:
       alert.signature_id: 1
 - filter:
     count: 1
+    min-version: 8
     match:
       event_type: alert
       alert.signature_id: 1
@@ -21,6 +22,14 @@ checks:
       payload_length: 40
 - filter:
     count: 1
+    min-version: 7.0.7
+    match:
+      event_type: alert
+      alert.signature_id: 1
+      payload_printable: "GET /1 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\n"
+- filter:
+    count: 1
+    min-version: 8
     match:
       event_type: alert
       alert.signature_id: 1
@@ -28,6 +37,14 @@ checks:
       payload_length: 80
 - filter:
     count: 1
+    min-version: 7.0.7
+    match:
+      event_type: alert
+      alert.signature_id: 1
+      payload_printable: "GET /1 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\nGET /2 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\n"
+- filter:
+    count: 1
+    min-version: 8
     match:
       event_type: alert
       alert.signature_id: 1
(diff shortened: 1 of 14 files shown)


tests/eve-payload-07-http-gap/test.yaml
```diff
@@ -14,39 +14,84 @@ checks:
...
+    version: 8.0
     match:
       event_type: alert
       alert.signature_id: 1
       payload_printable: "GET /1 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\n"
       payload_length: 40
 - filter:
     count: 1
+    version: 7.0.7
+    lt-version: 8.0
```
catenacyber, line 26: You do not need  lt-version: 8.0 here, do you ?

tests/eve-payload-07-http-gap/test.yaml
```diff
@@ -14,39 +14,84 @@ checks:
       alert.signature_id: 1
 - filter:
     count: 1
+    version: 8.0
```
jasonish, line 17: Should this be `min-version`?

tests/eve-payload-07-http-gap/test.yaml
```diff
@@ -14,39 +14,84 @@ checks:
...
     match:
       event_type: alert
       alert.signature_id: 1
       payload_printable: "GET /1 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\n"
       payload_length: 40
 - filter:
     count: 1
+    version: 7.0.7
```
jasonish, line 25: `min-version`?  `version` here would be for only `7.0.7`.

tests/eve-payload-07-http-gap/test.yaml
```diff
@@ -14,39 +14,84 @@ checks:
...
+    count: 1
+    version: 8.0
     match:
       event_type: alert
       alert.signature_id: 1
       payload_printable: "GET /1 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\nGET /2 HTTP/1.0\r\nUser-Agent: Mozilla\r\n\r\n"
       payload_length: 80
 - filter:
     count: 1
+    version: 7.0.7
```
jasonish, line 41: More of these.

jasonish (changes requested):
See comments inline.

tests/eve-payload-07-http-gap/test.yaml
jlucovsky: Yes -- this check if for main-7.0.x with the backport so [7.0.7, 8.0)

tests/eve-payload-07-http-gap/test.yaml
jlucovsky: Yes; I'll update everywhere.

tests/eve-payload-07-http-gap/test.yaml
jlucovsky: The test for 8.0 checks for the recently-added `payload_length` (which isn't in main-7.0.x)

catenacyber: So, this check also works for 8, you do not need lt-version, right ?

jlucovsky: That's true but there is a filter check immediately preceding this one that is targeted for 8+

jlucovsky: Will update.

victorjulien:
Please open a new PR, this one is a bit messy now.

jlucovsky:
Continued in #2038

PR #2017, opened 2024-08-23
Flow bytes pkts either/v1

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5646

### DIFF
--- a/tests/detect-flow-pkts-either/test.rules
+++ b/tests/detect-flow-pkts-either/test.rules
@@ -1,0 +1,1 @@
+alert ip any any -> any any (msg:"Flow has 1000 bytes in either direction";flow.pkts:either,=10;flow.bytes:either,>3000; sid:1;)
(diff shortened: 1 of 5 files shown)


tests/detect-flow-pkts-either/test.rules
```diff
@@ -0,0 +1 @@
+alert ip any any -> any any (msg:"Flow has 1000 bytes in either direction";flow.pkts:either,=10;flow.bytes:either,>3000; sid:1;)
```
catenacyber, line 1: > Flow has 1000 bytes in either direction

does not seem to match

> flow.pkts:either,=10;flow.bytes:either,>3000

tests/detect-flow-pkts/test.rules
```diff
@@ -1 +1 @@
-alert ip any any -> any any (msg:"Flow has 10 packets"; flow.pkts_toclient:10; flow.pkts_toserver:10; flow.bytes_toserver:<3000; flow.bytes_toclient:>8000; sid:1;)
+alert ip any any -> any any (msg:"Flow has 10 packets"; flow.pkts:toclient,=10; flow.pkts:toserver,=10; flow.bytes:toserver,<3000; flow.bytes:toclient,>8000; sid:1;)
```
catenacyber, line 1: Do you need the `=` in `=10` ?

inashivb: As far as I understood. This mode will work with or w/o the `=`. I'll remove the `=` for simplicity. Thank you. :)

tests/detect-flow-pkts-either/test.rules
inashivb: Thank you. Will update the msg

PR #2013, opened 2024-08-15
remove 6.0/v2

jasonish:
- **github-ci: remove 6.0.x builds**
- **tests: remove suricata 6 specific tests**
- **tests: remove tests for lt-version: 7**

PR #2018, opened 2024-08-25
Remove 6.0/v2.1/pcaps

catenacyber:
## Ticket

Redmine ticket: follow up on #2013 to move pcaps to directories where they actually get used after removal of tests only used by version 6

cc @jasonish should we do this for the 5 other dns tests ?


jasonish:
This doesn't fix all of them?

catenacyber:
>> cc jasonish should we do this for the 5 other dns tests ?
> This doesn't move all of them?

I take that as a yes, see https://github.com/OISF/suricata-verify/pull/2019

PR #2021, opened 2024-08-28
Coledishington/flow force reuse 6293 v7

catenacyber:

Redmine ticket:

https://github.com/OISF/suricata-verify/pull/1699 with additional commit to run only on arch x86_64


tests/flow-force-reuse/README.md
```diff
@@ -0,0 +1,9 @@
+# Test Purpose
+
+Tests Suricata with the configuration option flow.force-reuse on and
+off. This is done by having a low flow memcap and long-lived
+connections.
+
+## PCAP
+
+This PCAP was generated with scapy.
```
inashivb, line 9: Could we have the ticket number here?

tests/flow-force-reuse/flow-force-reuse-off/test.yaml
```diff
@@ -0,0 +1,20 @@
+requires:
+  min-version: 8
+  arch: x86_64
+
+pcap: ../test.pcap
+
+args:
+- --set flow.memcap-policy=drop-flow
```
inashivb, line 8: does this matter in this case?

inashivb (commented):
Looking good. Minor comments inline

tests/flow-force-reuse/README.md
catenacyber: added

catenacyber:
@coledishington closing here as Suricata PR, waiting for your PR

PR #2034, opened 2024-09-11
tests: add test for issue 7241/7242 for 7

victorjulien:
Add test that works with Suricata 7.

### DIFF
--- a/tests/bug-7241-02-pre8/test.rules
+++ b/tests/bug-7241-02-pre8/test.rules
@@ -1,0 +1,1 @@
+drop tcp any any -> any any (flow:established; app-layer-protocol:!tls; sid:1;)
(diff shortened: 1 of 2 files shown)


catenacyber (approved):
Ok for me : bug-7241-01-8plus cannot work for 7 even with the suricata backport because `app-layer-protocol:!tls,either` with the `either` is a new feature in 8 ( and also maybe prefilter for app-layer-protocol)

tests/bug-7241-02-pre8/test.rules
```diff
@@ -0,0 +1 @@
+drop tcp any any -> any any (flow:established; app-layer-protocol:!tls; sid:1;)
```
catenacyber, line 1: nit : could add a README.md, especially for explaining why we have 2 different tests for 7 and 8

catenacyber:
Oh, I guess we should test prefilter as well cf https://github.com/OISF/suricata/pull/11750/files#r1753510437

victorjulien:
replaced by #2035

PR #2035, opened 2024-09-11
tests: add test for issue 7241/7242 for 7

victorjulien:
Add test that works with Suricata 7.

Replace #2034 to add a prefilter rule as well.

### DIFF
--- a/tests/bug-7241-02-pre8/test.yaml
+++ b/tests/bug-7241-02-pre8/test.yaml
@@ -1,0 +1,20 @@
+requires:
+  min-version: 7
+
+pcap: ../tls/tls-random/input.pcap
+
+args:
+- -k none
+
+checks:
+- filter:
+    count: 0
+    match:
+      alert.signature_id: 1
+      event_type: alert
+checks:
+- filter:
+    count: 0
+    match:
+      alert.signature_id: 2
+      event_type: alert
--- a/tests/bug-7241-02-pre8/README.md
+++ b/tests/bug-7241-02-pre8/README.md
@@ -1,0 +1,4 @@
+Test for issue 7272
+===================
+
+Test specifically for the features supported in Suricata 7.
--- a/tests/bug-7241-02-pre8/test.rules
+++ b/tests/bug-7241-02-pre8/test.rules
@@ -1,0 +1,2 @@
+drop tcp any any -> any any (flow:established; app-layer-protocol:!tls; sid:1;)
+drop tcp any any -> any any (flow:established; app-layer-protocol:!tls; prefilter; sid:2;)


catenacyber (approved):
This looks good to me.

How do you test the Suricata patch to use `AppProtoEquals` ?

victorjulien:
> This looks good to me.
> 
> How do you test the Suricata patch to use `AppProtoEquals` ?

I suppose it can really only be tested with profiling, where you'd check prefilter results or something.

catenacyber:
Why is CI green here ? 🤔

victorjulien:
> Why is CI green here ? 🤔

That is odd... did adding the 2nd rule to the test break the first?

jufajardini:
> > This looks good to me.
> > How do you test the Suricata patch to use `AppProtoEquals` ?
> 
> I suppose it can really only be tested with profiling, where you'd check prefilter results or something.

Would this be useful? https://github.com/OISF/suricata-verify/commit/aa4883cca55c7491262446efa577216fd1a274c5

victorjulien:
> > > This looks good to me.
> > > How do you test the Suricata patch to use `AppProtoEquals` ?
> > 
> > 
> > I suppose it can really only be tested with profiling, where you'd check prefilter results or something.
> 
> Would this be useful? [aa4883c](https://github.com/OISF/suricata-verify/commit/aa4883cca55c7491262446efa577216fd1a274c5)



> > > This looks good to me.
> > > How do you test the Suricata patch to use `AppProtoEquals` ?
> > 
> > 
> > I suppose it can really only be tested with profiling, where you'd check prefilter results or something.
> 
> Would this be useful? [aa4883c](https://github.com/OISF/suricata-verify/commit/aa4883cca55c7491262446efa577216fd1a274c5)

Yes! Do you see it fail on `main-7.0.x`?

inashivb:
Do we consider this test correct then? This is staged for merging but the conversation above indicates this test may be exposing a new issue and not showing the current one (which should have led to failure).

victorjulien:
The issue it exposes is not related to the PRs.

tests/bug-7241-02-pre8/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+args:
+- -k none
+
+checks:
+- filter:
+    count: 0
+    match:
+      alert.signature_id: 1
+      event_type: alert
+checks:
```
inashivb, line 15: This test passes because this is wrong syntax for `test.yaml`..
Here, we're checking for an event that matches a checks type dictionary in the yaml alongwith the attributes mentioned above. Since it does not exist, the count for such an event is 0..

Found when I was building upon this test and saw inconsistencies in non negation behavior.
1. Non negation behavior is just fine. My analysis was incorrect because I was building upon an incorrect test.
2. Negation behavior is buggy as shown by the ticket. This test does and should fail ideally on `main-7.0.x`.

Sending a PR fixing this..

jufajardini:
> > > > This looks good to me.
> > > > How do you test the Suricata patch to use `AppProtoEquals` ?
> > > 
> > > 
> > > I suppose it can really only be tested with profiling, where you'd check prefilter results or something.
> > 
> > 
> > Would this be useful? [aa4883c](https://github.com/OISF/suricata-verify/commit/aa4883cca55c7491262446efa577216fd1a274c5)
> 
> > > > This looks good to me.
> > > > How do you test the Suricata patch to use `AppProtoEquals` ?
> > > 
> > > 
> > > I suppose it can really only be tested with profiling, where you'd check prefilter results or something.
> > 
> > 
> > Would this be useful? [aa4883c](https://github.com/OISF/suricata-verify/commit/aa4883cca55c7491262446efa577216fd1a274c5)
> 
> Yes! Do you see it fail on `main-7.0.x`?

The engine-analysis tests don't fail for `main-7.0.x`. There isn't enough info in the output, I guess, to generate any variation...

victorjulien:
closing in favor of #2043

tests/bug-7241-02-pre8/test.yaml
catenacyber: ❤️ yaml

PR #2025, opened 2024-08-30
pgsql: add tests with alert metadata - v5

jufajardini:
Check for transaction metadata in PGSQL alerts.
Add `engine-analysis` tests for the used rules, as well, to better describe them and compare with expected behavior.

Related to
Task #7000

Suricata PR: https://github.com/OISF/suricata/pull/11635

Previous PR: https://github.com/OISF/suricata-verify/pull/2014

Updates:
- Rebase
- deduplicate pcap used for `pgsql-7000-ids`

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket:
- https://redmine.openinfosecfoundation.org/issues/7000
- https://redmine.openinfosecfoundation.org/issues/7113

PR #2039, opened 2024-09-13
pgsql: add tests with alert metadata - v6

jufajardini:
Check for transaction metadata in PGSQL alerts.
Add engine-analysis tests for the used rules, as well, to better describe them and compare with expected behavior.

Related to
Task #7000

Suricata PR: https://github.com/OISF/suricata/pull/11776

Previous PR: https://github.com/OISF/suricata-verify/pull/2025

### Updates:
- Rebase
- add a disclaimer to the pgsql rules, as they're not good real-life rule examples
- Update `pgsql-bug-6983-ips` test as one of the checks would now fail due to better transaction completion tracking (the per-direction factor means that when we log the first alert, in ips, now, we don't have the response part of the transaction available)

### Expectation:
- all tests should pass with master, except `pgsql-bug-6983-ips`. The other tests pass because, as far as I understand, tracking tx completion per direction results in a similar behavior to what we used to see with where we were previously triggering raw stream reassembly.

Ticket

changes to the ticket status:

Redmine ticket:
- https://redmine.openinfosecfoundation.org/issues/7113
- https://redmine.openinfosecfoundation.org/issues/7000

### DIFF
--- a/tests/pgsql/pgsql-bug-6983-ips/test.yaml
+++ b/tests/pgsql/pgsql-bug-6983-ips/test.yaml
@@ -13,15 +13,26 @@ checks:
     match:
       event_type: pgsql
 - filter:
-    # in ips mode, as this rule inspects the stream only (no pgsql keywords), we end up getting two alerts instead of one
+    # in ips mode, as this rule inspects the stream only (no pgsql keywords),
+    # we end up getting two alerts instead of one
     count: 2
     match:
       event_type: alert
       alert.signature_id: 1
 - filter:
+    lt-version: 8
     count: 1
     match:
       event_type: alert
       alert.signature_id: 1
       pgsql.request.simple_query: "select * from rules where sid = 2021701;"
       pgsql.response.field_count: 10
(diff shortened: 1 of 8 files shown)


tests/pgsql/pgsql-bug-6983-ips/test.yaml
```diff
@@ -13,15 +13,26 @@ checks:
...
       pgsql.response.field_count: 10
+- filter:
+    # Once pgsql's completed transactions are tracked per-direction, we don't have the
+    # response part of the transaction when we get the alert, in IPS mode
+    min-version: 8
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 1
+      pgsql.request.simple_query: "select * from rules where sid = 2021701;"
```
catenacyber, line 38: I propose we keep simply this check without  pgsql.response.field_count for all versions

jufajardini:
followed by: https://github.com/OISF/suricata-verify/pull/2048

PR #2019, opened 2024-08-26
tests: move pcaps to tests where they are used

catenacyber:
Redmine ticket: follow up on https://github.com/OISF/suricata-verify/pull/2013 to move pcaps to directories where they actually get used after removal of tests only used by version 6

cc @jasonish

#2018 with all tests done

PR #1968, opened 2024-07-08
README/help: add info on IPS mode tests creation - v3

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/1965

Describe changes:
- it's not a straightforward process to use the test name alternative with createst, so I've removed that association from the README and createst help text.
- remove mention to unrelated ticket from 540b1a5a6fad675bebc85efdfb3b0a1b1dae141b

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7039

PR #2046, opened 2024-09-18
next/sv/578/20240918/v1

victorjulien:
Staging:
 - #2037


catenacyber:
Could we bundle in more PRs like #1968 and #2019 ?
I mean SV PRs that are orphan of a Suricata PR, where tests pass and that have been approved

victorjulien:
> Could we bundle in more PRs like #1968 and #2019 ? I mean SV PRs that are orphan of a Suricata PR, where tests pass and that have been approved

Not as part of the regular merge workflow. That will require a separate run.

jufajardini (approved):
Original PR is approved;
number of files and lines changed matches;
number of commits matches.

catenacyber:
> Not as part of the regular merge workflow. That will require a separate run.

Ok, Even #1968 which is doc only ?

inashivb:
> > Not as part of the regular merge workflow. That will require a separate run.
> 
> Ok, Even #1968 which is doc only ?

These next PRs I believe automagically pick up only the s-v PRs that are mentioned in the Suricata PRs that have been staged for merge. So, yeah it'll have to be separate even if it is doc only :)

PR #2047, opened 2024-09-19
Verify bypass behavior under different traffic profiles

lukashino:
Redmine ticket:

Note:
- min-version changes between tests between 7 and 8, the ones with min-version == 7 is where behavior is not expected to change, the others are where the behavior changes (TLS bypass decoupling, SSH bypass).


catenacyber:
@lukashino you should convert the pcapng to a regular pcap file (so that it works on all devices) ;-)

lukashino:
goto https://github.com/OISF/suricata-verify/pull/2078

PR #2048, opened 2024-09-19
pgsql: add tests with alert metadata - v7

jufajardini:
Previous PR: https://github.com/OISF/suricata-verify/pull/2039

Updates:
- Rebase
- `pgsql-bug-6983` - remove the check for response portion of transaction which isn't always present in alerts, so checks can be more Suricata-version agnostic

Expectation:
- tests should pass because, as far as I understand, tracking tx completion per direction in IDS mode results in a similar behavior to what we used to see with where we were previously triggering raw stream reassembly - just done more accurately, now

## Ticket

changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7113

### DIFF
--- a/tests/pgsql/pgsql-7000-ids/README.md
+++ b/tests/pgsql/pgsql-7000-ids/README.md
@@ -1,0 +1,18 @@
+# Description
+
+Showcase engine behavior with simple pgsql stream rules and alert metadata.
+
+## Behavior
+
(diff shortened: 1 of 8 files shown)


victorjulien (approved):

victorjulien:
Merged in #2050, thanks!

PR #2010, opened 2024-08-14
smtp: add tests for issue 7126

jlucovsky:
Ensure the SMTP applayer parser doesn't generate an error message while parsing the SMTP frames.

Suricata PR: https://github.com/OISF/suricata/pull/11630

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7126

PR #2005, opened 2024-08-10
sip: add tests for headers sticky buffers v4

glongo:
Ticket #6374

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6374

PR #2053, opened 2024-09-22
next/sv/582/20240922/v1

victorjulien:
Staging:
 - #2010
 - #2005


catenacyber (approved):
Consistent with the 2 approved PRs

catenacyber:
nit: I think we should remove tests/bug-7126/input.rules or add a check about it

PR #2067, opened 2024-09-27
mysql: add tests

QianKaiLin:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3446


tests/mysql-command/test.yaml
```diff
@@ -0,0 +1,11 @@
+args:
```
catenacyber, line 1: Could you please add min-version: 8 in the requires section as to have CI green for main7 branch ?

tests/mysql-command/test.yaml
```diff
@@ -0,0 +1,11 @@
...
+- -k none
+
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature: "test mysql"
+
```
catenacyber, line 11: Could we also test that we have mysql metadata in the signature ?

tests/mysql-command/test.rules
```diff
@@ -0,0 +1 @@
+alert mysql any any -> any any (msg:"test mysql";mysql.command; pcre:"/(?:[1-9]\d{5})(?:(?:1[89]\d{2}|2\d{3})(?:0[1-9]|1[012])(?:0[1-9]|[12][0-9]|3[01]))\d{2}(?:\d)(?:[0-9xX])/i""; sid:1;)
```
catenacyber, line 1: Why use a PCRE and not a fixed pattern ?

tests/mysql-rows/test.rules
```diff
@@ -0,0 +1 @@
+alert mysql any any -> any any (msg:"test mysql";mysql.rows; pcre:"/(?:[1-9]\d{5})(?:(?:1[89]\d{2}|2\d{3})(?:0[1-9]|1[012])(?:0[1-9]|[12][0-9]|3[01]))\d{2}(?:\d)(?:[0-9xX])/i""; sid:1;)
```
catenacyber, line 1: What is the difference with the test mysql-command ? Could we have a README also ?

tests/mysql-prepare-statement/test.yaml
```diff
@@ -0,0 +1,73 @@
...
+args:
+- -k none
+
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: mysql
+      mysql.command: select * from requests WHERE id =1 limit 1
+      mysql.tls: false
```
catenacyber, line 13: Would you have a test with a pcap that does TLS upgrade ?

jufajardini (changes requested):
Tests `mysql-query` and `mysql-prepare-statement` are missing the suricata.yaml file. As the protocol isn't enabled by default, without having a suricata.yaml in the test and enabling the protocol in the eve logs, the test will fail here - this also applies to the tests that do have the suricata.yaml, but are not enabling `mysql` event logs.

Could you add checks that showcase the log output, for us?

With new log additions, this is important for us not only to check that they look like expected, but also to discuss if the format needs changes :)

tests/mysql-command/test.rules
QianKaiLin: Because this signature is from my project, and I want to match the China identify card.

tests/mysql-rows/test.rules
QianKaiLin: Ok, i will add.

tests/mysql-command/test.yaml
QianKaiLin: `alert mysql any any -> any any (msg:"test mysql";mysql.command; content:"33030219971120201X"; metadata: mysql command; sid:1;)`,Did you mean this?

QianKaiLin:
> Could you add checks that showcase the log output, for us?
> 
> With new log additions, this is important for us not only to check that they look like expected, but also to discuss if the format needs changes :)

I'm sorry I don't get it.

tests/mysql-command/test.yaml
jufajardini: I think that what he means is something like what is seen in [this test](https://github.com/OISF/suricata-verify/blob/master/tests/pgsql/pgsql-7000-ids/test.yaml#L12-L24), in which we check that the alert event contains app-layer metadata. In the case here, `mysql` app-layer metadata.

jufajardini:
> > Could you add checks that showcase the log output, for us?
> > With new log additions, this is important for us not only to check that they look like expected, but also to discuss if the format needs changes :)
> 
> I'm sorry I don't get it.

I meant something like what is seen here: https://github.com/OISF/suricata-verify/blob/master/tests/pgsql/pgsql-simple-query-rollback/test.yaml

QianKaiLin:
> > > Could you add checks that showcase the log output, for us?
> > > With new log additions, this is important for us not only to check that they look like expected, but also to discuss if the format needs changes :)
> > 
> > 
> > I'm sorry I don't get it.
> 
> I meant something like what is seen here: https://github.com/OISF/suricata-verify/blob/master/tests/pgsql/pgsql-simple-query-rollback/test.yaml

I have added, please check again, thx.

PR #2065, opened 2024-09-26
detect/datasets: adds test for unset operation

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7195

Suricata feature was merged for https://github.com/OISF/suricata/pull/11815 and in 7 also with https://github.com/OISF/suricata/pull/11833

PR #2070, opened 2024-09-29
detect/datasets: adds test for unset operation

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7195

Suricata feature was merged for https://github.com/OISF/suricata/pull/11815 and in 7 also with https://github.com/OISF/suricata/pull/11833

https://github.com/OISF/suricata-verify/pull/2065 new version with custom made pcap and rules only using `save` instead of `load` and `save`


inashivb (approved):
LGTM 🚀

inashivb (changes requested):
Wait a second. Everything looks good but I think because there is no pcap, this test will simply be skipped so although it's nice you've added how you tested it but this test is of no value to our suite if we don't have a pcap..

catenacyber:
> Wait a second. Everything looks good but I think because there is no pcap, this test will simply be skipped so although it's nice you've added how you tested it but this test is of no value to our suite if we don't have a pcap..

Oops, added

inashivb:
> > Wait a second. Everything looks good but I think because there is no pcap, this test will simply be skipped so although it's nice you've added how you tested it but this test is of no value to our suite if we don't have a pcap..
> 
> Oops, added

Thanks. Looking perfect now. :)
Will you make another PR w a single commit?

catenacyber:
Done in https://github.com/OISF/suricata-verify/pull/2071

PR #2073, opened 2024-10-02
tests: showcase bug 7286 (tls) - v2

jufajardini:
Related to
Bug https://redmine.openinfosecfoundation.org/issues/7286

This also exposes https://redmine.openinfosecfoundation.org/issues/7287 and thus requires  that https://github.com/OISF/suricata/pull/11844 is merged, not to break SV for master

Describe changes:
- https://github.com/OISF/suricata-verify/pull/2068 rebased and with `min-version: 8` for CI checks for 7 to pass

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7286

PR #2068, opened 2024-09-27
tests: showcase bug 7286 (tls) - v1

jufajardini:
This also exposes https://redmine.openinfosecfoundation.org/issues/7287.

Related to
Bug https://redmine.openinfosecfoundation.org/issues/7286

To accompany https://github.com/OISF/suricata/pull/11843

Expectation:
- tests should fail not only due to the reported bug, but also because `certificate` and `chain` were absent from `schema.json`

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7286

PR #2074, opened 2024-10-03
Skip schema validation/v1

inashivb:
Just a suggestion following up on https://github.com/OISF/suricata-verify/pull/2068#discussion_r1780483581
I've taken up #2073 and added the feature there to show how it works. (Note that this feature should actually just be used for older version tests that cannot get that schema update)

This is for cases where because of broken schema, a test would just always fail. But, we can be lenient about the schema and ensure that the output actually matched in these cases.

Thoughts?


catenacyber:
Looks weird to me : I think the test fails.
You can add some output to describe the reason of the failure (or failures plural) if you want to post process these and allow tests that fail only because of schema.

You can also skip the schema validation by not having the eve-validator somehow, right ?

jufajardini:
> Looks weird to me : I think the test fails. You can add some output to describe the reason of the failure (or failures plural) if you want to post process these and allow tests that fail only because of schema.
> 

This is an interesting idea, too.
Set default to JSON schema failure not to be fatal, just issue a warning, and test runs normally.
But still have at least one CI check with schema validation as fatal, so these failures would not slip by.

inashivb:
> Looks weird to me : I think the test fails. You can add some output to describe the reason of the failure (or failures plural) if you want to post process these and allow tests that fail only because of schema.
> 

Noice and better. Thanks a lot! 🙇🏽‍♀️ 

> You can also skip the schema validation by not having the eve-validator somehow, right ?

hmm only found yesterday that there's a command line option to do this. So, apparently, given that we anyway won't check old versions in the CI, one could easily run s-v w `--no-validation` on old versions.

So, all in all, moving forward, if we want, we should have multiple failure reasons logged. and, that's about it.

jufajardini:
> > You can also skip the schema validation by not having the eve-validator somehow, right ?
> 
> hmm only found yesterday that there's a command line option to do this. So, apparently, given that we anyway won't check old versions in the CI, one could easily run s-v w `--no-validation` on old versions.
> 

TIL (or re-learned) about the `--no-validation` option. I'll use that for one of my PRs, I guess.

inashivb:
> TIL (or re-learned) about the `--no-validation` option. I'll use that for one of my PRs, I guess.

yeah it seems helpful. Should have guessed its availability since Jason worked on it too. He is generally very mindful of common required options for usability. :)

jufajardini:
> > TIL (or re-learned) about the `--no-validation` option. I'll use that for one of my PRs, I guess.
> 
> yeah it seems helpful. Should have guessed its availability since Jason worked on it too. He is generally very mindful of common required options for usability. :)

:P
Although that option didn't work as I expected it to xD

inashivb:
> > > TIL (or re-learned) about the `--no-validation` option. I'll use that for one of my PRs, I guess.
> > 
> > 
> > yeah it seems helpful. Should have guessed its availability since Jason worked on it too. He is generally very mindful of common required options for usability. :)
> 
> :P Although that option didn't work as I expected it to xD

oh. Then, maybe we should indeed have ticket for that

PR #2078, opened 2024-10-06
Verify bypass behavior under different traffic profiles

lukashino:
Follow-up of https://github.com/OISF/suricata-verify/pull/2047

Redmine ticket:

Note:


lukashino:
Continues in https://github.com/OISF/suricata-verify/pull/2114

PR #2080, opened 2024-10-07
add test for extra tls alert

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7318


tests/tls-extra-alert/README.md
```diff
@@ -0,0 +1,12 @@
+# Test Description
+
+This test shows that Suricata generates an additional alert for TLS
+for the given PCAP which shouldn't be there.
```
catenacyber, line 4: For which rule ?

inashivb: For 9901033 as per test.yaml: `alert.signature_id: 9901033`

catenacyber: At which packet does it happen ? The pseudo flush ?

catenacyber: Or packet 8 and 9 ?

inashivb: > At which packet does it happen ? The pseudo flush ?

The extra alert happens at pseudo flush. Yes. The test filter that indicates that i test.yaml is:
```yaml
- filter:
    count: 0
    match:
      event_type: alert
      not-has-key: pcap_cnt
```

Note that the event generated w the pseudo pkt misses `pcap_cnt` key and also has `pkt_src: stream (flow timeout)` (so subtest 1 in the test also helps making sure of that)

catenacyber: > The extra alert happens at pseudo flush. Yes.

So, it looks expected. Why would you not want this behavior ?
You should add a keyword in your rule that sets `SIG_FLAG_REQUIRE_PACKET` if you do not want it, right ?

inashivb: > So, it looks expected. Why would you not want this behavior ? 

Because ideally there should be nothing leading to the generation of that pseudo pkt from what I understood.
@jasonish JFYI, Philippe is also participating here in the discussion. ^^

catenacyber: How so ?

Rules keywords are only `flow` and `flowbits`, so it should match on both regular and pseudo packets

You should add a keyword in your rule that sets SIG_FLAG_REQUIRE_PACKET

jasonish: My thoughts are that it should not fire as-is. Its somewhat non-sensical, at least its being logged relative to a packet (start of flow) where it should not be. Its only because of information found in the future, and ends up in the pseudo-packet that we end up alerting. Or at least its confusing.

As we're only pushing this pseudo-packet through to wrap up transactions, my thought was to skip detection on non-transaction data, in particular packet rules per the discussion here: https://github.com/OISF/suricata/pull/11862. But maybe this is expected behavior and there is no issue. I think we're still discussing the behavior and if this is correct or not.

catenacyber: > at least its being logged relative to a packet (start of flow)

How so with pkt_src: stream (flow timeout) ? 

> information found in the future

Do we time travel ?

> As we're only pushing this pseudo-packet through to wrap up transactions,

Beware there are timeouts pseudo packets, but other one likes HTTP1->HTTP2 upgrade...
IIRC, there was one ticket where we needed the pseudo packet to match on the content of the last SSH packet before going encrypted cf https://redmine.openinfosecfoundation.org/issues/6578 (and one similar for frames)

jasonish: > Do we time travel ?

We do :)

In the log we have an alert at `2024-09-09T12:15:56.542526-0600`, followed by an alert from `2024-09-09T12:15:55.863332-0600`.

That second timestamp matches the first packet in the pcap, however hits on the rule:
```
alert tcp any any -> $EXTERNAL_NET 443 (flow: to_server; flowbits:isset, tls_error; sid:09901033; rev:1; msg:"Allow TLS error handling (outgoing packet)"; )
```
and is logged with flowbits, even detected as tls, however we don't know of that at `2024-09-09T12:15:55.863332-0600`, unless we time travel to the future :)

Kidding aside, that's why this alert looks so out of place.

jasonish: > How so with pkt_src: stream (flow timeout) ?

Flow timeout also seems odd, as it didn't timeout. Its cleanly closed.

catenacyber: So, there is a bug in the timestamp of timeout packets, right ?

> Flow timeout also seems odd, as it didn't timeout. It's cleanly closed.

I think we just time out cleanly closed flows after sometime checking there is no session reuse...

jasonish:
Ticket created: https://redmine.openinfosecfoundation.org/issues/7318

inashivb:
There's a new rebased version linked above.

PR #2084, opened 2024-10-08
update tests w no payload to reflect it

inashivb:
This test seems to be the correct behavior to me. Thoughts?


inashivb:
The flag is supposed to be for packet not requiring payload as per Victor and not about signatures.

PR #2093, opened 2024-10-15
Dataset set postmatch 5576 v4

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5576

#2000 with more tests for complex cases

Do I get right the expected behavior of datasets-multibuf-postmatch ?
Rule `alert http any any -> any any (msg:"HTTP learning"; http.request_header; content:"toto"; dataset:set,http_match,type string,save http_match.csv; sid:1;)`
will save in dataset only the headers having the content `toto` and (not the user agent curl for instance)


jufajardini (commented):
Tests look good to me :)

nit: https://github.com/OISF/suricata-verify/pull/2093/commits/08f412d6f740667b344264e42fed4479f84fa29a

has a typo:
:s/retreieves/retrieves

PR #2102, opened 2024-10-20
mqtt: add additional tests for reason codes

satta:
This PR adds some more checks for situations related to https://forum.suricata.io/t/question-about-mqtt-detection/4890/ among others.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7323


jufajardini:
Current Suricata PR: https://github.com/OISF/suricata/pull/11995

tests/mqtt-sub-rules/test.yaml
```diff
@@ -109,3 +119,9 @@ checks:
       match:
         event_type: alert
         alert.signature: MQTT SUBSCRIBE topicY
+
+  - filter:
```
jufajardini, line 123: This one I think we'll have to add a `min-version` check for 8, before this can be merged.

jufajardini (changes requested):
Thanks for this work :)

tests/mqtt-sub-rules/test.yaml
satta: Oh, I didn't know you could also have `min-version` on individual filters. Sure, makes sense.

satta:
Next PR: #2106

PR #2114, opened 2024-11-03
Verify bypass behavior under different traffic profiles v3

lukashino:
Follow-up of https://github.com/OISF/suricata-verify/pull/2078

Redmine ticket:

Note:

Describe changes:
v3:
- rebased


tests/ssh-hassh/test.yaml
```diff
@@ -4,7 +4,7 @@ features:
     - RUST
 
 args:
- - -k none --set stream.bypass=yes
+ - -k none --set stream.bypass=yes --set app-layer.protocols.ssh.encryption-handling=bypass
```
catenacyber, line 7: Why do we need this change ?

tests/bypass-tls-enabled/test.yaml
```diff
@@ -0,0 +1,18 @@
+requires:
+    min-version: 8
+
```
catenacyber, line 3: Is it reusing a pcap ?

tests/ssh-hassh/test.yaml
lukashino: behavior change

- Encrypted SSH traffic bypass is now independently controlled through
  `app-layer.protocols.ssh.encryption-handling` setting. The setting can either
  be `bypass`, `default` or `full`. 
  To retain the previous behavior of encrypted traffic bypass
  combined with stream depth bypass, set
  `app-layer.protocols.ssh.encryption-handling` to `bypass` (while also
  setting `app-layer.protocols.tls.encryption-handling` to `bypass` and
  `stream.bypass` to `true`).

Previously SSH was bypassed after the session turned encrypted, now we need to be more explicit.

lukashino:
cont in https://github.com/OISF/suricata-verify/pull/2170

P's comment is mentioned in the PR

tests/ssh-hassh/test.yaml
lukashino: Re 
> Previously SSH was bypassed after the session turned encrypted, now we need to be more explicit.

So previously when the SSH session turned encrypted and stream bypass was allowed then it bypassed the flow. This was not related to reaching the stream depth but only to the session state. 
Currently, you can replicate the behavior with setting encryption handling to bypass.

PR #2119, opened 2024-11-05
datasets: add test to show hash collisions

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7209

### DIFF
--- a/tests/dataset-hash-collisions/README.md
+++ b/tests/dataset-hash-collisions/README.md
@@ -1,0 +1,21 @@
+Test Description
+================
+
+Datasets use a static DJB2 hash function to hash all types of datasets. These hashed
+datasets are stored in the THash API which has no randomization in place. As
+a result of this, the hash table can be exploited with a worst case time scenario of
+O(n) where n is the total number of entries in the table as a result of excessive chaining
+in a single row.
+
+The test shows that it takes excess time for the THash API to load the datasets from the file
+as many of them evaluate the exact same hash using the algorithm so this is not even the worst
+case scenario. With bigger dataset and lesser system specs/availability of resources,
+this can be worse. Note that it is not just about the number of datasets as there already
+does exist a test already that loads 1m+ datasets.
+
+Test data procured from: https://bugs.php.net/bug.php?id=70644
+
+Redmine Ticket
+==============
+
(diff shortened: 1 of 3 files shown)


tests/dataset-hash-collisions/README.md
```diff
@@ -0,0 +1,21 @@
...
+datasets are stored in the THash API which has no randomization in place. As
+a result of this, the hash table can be exploited with a worst case time scenario of
+O(n) where n is the total number of entries in the table as a result of excessive chaining
+in a single row.
+
+The test shows that it takes excess time for the THash API to load the datasets from the file
+as many of them evaluate the exact same hash using the algorithm so this is not even the worst
+case scenario. With bigger dataset and lesser system specs/availability of resources,
+this can be worse. Note that it is not just about the number of datasets as there already
+does exist a test already that loads 1m+ datasets.
```
catenacyber, line 14: Could we do with less than a million entries ?

tests/dataset-hash-collisions/test.yaml
```diff
@@ -0,0 +1,12 @@
+pcap: ../flowbit-oring/input.pcap
+
+requires:
+  features:
+    - HAVE_LIBJANSSON
+  files:
+    - src/datasets.c
```
catenacyber, line 7: nit: these requires are not needed anymore

+ files requires are bad in case we move the datasets.c to rust

inashivb: Interesting insight. I had not thought about this. Thanks a lot! 🙇🏽‍♀️ 
Copy pasted this all from another test. Will change in the next rev

tests/dataset-hash-collisions/README.md
catenacyber: sounds enough

inashivb: Please note that Suricata does exit gracefully w 64k strings. Takes about 2m13s on my system w debug flags enabled. 😉

catenacyber: Do you mean 2m13s with or without the patch ?

inashivb: without the patch 😉

PR #2120, opened 2024-11-06
tests: add lua dataset tests

victorjulien:


tests/datasets/datasets-lua-02/dataset-dns.lua
```diff
@@ -0,0 +1,32 @@
...
+
+function match (args)
+    ipver, srcip, dstip, proto, sp, dp = SCFlowTuple()
+    query = DnsGetDnsRrname()
+    if query == nil then
+        return 0
+    end
+    str = ipver .. ":<" .. srcip .. ">:<" .. dstip .. ">:" .. dp .. "--" .. query
+
+    ret, err = dataset.add(dns_new, str, #str);
```
jasonish, line 23: I think idiomatic Lua is more likely see us a method on `dns_new`, for example, looking at the `io` library you'll seem something like this for writing to files:
```
local f = io.open("filename", "w")
f:write("hello world")
```
Not really saying one is better than the other, just commenting on what I see in the standard library.

PR #2125, opened 2024-11-08
rule-types: add check for more rule types - v1

jufajardini:
The goal is to have at least one rule for each rule type the engine has.

`DE_only` and `like_ip_only` were missing


victorjulien (approved):

victorjulien:
Merged in #2139, thanks!

PR #2127, opened 2024-11-12
applayer: add test for dcerpc req http resp

inashivb:


victorjulien:
I think the TCP session should be a bit cleaner on the TCP level.

tests/dcerpc-request-http-response/test.yaml
```diff
@@ -0,0 +1,12 @@
+args:
+- -k none
+- --set stream.midstream=true
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        app_proto: dcerpc
```
catenacyber, line 10: Should we not have http in the end ?

catenacyber: This diff may come from `stream.midstream=true`

inashivb: oh. How so?
without the midstream option, Suricata is unable to detect any appproto at all

inashivb: Interesting thing is that the stats event reports 0 flows of dcerpc or http despite a flow entry being logged.

catenacyber: > Interesting thing is that the stats event reports 0 flows of dcerpc or http despite a flow entry being logged.
Edit: learned that it only gets updated when Suricata has completed detection on either side and some error conditions, so, this test is in fact not in sync w the unittest it is supposed to mimic.

Stats are known to be incomplete and full of corner cases...

catenacyber: But ok for dcerpc as final proto like the unit test said you should test alproto_tc and alproto_ts like is done in the unit test

inashivb: ok. Thank you very much for explaining! 🙇🏽‍♀️

inashivb: I did a new rev and things work as you have explained only if i keep the TCP pkt a certain way. I have explained and asked for your opinion here https://github.com/OISF/suricata-verify/pull/2140#discussion_r1858363910
Thank you very much! 🙇🏽‍♀️

PR #2130, opened 2024-11-15
detect: add test for flow.rate keyword

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5647


tests/rules/flow_rate/test.rules
```diff
@@ -0,0 +1 @@
+pass tcp any any -> any any (msg:"Testing flow.rate"; flow.rate:500; alert; sid:1;)
```
catenacyber, line 1: Question seeing this rule :
What is the unit ?
Can it be a floating number ?

PR #2133, opened 2024-11-20
tests: add keyword check to requires test

jasonish:
Only for 8.0 for now.

requires-fail: With the change to unknown requires statements treated as
not meeting requirements, update the rule to use an unknown keyword to
make it fail out.

Ticket: #7403

### DIFF
--- a/tests/requires-fail/test.rules
+++ b/tests/requires-fail/test.rules
@@ -14,4 +14,4 @@ alert udp any any -> any any (vxlan_vni:10; requires: version >= 10; sid:2;)
 alert http any any => any any (requires: version >= 10; sid:3;)
 alert tcp any any -> any any (frame:smtp.not_supported; requires: version >= 10; sid:4;)
 
-alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)
+alert asdf any any -> any any (requires: version >= 6; foo: bar; sid:102; rev:1;)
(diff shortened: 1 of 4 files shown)


tests/requires-fail/test.rules
```diff
@@ -14,4 +14,4 @@ alert udp any any -> any any (vxlan_vni:10; requires: version >= 10; sid:2;)
 alert http any any => any any (requires: version >= 10; sid:3;)
 alert tcp any any -> any any (frame:smtp.not_supported; requires: version >= 10; sid:4;)
 
-alert asdf any any -> any any (requires: version >= 6, foo bar; sid:102; rev:1;)
+alert asdf any any -> any any (requires: version >= 6; foo: bar; sid:102; rev:1;)
```
catenacyber, line 17: Why do we change this existing test ?

jasonish: We want the test to fail parsing with an invalid requires expression, not just an unknown one.

jasonish: More details. In master, `requires: version >= 6, foo bar` is considered as all requirements met. So this rule fails out as `asdf` is unknown.  With the PR, this rule is considered without requirements met, so the `asdf` is never parsed. This is to test somewhat of an edge case documented in ticket https://redmine.openinfosecfoundation.org/issues/6710, which I've added to the README now as well.

catenacyber: But in the new version `foo: bar;` is totally unused then, right ?

And it would still be good to test `requires: version >= 6, foo bar;` somehow, right ?

victorjulien, line 17: is the failure here not due to `alert asdf...` ?

victorjulien: ah yes, maybe one day I will learn to read. Thanks Philippe.

victorjulien (approved):

victorjulien:
Merged in #2158, thanks!

PR #2088, opened 2024-10-10
tests: add rule type check for icmp_id - v1

AkakiAlice:
Adding SV tests for icmp_id keyword
 
Related to:
Ticket: #6360

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6360

Suricata PR: https://github.com/OISF/suricata/pull/11945

PR #2134, opened 2024-11-22
test: add test for vlan.id - v3

AkakiAlice:
Ticket: #1065

Description:
- Add Suricata-Verify test for vlan.id keyword

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/1065

Suricata PR: https://github.com/OISF/suricata/pull/12103

Previous PR: https://github.com/OISF/suricata-verify/pull/2124


tests/detect-vlan-id/test.rules
```diff
@@ -0,0 +1,3 @@
+alert ip any any -> any any (msg:"Vlan ID is equal to 200 with especific layer"; vlan.id:200,1; sid:1;)
```
catenacyber, line 1: nit typo `especific`

tests/detect-vlan-id/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+args:
+- -k none
+
+checks:
+- filter:
+   count: 1
+   match:
+     event_type: alert
+     alert.signature_id: 1
```
catenacyber, line 12: Could we check the vlan id in the alert data ?

AkakiAlice: Sorry, I didn't understand the question. Could you be more specific, please?

catenacyber: Can we we add in this match section a check about `vlan.id: 200` ? (with the good syntax and value)

AkakiAlice: I don't think so. Should I use these other commits as examples for the implementation?
- https://github.com/OISF/suricata/pull/11945/commits/a8fe7a54e70c4d19c4c2a0accd0e38996f023341
- https://github.com/OISF/suricata-verify/pull/2088/commits/e5861a895ceb07b6af667052156a923309983132

catenacyber: schema.json tells me that it should be `vlan[0]: 200`

PR #2137, opened 2024-11-25
detect-xor: do not depend on C code

inashivb:
@catenacyber one of the exhibits of the problem you mentioned 🥲

### DIFF
--- a/tests/detect-xor/test.yaml
+++ b/tests/detect-xor/test.yaml
@@ -1,7 +1,3 @@
-requires:
-  files:
-    - src/detect-transform-xor.c
-
 args:
   - -k none
 checks:


victorjulien (approved):

catenacyber:
Nice find, how did you find it ?

inashivb:
> Nice find, how did you find it ?

Was just trying to reproduce the memleak issue you fixed w xor transform..

catenacyber:
So, there are a lot more :
```
git grep '\- src/' | cut -d: -f2- | cut -d/ -f2- | sort | uniq -c
   4 detect-transform-casechange.c
   3 detect-transform-dotprefix.c
   1 detect-transform-xor.c
```

catenacyber:
`git grep '\- src/' | cut -d: -f2- | cut -d/ -f2- | sort | uniq | while read i; do echo -n $i; ls ../../suricata/src/$i; done | grep No` full one-liner

inashivb:
Please consider PR #2143 instead of this. That one is more complete.

PR #2140, opened 2024-11-26
applayer: add test for dcerpc req http resp

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/2127

Changes since v1:
- made the session cleaner on TCP level


tests/dcerpc-request-http-response/writepcap.py
```diff
@@ -0,0 +1,15 @@
...
+from scapy.all import *
+
+pkts = []
+
+load_layer("http")
+load_layer("dcerpc")
+pkts += Ether(dst='05:04:03:02:01:00', src='00:01:02:03:04:05')/Dot1Q(vlan=6)/IP(src='1.1.1.1', dst='2.2.2.2')/TCP(dport=8080,sport=12345,flags='S',seq=1,options=[('WScale', 14)])
+pkts += Ether(src='05:04:03:02:01:00', dst='00:01:02:03:04:05')/Dot1Q(vlan=6)/IP(dst='1.1.1.1', src='2.2.2.2')/TCP(sport=8080,dport=12345,flags='SA',seq=1000,ack=2,options=[('WScale', 14)],window=65535)
+pkts += Ether(dst='05:04:03:02:01:00', src='00:01:02:03:04:05')/Dot1Q(vlan=6)/IP(src='1.1.1.1', dst='2.2.2.2')/TCP(dport=8080,sport=12345,flags='AP',seq=2,ack=1001,window=65535)/DceRpc(ptype=0, call_id=1)
+pkts += Ether(src='05:04:03:02:01:00', dst='00:01:02:03:04:05')/Dot1Q(vlan=6)/IP(dst='1.1.1.1', src='2.2.2.2')/TCP(sport=8080,dport=12345,flags='AP',seq=1001,ack=18,window=65535)/HTTP()/HTTPResponse()
```
inashivb, line 11: I believe that the ack number here should be 18. However, by doing this, Suricata is unable to complete detection in toclient direction and it considers dcerpc to be the final applayer proto. Changing the ack number to 2 works though.
Are you able to tell what am I doing wrong?

inashivb: I had tried w that, didn't help. :(
From a recent issue, I had learned that ACKs could get lost and Suricata would continue to parse things normally (at least in master)..

catenacyber: I would still begin so to have a more normal pcap ;-)

inashivb: ok sending updated PR. Thank you!

inashivb: Turns out a standalone call to `DceRpc` w scapy creates a pkt with just the header. However, since the header contains the type of header i.e. `request`, our parser tries to parse the request and fails w insufficient data.
So, in a nutshell,
1. dcerpc request is incomplete hence rejected by Suricata parser
2. The error due to (1) tells the app layer to be disabled
3. Because of (2), the detection stops at dcerpc and never reaches http as applayer is disabled
4. Hence, final proto == dcerpc

TODO: Fix the dcerpc pkt and the changes in the TCP segments accordingly

PR #2145, opened 2024-11-28
Flow bytes pkts either/v4

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5646

Previous PR: https://github.com/OISF/suricata-verify/pull/2082

Changes since v3:
- Rolled back to the new proposed syntax as per discussion and made either in accordance w that
- Also kept tests for old syntax for backward compatibility

### DIFF
--- a/tests/detect-flow-pkts-either/test.rules
+++ b/tests/detect-flow-pkts-either/test.rules
@@ -1,0 +1,1 @@
+alert ip any any -> any any (msg:"Flow has more than 3000 bytes and 10 pkts in either direction";flow.pkts:either,=10;flow.bytes:either,>3000; sid:1;)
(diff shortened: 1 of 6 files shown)


tests/detect-flow-pkts/test.rules
```diff
@@ -1 +1,2 @@
-alert ip any any -> any any (msg:"Flow has 10 packets"; flow.pkts_toclient:10; flow.pkts_toserver:10; flow.bytes_toserver:<3000; flow.bytes_toclient:>8000; sid:1;)
+alert ip any any -> any any (msg:"Flow has 10 packets"; flow.pkts:toclient,=10; flow.pkts:toserver,=10; flow.bytes:toserver,<3000; flow.bytes:toclient,>8000; sid:1;)
```
catenacyber, line 1: Why did you change rule 1 ?

tests/detect-flow-pkts-either/test.rules
```diff
@@ -0,0 +1 @@
+alert ip any any -> any any (msg:"Flow has more than 3000 bytes and 10 pkts in either direction";flow.pkts:either,=10;flow.bytes:either,>3000; sid:1;)
```
catenacyber, line 1: I guess it would be nice to have the signatures with other directions and see which one match, and which ones do not match

tests/detect-flow-pkts/test.rules
inashivb: I think just to check that both the syntax work for equality check.

tests/detect-flow-pkts-either/test.rules
inashivb: Indeed. Will add them. Thank you! 🙇🏽‍♀️

PR #2148, opened 2024-11-28
Bug 7318/v3

victorjulien:
Various tests related to changes in handling of flow timeout packets.


tests/firewall/firewall-06-tls-sni-enforce/test.yaml
```diff
@@ -6,11 +6,24 @@ args:
 
 checks:
 - filter:
+    min-version: 8
+    count: 24
```
victorjulien, line 10: this I think makes sense, the sig is
```
drop ip any any -> any any (msg:"Drop all else"; flow:stateless; sid:3;)
```
If there are no "real" packets, why raise alerts?

tests/bug-5464-verdict-06/test.yaml
```diff
@@ -9,6 +9,22 @@ args:
 
 checks:
   - filter:
+      min-version: 8
+      count: 25
```
victorjulien, line 13: The rule here is
```
alert http any any -> any any (sid: 1;)
```
Should it match on the stream end packets?

@catenacyber are there cases where only at the end of the stream, with the pseudo packets we detect the app protocol? I guess we could, if data stays un-ack'd, then stream terminates, FFR will force handling of un-ACK'd data, detect protocol.

catenacyber: I also guess we could indeed...

victorjulien: https://github.com/OISF/suricata-verify/pull/2148/commits/45b19c663ad0f8aa6fe549513d854efd9dc11ac2

victorjulien: It matched in master on sid 2, in https://github.com/OISF/suricata/pull/12169 it won't match.

catenacyber:
I think this should get closed in favor of the other couple of Suricata + SV PR

victorjulien:
#2174 has just the sig packet type test updates

PR #2152, opened 2024-11-29
tests: verify valid json on long message

regit:
## Ticket 7419

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7419


tests/test-valid-json/test.yaml
```diff
@@ -0,0 +1,17 @@
+requires:
+  min-version: 8
+
+command: |
```
victorjulien, line 4: can this be replaced by something like
```
args:
- -T
pcap: false
```
?

victorjulien: e.g. https://github.com/OISF/suricata-verify/blob/master/tests/content-incomplete-hex-t-version-7-plus/test.yaml

regit: Really cleaner indeed.

regit: I've forced pushed but should I do another MR ?

victorjulien: I think force push is fine here for a case like this.

tests/test-valid-json/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+checks:
+  # check that we have the following entries in eve.json
+  # match rule load failure reason as it is super long
+  - filter:
+      count: 2
+      match:
+        event_type: engine
+        engine.module: detect
+
```
catenacyber, line 19: What is the valid json the name of the test refers to ?
Why does this rule fail to load ?

regit: The signature is so long that it triggers truncation of the log message which end up not being proper json because of that. For the signature it self, I think it is missing a semi coma but this is not critical for the test.

catenacyber: Ok thanks

victorjulien (approved):

victorjulien:
Merged in #2263, thanks!

PR #2121, opened 2024-11-07
add test for extra tls alert

victorjulien:
#2080 rebased

PR #2153, opened 2024-11-29
rule-types: add more rules - v1

jufajardini:
To accompany/ support: https://github.com/OISF/suricata/pull/12184
Sharing this because some of the examples and conclusions seen there are derived from experimentations as seen here.

The purpose of these is to serve as examples of several different types of rules and what their types will be, according to the engine.
This is a draft because:
- I do think I've introduced redundancy
- I want to have the `raw` rule accompanying each check, to make it easier if someone is using this for studies
- I think the `.rules` file should be better organized/ documented
- I'm not sure adding so many checks to this existing test will be well regarded 😅 

I tried to also add checks to see what we have https://github.com/OISF/suricata-verify/pull/2121 and https://github.com/OISF/suricata/pull/12095#issuecomment-2493822457 (to see the differences, so to speak)

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/7031


tests/rule-types/rule-types.rules
```diff
@@ -9,10 +9,18 @@ alert tcp-pkt any any -> any any (msg:"tcp-pkt, anchored content"; content:"abc"
 alert tcp any any -> any any (msg:"tcp, no content"; sid:301;)
 alert tcp any any -> any any (msg:"tcp, simple content"; content:"abc"; sid:302;)
 alert tcp any any -> any any (msg:"tcp, anchored content"; content:"abc"; startswith; sid:303;)
-alert tcp !192.168.0.1 any -> any any (msg:"tcp, negated IP address"; sid:304;)
+alert tcp !192.168.0.1 any -> any any (msg:"tcp, has negated IP address"; sid:304;)
+alert tcp [10.0.0.0/8,!10.10.10.10] any -> [10.0.0.0/8,!10.10.10.10] any (msg:"tcp, has negated IP address"; sid:305;)
```
catenacyber, line 13: Should we split this test in one test per rule type ?

tests/iprep-12-rule-types/scirius-categories.txt
```diff
@@ -0,0 +1,10 @@
+1,2520000,ET TOR Known Tor Exit Node Traffic
```
catenacyber, line 1: Do we need a big tests/iprep-12-rule-types/scirius-iprep.list ?

tests/iprep-12-rule-types/test.yaml
```diff
@@ -0,0 +1,44 @@
...
+      match:
+        id: 2
+        type: "ip_only"
+  - filter:
+      filename: rules.json
+      count: 1
+      match:
+        id: 3
+        type: "ip_only"
+  - filter:
```
catenacyber, line 27: Could we do with 2 rules instead of 6 ?

jufajardini: possibly there are likely duplicates here and there, but in many cases what I'm doing is trying to check if variations here and there affect the rule type (or state) in some way, so the volume will likely be inevitable.

tests/rule-types/rule-types.rules
jufajardini: Could be a good idea, except for cases where we want to see the interaction of keywords and rule types or states.

catenacyber (changes requested):
This is to be improved, right ?

tests/iprep-12-rule-types/scirius-categories.txt
jufajardini: you're right, we don´t

jufajardini:
Finally replaced by: https://github.com/OISF/suricata-verify/pull/2259

PR #2154, opened 2024-11-30
tests: add bug 7422 tests

victorjulien:
Tests various forms of RST triggering handling of unACK'd data.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7422


tests/tcp-rst-unacked-stream-12/test.yaml
```diff
@@ -0,0 +1,26 @@
...
+
+args:
+  - -k none
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: fileinfo
+        fileinfo.sha256: 2a6d1d2d85129cf9e84290a94e7b4d7cfe09d80c47a899dbc04cc61cc737c5a4
```
inashivb, line 12: Unable to understand the file behavior in this case.
Suricata logs `file.size: 17146` and `file.gaps: false`.
But, there's a gap right after 6th file segment. Also, there is a lot more data than 17146 bytes. Could you please tell:

Q1: When should the file processing ideally stop?
Q2: Why are no gaps logged in fileinfo event?

inashivb (commented):
The tests are looking good and match the analysis given on Suricata PR. Asked question inline about file behavior.

victorjulien:
replaced by #2215

PR #2158, opened 2024-12-02
next/sv/653/20241202/v1

victorjulien:
Staging:
 - #2133

### DIFF
--- a/tests/requires-fail/README.md
+++ b/tests/requires-fail/README.md
@@ -1,3 +1,5 @@
 Similar to `../requires-ok` but does include one rule that will fail
 to load. This is to test that a bad rule after "skipped" rule fails
 out and is not recorded as skipped.
+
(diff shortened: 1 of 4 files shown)


jufajardini:
What about the comment here? https://github.com/OISF/suricata-verify/pull/2133#discussion_r1864999886

jasonish:
> What about the comment here? [#2133 (comment)](https://github.com/OISF/suricata-verify/pull/2133#discussion_r1864999886)

I think the test serves its purpose, just need a rule to fail loading.  The previous way of relying on it to fail, use an unknown requires and a unknown action didn't cause it to fail with the new code, but be skipped.  Its catching and edge cases during development were a skipped rule followed by a failed to load rule wasn't being caught properly.

jasonish (approved):

PR #2170, opened 2024-12-08
Verify bypass behavior under different traffic profiles v4

lukashino:
Follow-up of https://github.com/OISF/suricata-verify/pull/2114

Redmine ticket:

Note:

Describe changes:
v4:
- rebased
- P's comment wrt PCAP reusing - yes, I am resuing the PCAP but not sure where you are heading with that - maybe dedup? not sure how to share that 

v3:
- rebased


lukashino:
Dedup in: https://github.com/OISF/suricata-verify/pull/2171

PR #2190, opened 2024-12-17
Run all tests 7461 v1

catenacyber:
## Ticket

- enable DHCP test : runs in less than a second like other tests for me, ticket is closed since long
- remove obsolete tests that are never run (like unified2)
- ja3/ja4: use skip for better log output `===> ja4-rules-requires-off: SKIPPED: not for feature HAVE_JA4` instead of `===> ja4-rules-requires-off: SKIPPED: script`
- Fix https://github.com/OISF/suricata-verify/security/dependabot/1


victorjulien (approved):

victorjulien:
Merged in #2321, thanks!

PR #2202, opened 2024-12-20
Issue 7357 v2

regit:
Update of https://github.com/OISF/suricata-verify/pull/2111:
- remove test "fix" to keep existing logic

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7357


tests/filestore-v2.10-wrong-direction/test.rules
```diff
@@ -0,0 +1,2 @@
+alert http any any -> any any (msg:"alert png images"; http.uri; content:".png"; nocase; sid:1; rev:1;)
+alert http any any -> any any (msg:"store png images"; http.uri; content:".png"; nocase; filestore; sid:2; rev:1;)
```
catenacyber, line 2: I do not think this should match, the others look ok

tests/filestore-v2.13-filedata/test.rules
```diff
@@ -0,0 +1 @@
+alert http any any -> any any (http.uri; content: "toto"; filestore; file.data; content: "tata"; sid:1;)
```
catenacyber, line 1: Thanks for adding this, but I think my remark was about false positive (instead of false negative)

So, we want to check that `alert http any any -> any any (http.uri; content: "toto"; filestore; file.data; content: "titi"; sid:2;)` does not match

tests/filestore-v2.10-wrong-direction/test.rules
catenacyber: see discussion at https://github.com/OISF/suricata/pull/12312#issuecomment-2610083420

catenacyber:
Closing as stale as discussed with Eric

PR #2205, opened 2024-12-22
Datajson v3.1

regit:
Update of #2187. It adds a few more tests for new feature.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7372


catenacyber (changes requested):
Tests should pass for Suricata 7 ;-)

PR #2206, opened 2024-12-23
test: add test for ldap.request.operation - v1

AkakiAlice:
Ticket: 7453

Description:
- Add S-V test for ldap.request.operation keyword

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7453
Suricata PR: https://github.com/OISF/suricata/pull/12321


tests/detect-ldap-request-operation/README.md
```diff
@@ -0,0 +1,7 @@
+# Test Purpose
+
+Test keyword ldap.request.operation.
+
+## PCAP
+
+This PCAP was generated with flowsynth v1.4.1.
```
catenacyber, line 7: I wonder if you can simply reuse some (not all) ldap tests and add a rules file with them

tests/detect-ldap-request-operation/test.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - -k none
+
+pcap: input.pcap
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 1
```
catenacyber, line 14: We want to test app-layer metadata logging also so that there is a `ldap.request.operation` or similar

tests/detect-ldap-request-operation/test.rules
```diff
@@ -0,0 +1,9 @@
+alert tcp any any -> any any (msg:"Test LDAP bind request"; ldap.request.operation:0; sid:1;)
```
catenacyber, line 1: This is good for a first iteration.

In a next iteration, we want to have both signatures ` ldap.request.operation:0;` and ` ldap.request.operation:bind;` (I say `bind` but we should take whatever is being logged as metadata)

See websocket.opcode and usage of `git grep EnumString` in suricata

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/2216 right ?

PR #2212, opened 2025-01-04
transform/test: from_base64 tests using set_error

jlucovsky:
These tests use the set_error transform option to signal buffers that cannot be base64-decoded.

The absent keyword detects the signaling on a buffer that can't be decoded; absent with the or_else option is used when a buffer can be decoded.

## Ticket

changes to the ticket status:

Redmine ticket:

### DIFF
--- a/tests/from_base64-04/test.rules
+++ b/tests/from_base64-04/test.rules
@@ -1,0 +1,6 @@
+# input pcap contains a query to http://home.regit.org/?arg=dGhpc2lzYXRlc3QK
+# "dGhpc2lzYXRlc3QK" is "thisisatest\n"
+alert tcp any any -> any any (msg:"from_base64: no-decode [mode rfc4648]"; flow:to_server,established;  http.uri; content:"/?arg="; from_base64: set_error; absent;  sid:1; rev:1;)
+alert tcp any any -> any any (msg:"from_base64: no-decode with or_else [mode rfc4648]"; flow:to_server,established;  http.uri; content:"/?arg="; from_base64: set_error; absent: or_else; content: "foobar";  sid:2; rev:1;)
+alert http any any -> any any (msg:"from_base64: decode with or_else [mode rfc4648]"; http.uri; content:"/?arg=dGhpc2lzYXRlc3QK"; from_base64: offset 10; absent: or_else; content:"sisatest"; sid:3; rev:1;)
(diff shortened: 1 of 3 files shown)


tests/from_base64-01/test.yaml
```diff
@@ -32,3 +32,8 @@ checks:
       match:
          event_type: alert
          alert.signature_id: 5
+  - filter:
+      count: 1
+      match:
+         event_type: alert
+         alert.signature_id: 6
```
catenacyber, line 39: Why this test change ?

tests/from_base64-04/test.rules
```diff
@@ -0,0 +1,6 @@
+# input pcap contains a query to http://home.regit.org/?arg=dGhpc2lzYXRlc3QK
+# "dGhpc2lzYXRlc3QK" is "thisisatest\n"
+alert tcp any any -> any any (msg:"from_base64: no-decode [mode rfc4648]"; flow:to_server,established;  http.uri; content:"/?arg="; from_base64: set_error; absent;  sid:1; rev:1;)
+alert tcp any any -> any any (msg:"from_base64: no-decode with or_else [mode rfc4648]"; flow:to_server,established;  http.uri; content:"/?arg="; from_base64: set_error; absent: or_else; content: "foobar";  sid:2; rev:1;)
```
catenacyber, line 4: `absent: or_else; content: "foobar";` looks strange to me, what did you mean ?
Did you mean `absent: or_else; content: !"foobar";` ?

tests/from_base64-04/test.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - -k none
+
+checks:
+  - filter:
+      count: 1
+      match:
+         event_type: alert
+         alert.signature_id: 1
+  - filter:
+      count: 0
```
catenacyber, line 16: Why sid 1 matches and not sid 2 ?
I would expect sid 2 to match as well

tests/from_base64-04/test.rules
```diff
@@ -0,0 +1,6 @@
...
+alert http any any -> any any (msg:"from_base64: decode with or_else [mode rfc4648]"; http.uri; content:"/?arg=dGhpc2lzYXRlc3QK"; from_base64: offset 10; absent: or_else; content:"sisatest"; sid:3; rev:1;)
```
catenacyber, line 5: How could `from_base64` without `set_error` match `absent` ?

This rule should be rejected as invalid...

catenacyber (changes requested):
Needs rebase + see comments/questions inline

PR #2215, opened 2025-01-06
Flow timing/v4.1 stream rst/v1.1

victorjulien:
#2154 and #2203


tests/tcp-rst-unacked-stream-09/suricata.yaml
```diff
@@ -0,0 +1,560 @@
+%YAML 1.1
```
catenacyber, line 1: Why do we need a so big suricata.yaml ?

tests/tcp-rst-unacked-stream-12/test.yaml
```diff
@@ -0,0 +1,26 @@
...
+  - filter:
+      count: 1
+      match:
+        event_type: http
+  - filter:
+      count: 1
+      match:
+        event_type: stats
+        stats.app_layer.tx.http: 1
+          #stats.app_layer.flow.http: 1
```
catenacyber, line 26: nit duplicate

tests/tcp-rst-unacked-stream-10/test.yaml
```diff
@@ -0,0 +1,23 @@
+requires:
+  min-version: 8
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: fileinfo
+        fileinfo.sha256: 8ff57c7fc0d4babd27e2e914ad9b556b1b980a69710d3917266ec1cb4edbb782
```
catenacyber, line 9: Why do we need multiple tests ? What is the difference between tcp-rst-unacked-stream-10 and tcp-rst-unacked-stream-9 ?

tests/tcp-rst-unacked-stream-09/suricata.yaml
victorjulien: it's not so big I think, just app later and eve sections really

tests/tcp-rst-unacked-stream-09/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: fileinfo
+        fileinfo.sha256: b95aa84c9ac4948c8565202e016933644c592c366525b2790857615ca7e6f665
+  - filter:
+      count: 1
+      match:
+        event_type: fileinfo
```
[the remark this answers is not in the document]
victorjulien: I don't think the reason is so important here, what matters is that there should obviously just be a single file if you look at the pcap.

The reasoning of the code changes is in the suri commits.

tests/tcp-rst-unacked-stream-12/test.yaml
victorjulien: I've left it there because ideally we uncomment it, but sadly the flow app-layer counting has issues and so it won't work

tests/tcp-rst-unacked-stream-10/test.yaml
victorjulien: they are just multiple failure cases from tlpw1, showing slightly different issues (e.g. the flow counting being off in one shows that there are multiple paths taken)

tests/tcp-rst-unacked-stream-09/suricata.yaml
catenacyber: Sure, but how is it different than the default suricata.yaml ?

tests/tcp-rst-unacked-stream-12/test.yaml
catenacyber: fair enough, your comment in the PR is better than the comment in the test.yaml ;-)

tests/tcp-rst-unacked-stream-09/suricata.yaml
catenacyber: (no big deal anyways)

tests/tcp-rst-unacked-stream-10/test.yaml
catenacyber: > (e.g. the flow counting being off in one shows that there are multiple paths taken)

I do not understand this

Running suricata master on these tests I find ` stats.app_layer.flow.http: 1` for both tcp-rst-unacked-stream-09 and tcp-rst-unacked-stream-10
(so my interpretation of ` flow counting being off in one` is not correct)

tests/tcp-rst-unacked-stream-09/test.yaml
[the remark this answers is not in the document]
catenacyber: Fair enough indeed : the test is clean and obvious, even if the way to make it fail is not obvious

victorjulien:
replaced by #2226

PR #2220, opened 2025-01-08
add extra tests around tls and tcp rulesets and one around using flow…

aneesh-patel:
This PR adds some tests to more thoroughly test TCP/TLS rule-sets. It also adds a test around using flow-established/not-established with tcp/tls protocols.

## Ticket

changes to the ticket status:

This is not related to a ticket specifically, but is meant to add extra tests to prevent regressions.

Redmine ticket:

N/A


catenacyber:
Thanks for this.

These tests are expected to pass for 7 and 8 without any need to change Suricata, right ?

Do you have a little more context about these tests ? (that could be put in some README.md files ? )

catenacyber (changes requested):
suricata-verify does not accept pcapng in order to run on the platforms which do not support it...

catenacyber:
Closing as stale, feel free to reopen a new PR which should pass CI (not using pcapng)

PR #2265, opened 2025-01-29
Introduce TLS-JA4 client/server handshake tests

rmcconnell-r7:
This update introduces two new tests to accompany the introduction of client/server handshake parameters and output via JSON-EVE.

- ja4-cl-handshake: client eve output test
- ja4-sv-handshake: server eve output test

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6695


tests/ja4-sv-handshake/test.yaml
```diff
@@ -0,0 +1,14 @@
+requires:
+  min-version: 8
+  features:
+    - HAVE_JA4
```
catenacyber, line 4: Could we have a test that does not require JA4 ?

rmcconnell-r7: Could you elaborate for me, please? So you want a test that mirrors the above but just with that removed? So are you expecting it to produce no output? Apologies, I've just not quite understood what you want here.

catenacyber: I expected that we can output `tls.client_handshake.exts` even without ja4 enabled...
Is that wrong ?

rmcconnell-r7:
Replaced by: https://github.com/OISF/suricata-verify/pull/2313

PR #2282, opened 2025-02-11
detect: add test for ldap.responses.result_code - v1

AkakiAlice:
Ticket: [#7532](https://redmine.openinfosecfoundation.org/issues/7532)

Description:
- Add S-V test for  ``ldap.responses.result_code`` keyword

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7532
Suricata PR: https://github.com/OISF/suricata/pull/12555

### DIFF
--- a/tests/detect-ldap-result/test.yaml
+++ b/tests/detect-ldap-result/test.yaml
@@ -1,0 +1,18 @@
+requires:
+  min-version: 8
+
+args:
+  - -k none
+  - --set stream.midstream=true
+
+pcap: ../ldap-unsolicited/ldap.pcap
+
+checks:
+  - filter:
+      count: 1
+      match:
+        pcap_cnt: 2
+        event_type: alert
+        ldap.responses[0].operation: extended_response
+        ldap.responses[0].extended_response.result_code: "unavailable"
+        alert.signature_id: 1
--- a/tests/detect-ldap-result/README.md
+++ b/tests/detect-ldap-result/README.md
@@ -1,0 +1,5 @@
+Test ldap.responses.result_code keyword.
+
+PCAP from ../ldap-search/ldap.pcap
+
+Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7532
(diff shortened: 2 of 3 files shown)


AkakiAlice:
I don't understand why this test is not generating any alert. I tried with the pcap from the file ``ldap-search`` and it was working normally

tests/detect-ldap-result/test.yaml
```diff
@@ -0,0 +1,18 @@
+requires:
+  min-version: 8
+
+args:
+  - -k none
+  - --set stream.midstream=true
+
+pcap: ../ldap-unsolicited/ldap.pcap
```
inashivb, line 8: perhaps because it's using a different pcap file?

AkakiAlice: This pcap ``../ldap-unsolicited/ldap.pcap`` is the one that I want to test. I was using the pcap from ``ldap-search`` and it was working, but I changed the pcap because I also want to test the ldap error message that is present here

tests/detect-ldap-result/test.yaml
```diff
@@ -0,0 +1,18 @@
...
+
+checks:
+  - filter:
+      count: 1
+      match:
+        pcap_cnt: 2
```
catenacyber, line 14: Second packet is not in toclient direction !

catenacyber:
> I don't understand why this test is not generating any alert. I tried with the pcap from the file `ldap-search` and it was working normally

It looks like this transaction gets cleaned by `AppLayerParserTransactionsCleanup` before detection is run in the to_client direction. It should not cf usage of `detect_flags_tc = AppLayerParserGetTxDetectFlags(txd, STREAM_TOCLIENT);

catenacyber:
`f->sgh_toclient == NULL` when it should have the `ldap.responses.result_code:unavailable; sid:1;`

catenacyber:
So, it looks like there is a suricata bug to dig with this condition of stream.midstream=true, first packet detecting protocol and creating a tx to_client...

AkakiAlice:
Replaced by: https://github.com/OISF/suricata-verify/pull/2291

PR #2291, opened 2025-02-14
detect: add test for LDAPResult keywords - v2

AkakiAlice:
Ticket: [#7532](https://redmine.openinfosecfoundation.org/issues/7532)

Description:
- Add S-V test for  ``ldap.responses.result_code`` and ``ldap.responses.error_message`` keywords

Changes:
- Create a new pcap using flowsynth.py, so the error message can be tested
- Add test for ``ldap.responses.error_message``

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7532

Suricata PR: https://github.com/OISF/suricata/pull/12555
Previous PR: https://github.com/OISF/suricata-verify/pull/2282


tests/detect-ldap-result/test.yaml
```diff
@@ -13,3 +13,12 @@ checks:
...
         ldap.responses[0].search_result_done.result_code: size_limit_exceeded
         alert.signature_id: 1
+checks:
+  - filter:
+      count: 1
+      match:
+        pcap_cnt: 4
+        event_type: alert
+        ldap.responses[0].operation: search_result_done
+        ldap.responses[0].search_result_done.message: "Size limit exceeded"
```
inashivb, line 23: @catenacyber this seems to be way off from the keyword and increases parity b/w rule keyword and eve output. Is this ok?

tests/detect-ldap-result/test.rules
```diff
@@ -1 +1,2 @@
 alert ldap any any -> any any (msg:"Test LDAP result code"; ldap.responses.result_code:size_limit_exceeded; sid:1;)
+alert ldap any any -> any any (msg:"Test LDAP error message"; ldap.responses.error_message; content:"Size limit exceeded"; sid:2;)
```
inashivb, line 2: So, both of them are the same things, one a sticky buffer and one a value?
Was expecting `result_code` to be an int only.. which it does seem to accept as well. Is there a special name for the keywords that accept a string (?) value as well as what that string means? @catenacyber

tests/detect-ldap-result/test.yaml
catenacyber: > this seems to be way off from the keyword and decreases parity b/w rule keyword and eve output.

How so ?

`ldap.responses.error_message; content:"Size limit exceeded";` looks to match `ldap.responses[0].search_result_done.message: "Size limit exceeded"`

This ( with the suricata PR ) increases parity because the eve output has a new keyword to match on

tests/detect-ldap-result/README.md
```diff
@@ -0,0 +1,5 @@
+Test ldap.responses.result_code and ldap.responses.error_message keywords.
+
+PCAP created with flowsynth.py
```
catenacyber, line 3: This is ok but I wonder if we could find "real" pcaps (not synthetic ones, but ones with a real LDAP client talking to a real LDAP server)

tests/detect-ldap-result/ldap.syn
```diff
@@ -0,0 +1,2 @@
+flow default tcp 1.1.1.1:5555 > 2.2.2.2:389 (tcp.initialize; mss:9000;);
+default < (content:"\x30\x1f\x02\x01\x02\x65\x1a\x0a\x01\x04\x04\x00\x04\x13\x53\x69\x7a\x65\x20\x6c\x69\x6d\x69\x74\x20\x65\x78\x63\x65\x65\x64\x65\x64";);
```
catenacyber, line 2: If this is synthetic, I would put another message than `"Size limit exceeded"` to avoid confusion, like `"Message size limit got exceeded : max 65k"`

tests/detect-ldap-result/test.rules
```diff
@@ -0,0 +1,2 @@
+alert ldap any any -> any any (msg:"Test LDAP result code"; ldap.responses.result_code:size_limit_exceeded; sid:1;)
+alert ldap any any -> any any (msg:"Test LDAP error message"; ldap.responses.error_message; content:"Size limit exceeded"; sid:2;)
```
catenacyber, line 2: add `startwith` and `endswith` ;-)

tests/detect-ldap-result/test.rules
catenacyber: > So, both of them are the same things, one a sticky buffer and one a value?

No

There is one integer which encodes a status as an enumeration, and a string which can be whatever...

See https://docs.suricata.io/en/latest/rules/integer-keywords.html#enumerations for the enumeration keywords and also `EnumStringU8` and such in the code

inashivb: ah ok. Thanks. I see you've already given feedback to change the string to something else. Thanks!

tests/detect-ldap-result/test.yaml
inashivb: actually, maybe I'm using the wrong word. I meant that the rule language does not map closely to the eve output. `error_message` != `search_result_done.message`

AkakiAlice:
Replaced by: https://github.com/OISF/suricata-verify/pull/2310

PR #2328, opened 2025-02-28
Tests: add bug-5486

cccs-jqyu:
- add test to check presence of ethernet metadata in events triggered on flow timeout pseudopackets

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5486


tests/bug-5486/test.yaml
```diff
@@ -0,0 +1,11 @@
...
+  min-version: 7
+
+pcap: 154.pcap
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: http
+        has-key: ether
```
inashivb, line 11: Could we also check the values logged in the eve?

tests/bug-5486/test.yaml
```diff
@@ -0,0 +1,11 @@
+requires:
+  min-version: 7
```
inashivb, line 2: For now, please keep the min-version to 8 to avoid having CI in a broken state until this is backported

PR #2366, opened 2025-03-26
http: use lua lib to access data

inashivb:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7604


victorjulien:
Looks like the testing is really light for lua http, so adding some more tests would be great. Esp detection scripts. Seems we had none so until now.

inashivb:
> Looks like the testing is really light for lua http, so adding some more tests would be great. Esp detection scripts. Seems we had none so until now.

Indeed. Will do so in the next rev. Thank you!

PR #2374, opened 2025-03-27
tests: lua dnp3 tests plus other minor fixes

jasonish:


tests/dns-lua-rules/test-rrname.lua
```diff
@@ -1,11 +1,14 @@
+local dns = require("suricata.dns")
+
 function init (args)
    local needs = {}
-   needs["dns.rrname"] = tostring(true)
+   needs["dns.request"] = true
    return needs
 end
```
jasonish, line 7: This could also be:
```
return {["dns.request"] = true}
```
The `[]` is a way of quoting the string such that `.` can be used.

victorjulien (approved):

victorjulien:
Merged in #2378, thanks!

PR #2380, opened 2025-03-28
Firewall/v4

victorjulien:
Firewall related changes tests.


tests/dnp3/dnp3-lua/rule.lua
```diff
@@ -1,7 +1,7 @@
 local dnp3 = require("suricata.dnp3")
 
 function init (args)
-   return {dnp3 = true}
+   return {}
 end
```
jasonish, line 5: Does this mean this type of syntax would not longer work? You have to transition to `dnp3:request_complete` or `dnp3:request_update`? What if you wanted to do inspection on request and response?

victorjulien: there is also `response_complete` and `response_update`. Those are generic. Protocols like http and tls implement their own set of states too, that are more specific.

victorjulien: in general this whole effort shows that much of our tx state progression handling is quite sloppy. Just a few protocols have a more complex state machine, and with those many keywords aren't registered in the correct state.

In many simpler implementations there is just a simple pattern of (0 meaning tx exists but isn't done, 1 done). Looking at where keywords are registered 0 or 1, it seems not very consistent. Missing comments suggest that there usually wasn't made a conscious decision about it.

victorjulien: Btw I've removed `request_update`, it was an artifact from an earlier approach.

Each proto now has `request_started` (0) and `request_completed` (registered completion value), in addition to possible per protocol states.

victorjulien:
non-draft in #2421

PR #2417, opened 2025-04-05
pgsql: update tests for version 8 logging - v1

jufajardini:
Many PGSQL messages were lacking more context due to not having an encapsulating object that indicated which PostgreSQL message they were coming from.

Update test checks to expect the new log style, and leave checks for old style logs for Suri 7.

## Ticket

Redmine ticket:
none yet


jufajardini:
Closing this as it needs more rework.

PR #2420, opened 2025-04-05
ssh: adds test for lua

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7607

PR #2445, opened 2025-04-15
ssh: adds test for lua

catenacyber:
## Ticket

#2420 with new hook syntax


catenacyber:
Next in https://github.com/OISF/suricata-verify/pull/2451

PR #2451, opened 2025-04-18
ssh: adds test for lua

catenacyber:
## Ticket

#2445 with
- using hook indicating direction `response_banner_done` instead of just `banner_done`
- no debug print in lua
- checking only one alert, that happens at flow timeout

### DIFF
--- a/tests/ssh-lua-rules/suricata.yaml
+++ b/tests/ssh-lua-rules/suricata.yaml
@@ -1,0 +1,4 @@
+%YAML 1.1
+---
+
+include: ../../etc/suricata-4.0.3.yaml
--- a/tests/ssh-lua-rules/test-ssh.lua
+++ b/tests/ssh-lua-rules/test-ssh.lua
@@ -1,0 +1,18 @@
+local ssh = require("suricata.ssh")
+
+function init (args)
+   local needs = {}
+   return needs
+end
+
+function match(args)
+   local tx = ssh.get_tx()
+   local proto = tx:server_proto()
+   if proto == "2.0" then
+      local soft = tx:server_software()
+      if soft == "OpenSSH_7.4" then
+         return 1
+      end
+   end
+   return 0
+end
--- a/tests/ssh-lua-rules/test.yaml
+++ b/tests/ssh-lua-rules/test.yaml
@@ -1,0 +1,15 @@
+pcap: ../ssh-banner-only/input.pcap
+
+requires:
+  min-version: 8
+
+args:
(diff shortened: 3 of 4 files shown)


tests/ssh-lua-rules/suricata.yaml
```diff
@@ -0,0 +1,4 @@
+%YAML 1.1
```
victorjulien, line 1: do we need this yaml? can just use default?

tests/ssh-lua-rules/test-ssh.lua
```diff
@@ -0,0 +1,18 @@
+local ssh = require("suricata.ssh")
+
+function init (args)
+   local needs = {}
```
victorjulien, line 4: think we can just do `return {}`

tests/ssh-lua-rules/test.yaml
```diff
@@ -0,0 +1,15 @@
+pcap: ../ssh-banner-only/input.pcap
+
+requires:
+  min-version: 8
+
+args:
+  - --set security.lua.allow-rules=true -k none
```
victorjulien, line 7: `--set security.lua.allow-rules=true` shouldn't be needed

tests/ssh-lua-rules/test.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+checks:
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 1
+        # we do not have a toclient packet after server data is acked
+        pkt_src: "stream (flow timeout)"
```
victorjulien, line 15: can you add a test with `--simulate-ips` and match pcap_cnt to see it matches the expected pkt?

tests/ssh-lua-rules/suricata.yaml
catenacyber: Can be replaced with just `--set default-rule-path=.`

tests/ssh-lua-rules/test-ssh.lua
catenacyber: indeed

tests/ssh-lua-rules/test.yaml
catenacyber: right

tests/ssh-lua-rules/test.yaml
catenacyber: ok

catenacyber:
Next in https://github.com/OISF/suricata-verify/pull/2455

PR #2455, opened 2025-04-19
Lua ssh 7607 v4

catenacyber:
## Ticket

#2451 with review taken into account, and also backported for dns-lua test where this new test comes from


jasonish (commented):
I think it looks OK to me.  2 things tho:
- I think we need a test for ssh use in output
- @victorjulien's opinion on the `hook` names.

victorjulien (approved):

PR #2470, opened 2025-04-28
ssh: adds test with lua and hassh

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7603


catenacyber:
next version in https://github.com/OISF/suricata-verify/pull/2471

PR #2471, opened 2025-04-28
ssh: adds test with lua and hassh

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7603

#2470 without hassh specific lib, but needs["ssh.hassh"]


catenacyber:
Next in https://github.com/OISF/suricata-verify/pull/2475

PR #2481, opened 2025-05-02
ja3: adds tests for lua

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7605

### DIFF
--- a/tests/ja3-lua-rules-quic/test-ja3.lua
+++ b/tests/ja3-lua-rules-quic/test-ja3.lua
@@ -1,0 +1,18 @@
+local ja3 = require("suricata.ja3")
+
+function init (args)
+   ja3.enable_ja3()
+   local needs = {}
+   needs["ja3"] = tostring(true)
+   return needs
+end
+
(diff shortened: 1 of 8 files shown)


tests/ja3-lua-rules-quic/test-ja3.lua
```diff
@@ -0,0 +1,18 @@
+local ja3 = require("suricata.ja3")
+
+function init (args)
+   ja3.enable_ja3()
+   local needs = {}
+   needs["ja3"] = tostring(true)
```
catenacyber, line 6: @jasonish is this ok ?

As we cannot use a hook since ja3 is in both TLS and QUIC protocols, I need to convey the information from lua to suricata that I want a ja3 and ja3 util lib functions do not have access to the current rule, so I resort to `init` return used by `DetectLuaSetup` (and which has access to the rule)

catenacyber:
Next in https://github.com/OISF/suricata-verify/pull/2482

PR #2482, opened 2025-05-02
ja3: adds tests for lua

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7605

#2481 with tls test modified

### DIFF
--- a/tests/ja3-lua-rules-quic/README.md
+++ b/tests/ja3-lua-rules-quic/README.md
@@ -1,0 +1,11 @@
+# Test Description
+
+Test JA3 lua rules with quic
+
+## Related issue
+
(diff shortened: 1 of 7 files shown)


inashivb (commented):
Should we have a simple logging test too?

victorjulien (approved):

jasonish:
> > Should we have a simple logging test too?
> 
> @jasonish does it make sense ?
> 
> By the way in ssh-lua-output test, how could we get/log information about the flow ?

I think it makes sense to have a rule, and an output test for each lib where both is supported.

For flow info, we have a test that logs flow info: `lua-flowtuple`.

victorjulien:
Merged in #2493, thanks!

PR #2499, opened 2025-05-08
test: add lua test for suricata.file lib

jasonish:
Ticket: #7491

### DIFF
--- a/tests/lua-scfileinfo-pre8/expected/scfileinfo.log
+++ b/tests/lua-scfileinfo-pre8/expected/scfileinfo.log
@@ -4,13 +4,9 @@ size: 1054 [**] magic: nomagic
 md5: c0b818de79d068db0a989ed35044cd62
 sha1: afb4fb1ad18223825cfb7ce5bf506e0f9dc47d5c
 sha256: 57b43ee07432cf8a8b8a17d9d712138194e4564e4b36963a34c495b576b404fe
-state: CLOSED
-is_stored: true
 ** SCFileInfo is: [**] fileid: 2 [**] txid: 1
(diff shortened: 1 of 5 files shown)


victorjulien:
Can we also have a detect script test?

jasonish:
> Can we also have a detect script test?

The file lib is for logging only.

victorjulien:
> > Can we also have a detect script test?
> 
> The file lib is for logging only.

Why?

jasonish:
> > > Can we also have a detect script test?
> > 
> > 
> > The file lib is for logging only.
> 
> Why?

In current git master its only implemented, and documented as a logging API.

victorjulien:
> > > > Can we also have a detect script test?
> > > 
> > > 
> > > The file lib is for logging only.
> > 
> > 
> > Why?
> 
> In current git master and 7.0 its only implemented, and documented as a logging API.

Hmm ok. Guess we need to add it to the detect side too then. Is there a ticket for it?

victorjulien (approved):

jasonish:
> > > > > Can we also have a detect script test?
> > > > 
> > > > 
> > > > The file lib is for logging only.
> > > 
> > > 
> > > Why?
> > 
> > 
> > In current git master and 7.0 its only implemented, and documented as a logging API.
> 
> Hmm ok. Guess we need to add it to the detect side too then. Is there a ticket for it?

Ticket: https://redmine.openinfosecfoundation.org/issues/7703

A quick try when I was working on this made it appear not so simple, is that why it wasn't done before? Would this require file state hooks to be added now? So you might only trigger on file completes, as the logging access is based around the file being complete.

victorjulien:
Merged in #2505, thanks!

PR #2527, opened 2025-05-29
tests/rules: add test for bug 5177 - v1

jufajardini:

Bug #5177

Sharing the SV test to show the behavior. I can't see the output in other logs, but as the test shows, the warning is present in the engine analysis `rules.json` output.

## Ticket

Redmine ticket:

### DIFF
--- a/tests/rules/bug-5177/test.yaml
+++ b/tests/rules/bug-5177/test.yaml
@@ -1,0 +1,25 @@
+requires:
+    min-version: 7
+    pcap: false
+
+args:
+- --engine-analysis
+
+checks:
+    - filter:
+        # rule 61 uses the new buffer and shouldn't have a warning issued
+        filename: rules.json
+        count: 1
+        match:
+            id: 61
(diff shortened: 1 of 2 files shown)


tests/rules/bug-5177/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+checks:
+    - filter:
+        # rule 61 uses the new buffer and shouldn't have a warning issued
+        filename: rules.json
+        count: 1
+        match:
+            id: 61
+            raw: "alert http any any -> any any (http.request_line; content:\"GET /index.html HTTP/1.0\"; sid:61;)"
+            warnings[0]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+            warnings[1]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
```
inashivb, line 17: is this the expected behavior? Note that the test must show the expected behavior and a corresponding failure in CI accordingly.

PR #2528, opened 2025-05-29
tests/rules: add test for bug 5177 - v2

jufajardini:

Bug #5177

Previous PR: https://github.com/OISF/suricata-verify/pull/2527

Describe changes:
- change check with unexpected behavior to make test fail with CI

## Ticket
Redmine ticket:


inashivb (changes requested):
Thanks for this @jufajardini ! :)
Seems like we have a fix: https://github.com/OISF/suricata/pull/13332

Could you please set min-version to 8 for this PR and update the test? This can be merged independently of the Suricata PR, I think.

jufajardini:
Replaced by https://github.com/OISF/suricata-verify/pull/2530

PR #2530, opened 2025-05-30
tests/rules: add test for bug 5177 - v3

jufajardini:

Bug #5177

Previous PR: https://github.com/OISF/suricata-verify/pull/2528

Changes from previous PR:
- change min-version to 8

## Ticket

Redmine ticket:

### DIFF
--- a/tests/rules/bug-5177/test.yaml
+++ b/tests/rules/bug-5177/test.yaml
@@ -1,0 +1,25 @@
+requires:
+    min-version: 8
+    pcap: false
+
+args:
+- --engine-analysis
+
+checks:
+    - filter:
+        # rule 61 uses the new buffer and shouldn't have a warning issued
+        filename: rules.json
+        count: 0
+        match:
+            id: 61
+            raw: "alert http any any -> any any (http.request_line; content:\"GET /index.html HTTP/1.0\"; sid:61;)"
+            warnings[0]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+            warnings[1]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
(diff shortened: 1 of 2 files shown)


inashivb:
I'll have to reopen the corresponding Suricata PR to answer so, bringing it here.

> This doesn't really change the validation, though, right?

There was no validation done which is why it looked more appropriate as a `Note`. However, if we do wanna warn, we should do validation first as otherwise a "warning" won't make much sense. Perhaps when it was written, it made sense to have it as warning to increase the adaption of keywords but this is only a guess.

tests/rules/bug-5177/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+            warnings[0]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+            warnings[1]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+    - filter:
+        filename: rules.json
+        count: 1
+        match:
+            id: 62
+            raw: "alert http any any -> any any (http_request_line; content:\"GET /index.html HTTP/1.0\"; sid:62;)"
+            warnings[0]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+            warnings[1]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
```
inashivb, line 25: I think this might fail master

catenacyber:
Redmine ticket was closed yesterday, what is the status of this ?

jufajardini:
> Redmine ticket was closed yesterday, what is the status of this ?

Either this should be updated, or closed, if not considered important. The SV test was just to show the behavior. 

I'll update it to adjust to what we have in master now.

PR #2535, opened 2025-06-01
Datajson v6.5

regit:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7372


tests/datajson/datajson-01-ip/src.lst
```diff
@@ -0,0 +1 @@
+[{"ip": "10.16.1.11", "test": "success", "context":3}]
```
jasonish, line 1: I don't think I see a dataset quite like this in the documentation. Can we get an example there?

PR #2545, opened 2025-06-05
test/dns: update src and dest addresses - v1

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/6400


victorjulien:
For TCP, do we have tests for both IDS and IPS modes?

victorjulien (approved):

PR #2546, opened 2025-06-05
tests: add tests for decode encapsulation types - v1

jufajardini:
Add simple tests for alert to match on TCP traffic over:

- IPv4 over IPv4 - bug-4571-01
- IPv6           - bug-4571-02
- IPv6 over IPv6 - bug-4571-03
- IPv6 over IPv4 - bug-4571-04
- IPv4           - bug-4571-05
- IPv4 over IPv6 - bug-4571-06

Of notice:
- With these tests, I've observed that for encapsulation over IPv6 we see only one flow, even if Suri sees and logs the tunnel.
- Test `bug-4571-01` is for similar situation as test `bug-7725`.

The tests pass, but... `bug-4571-03` and `bug-4571-06` maybe shouldn't

Related to
Bug #4571
Bug #7725

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/4571
https://redmine.openinfosecfoundation.org/issues/7725


catenacyber:
> The tests pass, but... bug-4571-03 and bug-4571-06 maybe shouldn't

What do you mean ?

catenacyber:
> > > The tests pass, but... bug-4571-03 and bug-4571-06 maybe shouldn't
> > 
> > 
> > What do you mean ?
> 
> Because we don't see two flows for those, even though there's tunneled traffic.

So, I think :
- it is ok that the tests pass
- there may be another check to add that currently fails

The question is what is a flow for Suricata ? Let's look at `FlowHandlePacket`, then caller `FlowSetupPacket` we see it is only called in a few cases UDP/TCP/ESP/SCTP/ICMP end IPv4 tunnels
So, it is likely we are missing the call to `FlowSetupPacket` in decode-ipv6.c like in `DecodeIPv4inIPv6`

jufajardini:
> > > > The tests pass, but... bug-4571-03 and bug-4571-06 maybe shouldn't
> > > 
> > > 
> > > What do you mean ?
> > 
> > 
> > Because we don't see two flows for those, even though there's tunneled traffic.
> 
> So, I think :
> 
>     * it is ok that the tests pass
> 
>     * there may be another check to add that currently fails
> 
> 
> The question is what is a flow for Suricata ? Let's look at `FlowHandlePacket`, then caller `FlowSetupPacket` we see it is only called in a few cases UDP/TCP/ESP/SCTP/ICMP end IPv4 tunnels So, it is likely we are missing the call to `FlowSetupPacket` in decode-ipv6.c like in `DecodeIPv4inIPv6`

Indeed.

We also don't have counters for `ipv4-in-ipv4` or ipv6-in-ipv4`

jufajardini:
> > > > The tests pass, but... bug-4571-03 and bug-4571-06 maybe shouldn't
> > > 
> > > 
> > > What do you mean ?
> > 
> > 
> > Because we don't see two flows for those, even though there's tunneled traffic.
> 
> So, I think :
> 
>     * it is ok that the tests pass
> 
>     * there may be another check to add that currently fails
> 
> 
> The question is what is a flow for Suricata ? Let's look at `FlowHandlePacket`, then caller `FlowSetupPacket` we see it is only called in a few cases UDP/TCP/ESP/SCTP/ICMP end IPv4 tunnels So, it is likely we are missing the call to `FlowSetupPacket` in decode-ipv6.c like in `DecodeIPv4inIPv6`

https://redmine.openinfosecfoundation.org/issues/7752 waiting for the release to share a PR.

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/2569

PR #2548, opened 2025-06-05
tests/rules: add test for bug 5177 - v4

jufajardini:
The engine analyzer issues the same note about using new rule buffers regardless of the rule usage of new buffer or the old ones. This test is just to showcase this.

Bug #5177

Previous PR: https://github.com/OISF/suricata-verify/pull/2530

Changes from previous PR:
- change from `warnings` to `notes`
- rebase
- reword commit message

## Ticket

Redmine ticket:

### DIFF
--- a/tests/rules/bug-5177/test.yaml
+++ b/tests/rules/bug-5177/test.yaml
@@ -1,0 +1,24 @@
+requires:
+    min-version: 8
+    pcap: false
+
+args:
+- --engine-analysis
+
+checks:
+    - filter:
+        filename: rules.json
+        count: 1
+        match:
+            id: 61
+            raw: "alert http any any -> any any (http.request_line; content:\"GET /index.html HTTP/1.0\"; sid:61;)"
+            notes[0]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+            notes[1]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
+    - filter:
+        filename: rules.json
+        count: 1
+        match:
+            id: 62
+            raw: "alert http any any -> any any (http_request_line; content:\"GET /index.html HTTP/1.0\"; sid:62;)"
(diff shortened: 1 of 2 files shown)


tests/rules/bug-5177/test.yaml
```diff
@@ -0,0 +1,24 @@
...
+- --engine-analysis
+
+checks:
+    - filter:
+        filename: rules.json
+        count: 1
+        match:
+            id: 61
+            raw: "alert http any any -> any any (http.request_line; content:\"GET /index.html HTTP/1.0\"; sid:61;)"
+            notes[0]: "pattern looks like it inspects HTTP, use http.request_line or http.method and http.uri instead for improved performance"
```
catenacyber, line 15: Why do we get a warning when we do the proper thing ?

jufajardini: Because currently there's no real buffer validation, we just check that the pattern in the `content` buffer aligns with what could be seen in those keywords, and issues that note, regardless of which type of keyword is being used (new or old).

catenacyber: So, the test shows current behavior, but not the right behavior ? Am I understanding correctly ?

jufajardini: This is the behavior that was recently decided upon, for now. From what I understood from the Suricata PR that changed these from Warnings to Notes

catenacyber: I am confused.

Is this SV PR supposed to be merged as is ?
Does it expect a Suricata PR or fix ?

jufajardini: Yes, this should be good as is, as this was the decision for the "bug", cf https://github.com/OISF/suricata/pull/13332 and https://github.com/OISF/suricata-verify/pull/2530#issuecomment-2928791450

victorjulien (approved):

jufajardini:
closed by mistake >__<'

PR #2550, opened 2025-06-06
tests: remove exit-code check for datasets-absolute-allowed

norg:

Ticket: 7083

## Ticket

changes to the ticket status:


victorjulien:
Needs to pass on 70x to be considered for merge.

norg:
> Needs to pass on 70x to be considered for merge.

the only way to achieve this currently, would be to keep that test and set `lt-version: 8` and have another one that is `min-version` doing the same or is there a way to have different exit code checks based on a version directly?

norg:
> We have many tests like that, ya.

https://github.com/OISF/suricata-verify/pull/2553 updated version with two testes.

PR #2553, opened 2025-06-07
tests: remove exit-code check for datasets-absolute-allowed on Surica…

norg:

## Ticket

changes to the ticket status:

### DIFF
--- a/tests/datasets/datasets-absolute-allowed-post8-winonly/test.yaml
+++ b/tests/datasets/datasets-absolute-allowed-post8-winonly/test.yaml
@@ -1,0 +1,22 @@
+pcap: ../../datasets/datasets-parent-path/one-packet.pcap
+
+requires:
+  min-version: 8
+  lambda: sys.platform == "win32"
+
+args:
+  - -vvv
+
+checks:
+  - filter:
+      count: 1
+      match:
+        engine.message: "Allowing absolute filename for dataset rule: C:\Windows\Temp\dns-seen.txt"
+  - filter:
+      count: 1
+      match:
+        engine.message: "Allowing absolute filename for dataset rule: C:\Windows\Temp\dns-seen-save.txt"
+  - filter:
+      count: 1
+      match:
+        engine.message: "Allowing absolute filename for dataset rule: C:\Windows\Temp\..\Temp\dns-seen.txt"
--- a/tests/datasets/datasets-absolute-allowed-post8-winonly/README.md
+++ b/tests/datasets/datasets-absolute-allowed-post8-winonly/README.md
@@ -1,0 +1,2 @@
+Test that the configuration option to allow absolute dataset filenames
+in rules works.
--- a/tests/datasets/datasets-absolute-allowed-post8-winonly/suricata.yaml
+++ b/tests/datasets/datasets-absolute-allowed-post8-winonly/suricata.yaml
@@ -1,0 +1,13 @@
+%YAML 1.1
+---
+
+datasets:
+  rules:
+    allow-absolute-filenames: true
+
+logging:
+  outputs:
+    - file:
+        enabled: yes
+        filename: eve.json
+        type: json
--- a/tests/datasets/datasets-absolute-allowed-post8-winonly/test.rules
+++ b/tests/datasets/datasets-absolute-allowed-post8-winonly/test.rules
@@ -1,0 +1,3 @@
+alert dns any any -> any any (dns.query; dataset: isnotset, dns-seen, type string, state C:\Windows\Temp\dns-seen.txt; sid:1; rev:1;)
+alert dns any any -> any any (dns.query; dataset: isnotset, dns-seen-save, type string, save C:\Windows\Temp\dns-seen-save.txt; sid:2; rev:1;)
+alert dns any any -> any any (dns.query; dataset: isnotset, dns-seen-parent, type string, state C:\Windows\Temp\..\Temp\dns-seen.txt; sid:3; rev:1;)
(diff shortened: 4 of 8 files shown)


norg:
The test might need adjustments for Windows MINGW builds if they don't have a `/tmp`folder

jasonish:
The test for < Suricata 8 should be suffixed with "-pre8". The test for 8 should not have a -pre or -post.

norg:
> The test for < Suricata 8 should be suffixed with "-pre8". The test for 8 should not have a -pre or -post.

Good point, will update this once I get the CI to fully pass

jasonish:
Just to summarize some conversation elsewhere. I don't think there is a way to make this work on Windows in a way that works on GitHub, my local Windows install or another CI system.

GitHub, about a month ago did something in their Windows environment that broke `/tmp/` from being used in tests. We used to simply drop the log files there while running unit tests (fixed with 11cef2980be3626153c82ebe3a2ca7485651dcd2).

However, `/tmp/` still works on my local Windows install. The only temp directory we might be able to construct a full path to on GitHub Windows might be something like `C:\\Users\githubrunner` (or whatever the username is), but that's not going to work outside of Github.

So at this point in time, I don't see a way to enable this test on Windows. In theory, we could use the S-V "script" feature to use a rule template file, output rule files with a localized hard coded path, then run the test. But I think using `script` in an S-V test automatically disabled it on Windows.

One other concern, we're not writing to `/tmp/` using a fixed path. If you have 2 S-V running simultaneously and hit this test at the same time, it could lead to failures. So maybe pre-processing the rule as described above is a good idea.

tests/datasets/datasets-absolute-allowed-post8-winonly/test.yaml
```diff
@@ -0,0 +1,22 @@
...
+  lambda: sys.platform == "win32"
+
+args:
+  - -vvv
+
+checks:
+  - filter:
+      count: 1
+      match:
+        engine.message: "Allowing absolute filename for dataset rule: C:\Windows\Temp\dns-seen.txt"
```
jasonish, line 14: I think these lines are messing up the YAML, `C:\\Windows\\Temp\\dns-seen.txt` works. In short, these tests aren't running on Windows.

jasonish: But with this fix, they are run and pass.

norg: So what is your suggestion in that case for the S-V on windows?

Switch to `\\` instead of `\` in the `engine.message` part?

Or even skip the test on Windows in general due to the nature of the absolute path being tricky on Windows?
(and also your comment about the pcap)

tests/datasets/datasets-absolute-allowed-post8-winonly/test.yaml
```diff
@@ -0,0 +1,22 @@
+pcap: ../../datasets/datasets-parent-path/one-packet.pcap
```
norg, line 1: @jasonish is this the `pcap not found` issue that you mentioned? Wouldn't this be a generic issue with S-V on windows and not just this specific test?

Since `../../` is used in other tests as well with subdir, like https://github.com/OISF/suricata-verify/blob/master/tests/alert-max/alert-max-20/test.yaml

tests/datasets/datasets-absolute-allowed-post8-winonly/test.yaml
victorjulien: lets not skip the test on windows

There are other tests that use `grep` to match on an error message, this could avoid having to be precise about the path

tests/datasets/datasets-absolute-allowed-post8-winonly/test.yaml
victorjulien: perhaps easiest to copy the pcap for now, we can clean up after the release

norg: Do the copy within the test logic or within the CI or manually on local tests?

I ask cause this would be an issue on quite an amount of existing S-V tests on Windows that also have `../../` in use

norg: I guess this error message is a bit misleading, I even had it locally on Linux for that test:

```
/home/andi/code/suricata-verify/run.py datasets-absolute-allowed
    Finished `release` profile [optimized] target(s) in 0.03s
Number of concurrent jobs: 8
while scanning a double-quoted scalar
  in "/home/andi/code/suricata-verify/tests/datasets/datasets-absolute-allowed-winonly/test.yaml", line 14, column 25
found unknown escape character 'W'
  in "/home/andi/code/suricata-verify/tests/datasets/datasets-absolute-allowed-winonly/test.yaml", line 14, column 74
===> datasets-absolute-allowed-pre8: SKIPPED: for version less than 8
===> datasets-absolute-allowed-winonly: SKIPPED: No pcap file found
===> datasets/datasets-absolute-allowed: OK

PASSED:  1
FAILED:  0
SKIPPED: 2
```

When I switched to the `grep` way of matching for that test it actually reaches the skip for the platform:

```
/home/andi/code/suricata-verify/run.py datasets-absolute-allowed
    Finished `release` profile [optimized] target(s) in 0.03s
Number of concurrent jobs: 8
===> datasets-absolute-allowed-winonly: SKIPPED: sys.platform == "win32"
===> datasets-absolute-allowed-pre8: SKIPPED: for version less than 8
===> datasets/datasets-absolute-allowed: OK

PASSED:  1
FAILED:  0
SKIPPED: 2
```
So the pcap error was misleading IMHO.

PR #2563, opened 2025-06-11
tests: remove exit-code check for datasets-absolute-allowed on v8

norg:

The new test is split between Unix like systems and Windows by using different absolute paths for the filenames.

## Ticket

changes to the ticket status:

### DIFF
--- a/tests/datasets/datasets-absolute-allowed-winonly/test.yaml
+++ b/tests/datasets/datasets-absolute-allowed-winonly/test.yaml
@@ -1,0 +1,19 @@
+pcap: ../../datasets/datasets-parent-path/one-packet.pcap
+
+requires:
+  min-version: 8
+  lambda: sys.platform == "win32"
+
+args:
+  - -vvv
+
+checks:
+  - shell:
+      args: grep "Allowing absolute filename for dataset rule" eve.json | grep "dns-seen.txt" | wc -l 
+      expect: 1
+  - shell:
+      args: grep "Allowing absolute filename for dataset rule" eve.json | grep "dns-seen-save.txt" | wc -l 
+      expect: 1
+  - shell:
+      args: grep "Allowing absolute filename for dataset rule" eve.json | grep "dns-seen2.txt" | wc -l 
+      expect: 1
--- a/tests/datasets/datasets-absolute-allowed-pre8/README.md
+++ b/tests/datasets/datasets-absolute-allowed-pre8/README.md
@@ -1,0 +1,2 @@
+Test that the configuration option to allow absolute dataset filenames
+in rules works.
--- a/tests/datasets/datasets-absolute-allowed-pre8/suricata.yaml
+++ b/tests/datasets/datasets-absolute-allowed-pre8/suricata.yaml
@@ -1,0 +1,13 @@
+%YAML 1.1
+---
+
+datasets:
+  rules:
+    allow-absolute-filenames: true
+
+logging:
+  outputs:
+    - file:
+        enabled: yes
+        filename: eve.json
+        type: json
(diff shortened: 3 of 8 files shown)


tests/datasets/datasets-absolute-allowed-winonly/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+requires:
+  min-version: 8
+  lambda: sys.platform == "win32"
+
+args:
+  - -vvv
+
+checks:
+  - shell:
```
victorjulien, line 11: put you on the wrong track here. On Windows shell checks are actually skipped.

tests/datasets/datasets-absolute-allowed-winonly/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+      args: grep "Allowing absolute filename for dataset rule" eve.json | grep "dns-seen.txt" | wc -l 
+      expect: 1
+  - shell:
+      args: grep "Allowing absolute filename for dataset rule" eve.json | grep "dns-seen-save.txt" | wc -l 
+      expect: 1
+  - shell:
+      args: grep "Allowing absolute filename for dataset rule" eve.json | grep "dns-seen2.txt" | wc -l 
+      expect: 1
```
victorjulien, line 19: seeing the output, but the loading of the sets fails, is that expected?
image

tests/datasets/datasets-absolute-allowed-winonly/test.yaml
norg: oh, so what alternative do we have on windows in that case?

tests/datasets/datasets-absolute-allowed-winonly/test.yaml
norg: so the `state` fails but the `save` does not. The relevant diff might be that `save` only tries to store on exit, since it's in memory during runtime. But `state` on the other hand might already require the file to be created there.

Is there actually a `C:\\Windows\\Temp` folder on that runner? If not, that would explain it and leads us back to the issue, that is tricky to find a common ground on which absolute folder to use on Windows.

victorjulien: yeah the folder exists and I can read and write my file. Perhaps only using `save` would work.

tests/datasets/datasets-absolute-allowed-winonly/test.yaml
victorjulien: maybe something like
```
engine.message.__startswith: "Lua script failed to run successfully"
```
`tests/lua/lua-blocked-function-1/test.yaml`

norg: But to properly check it for each rule I would still have to add the path (like Jason mentioned, we might need to use `\\` instead of `\`) or just check if 3 times the `Allowing absolute filename` appeared to avoid adding the Windows path.

tests/datasets/datasets-absolute-allowed-winonly/test.yaml
norg: So we just use `save` for the windows test?

tests/datasets/datasets-absolute-allowed-winonly/test.yaml
victorjulien: trying this https://github.com/OISF/suricata-verify/pull/2566/files#diff-c5e305d4413fb61397da08af7f6e61624d49f89a94d67f98cb872b2962e439a6R10

norg:
closed in favor of https://github.com/OISF/suricata-verify/pull/2566

PR #2566, opened 2025-06-12
tests: remove exit-code check for datasets-absolute-allowed on v8

victorjulien:

The new test is split between Unix like systems and Windows by using different absolute paths for the filenames.

## Ticket

changes to the ticket status:

Redmine ticket:


inashivb (approved):
LGTM 🚀

PR #2567, opened 2025-06-12
tests: add vxlan with non-zero reserved fields test

AmazingPP:

## Ticket


tests/vxlan-non-zero-reserved-fields/test.yaml
```diff
@@ -0,0 +1,17 @@
+args:
```
jufajardini, line 1: Since this change is for master, for now, let's add a `min-version: 8.0` requirement here. (as seen in https://github.com/OISF/suricata-verify/blob/master/tests/datajson/datajson-01-ip/test.yaml#L2)

tests/vxlan-non-zero-reserved-fields/test.yaml
```diff
@@ -0,0 +1,17 @@
...
+        event_type: flow
+        src_port: 51252
+        dest_port: 80
+        flow.pkts_toserver: 5
+        flow.pkts_toclient: 5
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        dest_port: 4789
```
jufajardini, line 17: Since this change is to ensure that the received packets are properly parsed by Suricata, I think we could have a check for the stats here that show that:
```yaml
- filter:
      count: 1
      match:
        event_type: stats
        stats.decoder.vxlan: 10
```
This will require a minimum suricata.yaml

jufajardini (changes requested):
Thanks for adding an SV test!

Minor changes inline.

Could you indicate in the readme file how did you generate the pcap file?

tests/vxlan-non-zero-reserved-fields/test.yaml
AmazingPP: OK.

PR #2574, opened 2025-06-17
tests: add vxlan with non-zero reserved fields test - v2

AmazingPP:
Replaces: #2567 

Changes since v1:
- Set the minimum Suricata version requirement to 8.
- Added a subtest to explicitly verify correct VXLAN decoding, by checking that a flow with destination port 80 is extracted from the inner packet.
- Expanded the PCAP section with details about its origin and construction rationale.

## Ticket


AmazingPP:
Replaced by https://github.com/OISF/suricata-verify/pull/2581

PR #2575, opened 2025-06-17
tests: remove wrong lt-version: 7

catenacyber:
## Ticket

### DIFF
--- a/run.py
+++ b/run.py
@@ -140,6 +140,9 @@ class UnsatisfiedRequirementError(Exception):
 class ImpossibleRequirementError(Exception):
     pass
 
+class UnnecessaryRequirementError(Exception):
+    pass
+
 class TerminatePoolError(Exception):
     pass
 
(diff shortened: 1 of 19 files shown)


jasonish (approved):
I think this is good stuff to add.

victorjulien (approved):

victorjulien:
exception-policy-applayer-03/test.yaml also has this issue, which tripped up my staging pipeline. It depends on a DEBUG build, so might be good to enable that on one set of CI jobs?

victorjulien (changes requested):
see inline

catenacyber:
> exception-policy-applayer-03/test.yaml also has this issue, which tripped up my staging pipeline. It depends on a DEBUG build, so might be good to enable that on one set of CI jobs?

Nice catch, next version in https://github.com/OISF/suricata-verify/pull/2582

PR #2576, opened 2025-06-17
test/byte-vars: Restrict var usage to single-buffer

jlucovsky:
2 new test cases
- bytemath-07 Test 05 for release 8 and later
- bytemath-08 Using additional "lol" signature

Issue: 1412

## Ticket

changes to the ticket status:

Redmine ticket:


catenacyber:
> additional "lol" signature

Wonder where it comes from :-p

tests/detect-bytemath-05/test.yaml
```diff
@@ -2,6 +2,7 @@ pcap: ../detect-bytemath-01/input.pcap
 
 requires:
   min-version: 7
+  lt-version: 8
```
catenacyber: nit: can be a simple `version: 7`

tests/detect-bytemath-07/test.rules
```diff
@@ -0,0 +1,5 @@
+alert tcp any any -> any any (msg:"byte_math varname test sig"; \
```
catenacyber, line 1: Readme linking to issue would be nice

We want to test somehow that the rule currently matches

jlucovsky: I'll add a README file.

The new tests ensure that the no longer supported cases are detected. There are existing tests that check for valid behavior. Could you elaborate on what additional tests are needed?

catenacyber: See https://redmine.openinfosecfoundation.org/issues/1412#note-7

I would like a SV test with the rule and pcap from https://redmine.openinfosecfoundation.org/issues/1412 where we see that the rule currently matches.

> There are existing tests that check for valid behavior

Suricata now works "by luck" on this case, maybe people do not expect their rules to break...
Did you investigate rulesets for such cases ?

tests/detect-bytemath-09/test.yaml
```diff
@@ -0,0 +1,13 @@
+requires:
+  min-version: 8
```
catenacyber: Only the shell check requires min-version 8, right ?

tests/detect-bytemath-07/test.rules
catenacyber: done in test 9 thanks

tests/detect-bytemath-05/test.yaml
```diff
@@ -1,7 +1,7 @@
 pcap: ../detect-bytemath-01/input.pcap
 
 requires:
-  min-version: 7
+  version: 7
```
catenacyber: Do we really need this change ?

jlucovsky: No; i'll revert.

tests/detect-bytemath-09/test.yaml
jlucovsky: Yes, correct.

tests/detect-bytemath-09/test.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - shell:
+     min-version: 8
+     args: grep "Warning. detect-byte. Using byte variable from a different buffer may produce indeterminate results; variable. \"content-length\"" suricata.log | wc -l | xargs
+     expect: 1
+
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 44412999
```
catenacyber, line 12: I would like a README, or a comment here ;-)

catenacyber:
Closing as stale like https://github.com/OISF/suricata/pull/13809

PR #2578, opened 2025-06-19
tests/lua: simply return of empty needs

jasonish:
Just "return {}". Makes it easier to extract what is actually being returned here for comparison with the documentation.


victorjulien (approved):

victorjulien:
Merged in #2584, thanks!

PR #2581, opened 2025-06-24
tests: add vxlan with non-zero reserved fields test - v3

AmazingPP:
Replaces: #2574 

Changes since v2:
- Added `decoder.vxlan.reserved-bits-check` configuration and set it to permissive mode.

Changes since v1:
- Set the minimum Suricata version requirement to 8.
- Added a subtest to explicitly verify correct VXLAN decoding, by checking that a flow with destination port 80 is extracted from the inner packet.
- Expanded the PCAP section with details about its origin and construction rationale.

## Ticket


tests/vxlan-non-zero-reserved-fields/README.md
```diff
@@ -0,0 +1,11 @@
+# Test Description
+
+Test VXLAN decoding with non-zero reserved fields.
+
+## PCAP
+
+Pcap provided by Fupeng Zhao. Manually crafted to reflect [vxlan: Support user-defined reserved bits](https://lwn.net/Articles/998601/).
```
catenacyber, line 7: Thanks for this link :-)

tests/vxlan-non-zero-reserved-fields/test.yaml
```diff
@@ -0,0 +1,25 @@
+requires:
+  min-version: 8
```
victorjulien, line 2: this will need to be updated to 9 now

PR #2582, opened 2025-06-25
Tests pre7 7461 v2

catenacyber:
## Ticket

#2575 with additional fix

### DIFF
--- a/tests/dns/dns-invalid-opcode/test.yaml
+++ b/tests/dns/dns-invalid-opcode/test.yaml
@@ -37,7 +37,6 @@ checks:
 # Generated checks below.
       
 - filter:
-    min-version: 8
     count: 1
     match:
       alert.action: allowed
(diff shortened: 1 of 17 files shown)


victorjulien (approved):

tests/dns/dns-invalid-opcode/test.yaml
```diff
@@ -72,43 +71,6 @@ checks:
       src_ip: 1.1.1.1
       src_port: 5333
       tx_id: 0
-
-- filter:
-    lt-version: 8
-    count: 1
-    match:
-      alert.action: allowed
-      alert.category: Generic Protocol Command Decode
```
inashivb, line 81: Q: Sure this test wasn't compatible with the main `min-version` requirement but why do we remove this instead of removing the main `min-version` requirement and adding a `min-version` to individual tests for 8 (or a new test altogether for 7)? This way we remove what should have been a valid test in 7.

catenacyber: There is already tests/dns/v2/dns-invalid-opcode right @jasonish ?

PR #2583, opened 2025-06-25
tests: update for exceptions stats counters for suricata 7 - v1

jufajardini:
Related to backport ticket
Task #6509

## Ticket

Redmine ticket:
https://redmine.openinfosecfoundation.org/issues/6509

### DIFF
--- a/tests/exception-policy-applayer-03/test.yaml
+++ b/tests/exception-policy-applayer-03/test.yaml
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  min-version: 7.0.11
   features:
     - DEBUG
 pcap: ../bittorrent-dht/input.pcap
@@ -66,6 +66,7 @@ checks:
       src_ip: 190.0.0.2
       src_port: 50000
 - filter:
+    min-version: 7.0.11
     count: 1
     match:
       event_type: stats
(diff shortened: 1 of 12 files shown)


tests/exception-policy-applayer-03/test.yaml
```diff
@@ -66,6 +66,7 @@ checks:
       src_ip: 190.0.0.2
       src_port: 50000
 - filter:
+    min-version: 7.0.11
```
catenacyber, line 69: This is wrong, the test itself has min-version 8 see https://github.com/OISF/suricata-verify/pull/2582 ;-p

catenacyber (changes requested):
Needs at least one fix

Nice to see how this kind of error could creep in

tests/exception-policy-applayer-03/test.yaml
jufajardini: 🙇🏽

jufajardini: I honestly thought, somehow, in my head, that a sub-check would overwrite that master `min-version` requirement :x

catenacyber:
Will likely need a rebase after merge of #2582

tests/exception-policy-applayer-03/test.yaml
catenacyber: @jufajardini this is currently failing : 
===> exception-policy-applayer-03: FAILED: Internal runtime error: test already requires min 7.0.11 not needed for the check 7.0.11

PR #2586, opened 2025-07-03
tests: add coverage for pcap-file.delete-non-alerts-only behavior

oferda4:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7786?next_issue_id=7785


tests/pcap-file-delete-non-alerts-only-no-alerts/empty.rules
```diff
@@ -0,0 +1 @@
+
```
catenacyber, line 1: you can just remove this file ;-)

tests/pcap-file-delete-non-alerts-only-no-alerts/suricata.yaml
```diff
@@ -0,0 +1,7 @@
+%YAML 1.1
+---
+
+include: ../alert-testmyids/default.yaml
+
+pcap-file:
```
catenacyber, line 6: test.yaml can also have an args : `--set pcap-file.delete-non-alerts-only = true` (or something like this) if you prefer

tests/pcap-file-delete-non-alerts-only-no-alerts/test.yaml
```diff
@@ -0,0 +1,21 @@
+setup:
+  - script: cp ../input_origin.pcap ../input.pcap
+
+pcap: false
+
+command: |
+  ${SRCDIR}/src/suricata --pcap-file-delete -k none \
```
catenacyber, line 7: Can `--pcap-file-delete` be simply added to args ?

PR #2588, opened 2025-07-08
test/entropy: Update checks to include sid

jlucovsky:
The sid is part of the log -- e.g., sticky-buffer_sid -- so update the test checks for each sid that alerts.

## Ticket

changes to the ticket status:

Redmine ticket:


catenacyber:

tests/entropy/entropy-01/test.yaml
```diff
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  version: 8
```
catenacyber: Why this change ? Will this not work in 9 ?

tests/entropy/entropy-02/test.rules
```diff
@@ -0,0 +1,12 @@
...
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value > 4; sid:1;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value >= 4; sid:2;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value >= 5; sid:3;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value 4-5; sid:4;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value !4-5; sid:5;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value < 4; sid:6;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value != 4; sid:7;)
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: value = 4; sid:8;)
+# The entropy value is 4.150007324019584
+alert http any any -> any any (msg:"entropy simple test"; file.data; entropy: offset 10, value > 4.14; sid:10;)
```
catenacyber, line 11: Not sure I understand the ticket here, but
- a flow may have multiple txs
- a tx may have multiple files (for example I could write a signature to detect a HTTP POST with 2 files : one having a low entropy, and the other file having a high entropy)

So, is it good to combine sid and sticky buffer name ?

jlucovsky: The entropy value is saved as a flow-related variable -- using `VarNameStorageRegister` -- into a global namespace. The `sid` is being added to "scope" the name to a rule.

I think you're saying that there needs to be further scoping when entropy is evaluated over multiple transactions?

catenacyber: Yes, I fear flow+sticky buffer+sid is not unique enough

inashivb: Great analysis.
I suppose we could use a combination of `file.md5`  and the `entropy` to make it unique?

catenacyber: But I am not sure I totally get the point : do you just want to have logged with an alert the entropy(s) that matched ?

jlucovsky: Yes -- the idea is to log the calculated entropy value -- not only in the alert but in all places where the flow is logged.

Consider a rule with mult. entropy statements --- we'd like to know the calculated entropy values of each so they can be adjusted if needed

catenacyber: > not only in the alert but in all places where the flow is logged.

It does not make much sense to me to log in a flow event, some value that belongs to a transaction

And even then, each transaction may have multi-buffers (like dns requests) and you would want to log these different values...

I feel this needs more design, instead of reusing what is available in the engine even if that is not fit...

jufajardini: Considering this affects output, I tend to agree with Jeff. Could we maybe take some time in the next dev meeting for this, or open a thread to discuss and propose?

jlucovsky: I've updated the entropy PR to use <sticky_buffer>_<sid>_<occurrence> where <occurrence> tracks the entropy keyword usage; where 1 is for the first occurrence, 2 for the 2nd, and so on.

catenacyber: Does it work for a packet having multiple transactions ?

tests/entropy/entropy-01/test.yaml
```diff
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  version: 8.0.0
```
catenacyber, line 2: Why this change ? Should it not work in 8.0.1 ?

tests/entropy/entropy-01/test.yaml
jlucovsky: Good catch -- the test will work for 8.0.0+ so I'll update things.

tests/entropy/entropy-02/test.yaml
```diff
@@ -0,0 +1,89 @@
+requires:
+  min-version: 8.0.1
```
catenacyber: Needs update... 9 ?

tests/entropy/entropy-02/test.yaml
```diff
@@ -0,0 +1,94 @@
...
+  - --set logging.entropy.make-unique=on
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 1
+        metadata.entropy[0].__ignore_type: "{'sid:1;buffer:file_data;instance:1': 4.137370175000773}"
+        metadata.entropy[8].__ignore_type: "{'sid:10;buffer:file_data;instance:1': 4.150007324019584}"
```
jasonish: Can we just use the object here instead of `__ignore_type`, for example: https://github.com/OISF/suricata-verify/pull/2665/files#diff-01fcb3749573dcf5e88b90390ba73a6b8fe3b38f4677e2df07026e2f52ddb198

tests/entropy/entropy-01/test.yaml
catenacyber: ping @jlucovsky

PR #2589, opened 2025-07-09
requires: update test to match on version 8.0.1

inashivb:


jasonish:
> @jasonish is this ok or would you want to update the rule?

We should change the rule to not change the intention of the test.

New PR: https://github.com/OISF/suricata-verify/pull/2592

inashivb:
> > @jasonish is this ok or would you want to update the rule?
> 
> We should change the rule to not change the intention of the test.
> 
> New PR: #2592

That sounds better indeed. Thank you!

PR #2594, opened 2025-07-10
Use treading instead of processes: for Windows support

jasonish:
- **runner: raise UnsatisfiedRequirementError for windows shell checks**
  Instead of just printing that shell checks are not supported on
  Windows, properly skip the test so it is counted as skipped.
  

- **runner: remove os.chdir, use full paths as needed**
  In preparation for multi-threading, where we can't chdir as that would
  affect other threads.
  

- **runner: convert to multi-threading from multi-processing**
  Multi-processing has issues on Windows, moving to threading does allow
  it to work on Windows (with a few issues to still figure out).
  
  This removes the single thread runner, instead for Windows and Mac
  we'll just use one job for now, and change that default as we prove
  its reliaable.
  
  Update ctrl-c handling as well, for reliable ctrl-c cancellation.
  

- **github-ci: add windows build**


jasonish:
> Ah ya with `--fail` it works, but it does not show the summary anymore.
> 
> ```
> victor@cubi150u UCRT64 ~/suricata-ucrt
> $ time python3 ../suricata-verify/run.py -j 12 --fail -q
> Number of concurrent jobs: 12
> ===> datasets-absolute-allowed-winonly: FAILED: got exit code 1, expected 0
> 
> Shutting down executor...
> 
> real    0m32.698s
> user    0m0.000s
> sys     0m0.031s
> 
> victor@cubi150u UCRT64 ~/suricata-ucrt
> $ echo $?
> 1
> ```

Fixed with latest push.

run.py
```diff
@@ -1167,9 +1196,12 @@ def main():
                         help="Clean up output directories of passing tests")
     parser.add_argument("--no-validation", action="store_true", help="Disable EVE validation")
     parser.add_argument("patterns", nargs="*", default=[])
-    if MP:
-        parser.add_argument("-j", type=int, default=min(8, mp.cpu_count()),
-                        help="Number of jobs to run")
+    if WIN32 or DARWIN:
+        parser.add_argument("-j", type=int, default=1,
```
victorjulien: I feel this is not needed. We should enable it by default and let CI run it here and elsewhere. We'll find out quickly enough if there are issues.

jasonish: Removed, same logic for Linux across Windows and Mac now.

PR #2587, opened 2025-07-06
tests: add coverage for new delete-when-done value in configuration

oferda4:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7786?next_issue_id=7785

Previous PR: https://github.com/OISF/suricata-verify/pull/2586
Changes:
   - Fit to new changes in feature
   - Cleanup
   - Fix setup stage

PR #2596, opened 2025-07-14
tests: add coverage for new delete-when-done value in configuration

oferda4:
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7786?next_issue_id=7785

Previous PR: https://github.com/OISF/suricata-verify/pull/2587
Changes:
   - Use proper output dir.
   - Run only from version 8.


catenacyber (approved):
Test looks good

tests/pcap-file-delete-non-alerts-only-no-alerts/test.yaml
```diff
@@ -0,0 +1,23 @@
+requires:
+  min-version: 8
```
catenacyber, line 2: Now we will want version 9

PR #2599, opened 2025-07-15
tcp: fast open detection on first packet when inline

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6744


catenacyber:
Interestingly, this passes on 7

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/2805 :-)

PR #2604, opened 2025-07-18
snmp: add test when snmp is disabled

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7820

### DIFF
--- a/tests/snmp-disabled/test.yaml
+++ b/tests/snmp-disabled/test.yaml
@@ -1,0 +1,11 @@
+args:
+  - -k none
+
+pcap: ../snmp-v2c-get/SNMPv2c_get_requests.pcap
+
+checks:
+ - filter:
+     count: 0
+     match:
+       event_type: flow
+       app_proto: snmp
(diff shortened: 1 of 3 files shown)


tests/snmp-disabled/test.yaml
```diff
@@ -0,0 +1,10 @@
+args:
+  - -k none
+
+pcap: ../snmp-v2c-get/SNMPv2c_get_requests.pcap
+
+checks:
+ - filter:
+     count: 0
+     match:
+       event_type: snmp
```
inashivb: I think this test needs rework. It passes even when snmp is enabled.

We should either check for `flow.app_proto` or add `snmp` to the `eve-log.types` section.

inashivb (changes requested):
inline

tests/snmp-disabled/test.yaml
catenacyber: Thanks Shivani, fixed by patch
```diff
diff --git a/tests/snmp-disabled/suricata.yaml b/tests/snmp-disabled/suricata.yaml
index e51cf720..272413a7 100644
--- a/tests/snmp-disabled/suricata.yaml
+++ b/tests/snmp-disabled/suricata.yaml
@@ -9,6 +9,7 @@ outputs:
       types:
         - flow
         - alert
+        - snmp
```

PR #2606, opened 2025-07-19
Mime type reload

regit:
Add tests for new mime type computation and match.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7816


tests/ftp-mimetype-detection/test.yaml
```diff
@@ -0,0 +1,19 @@
+requires:
+  features:
+      - HAVE_MIMETYPE
```
catenacyber, line 3: @jasonish seeing green CI for this highlights that we need SV coverage ;-)

jasonish: Please elaborate? What would this look like?

catenacyber: You seemed to have more ideas than me.

I was just saying that manual runs should be done for releases to ensure that every SV test/check is run.
You were saying that there were tools to do that ;-)

jasonish: No tools that I know of.  I think Victor put out a wish directed my way that tools exist, to which I had no comment :)

tests/ftp-mimetype-detection/test.yaml
```diff
@@ -0,0 +1,19 @@
...
+# disables checksum verification
+args:
+- -k none --runmode=single --set outputs.1.eve-log.types.6.files.force-mimetype=yes
+
+checks:
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 1338
```
catenacyber, line 14: nit: we could check that the alert event has the `mimetype `

PR #2607, opened 2025-07-21
pppoe: Add test for pppoe decoder protocol alerts

thomasjwinter:
Suricata commit b23fa51e ("detect: fix decoder only events") fixed decoder events to go off properly. However it was found that the pppoe decoder was going off on valid ppp packets.
With drop rules isntead of alert rules, a ppp connection could not be established.

The following pcap is a valid ppp connection but valid pppoe packets are being detected as wrong_type or unsup_proto.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7651


tests/pppoe/test.yaml
```diff
@@ -0,0 +1,5 @@
+checks:
```
catenacyber, line 4: Thanks, could you add a min-version 8 requires ?

Process is first have master being ok with this SV, then we can have a backport PR for Suricata, and a SV PR that just decreases the min-version

catenacyber: You solved this, thanks

catenacyber (approved):
Test looks good to me

victorjulien:
Merged in #2643, thanks!

PR #2642, opened 2025-09-08
tests/etag: Basic test case for ETAG decode

AmazingPP:
I'm not sure which version this feature needs to be integrated in, so the minimum version is temporarily set to 8.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3953


catenacyber:
> I'm not sure which version this feature needs to be integrated in, so the minimum version is temporarily set to 8.
> 

That was the good thing to do for now

AmazingPP:
Replaced by https://github.com/OISF/suricata-verify/pull/2650

PR #2644, opened 2025-09-09
dns: add test with rrtype keyword and index

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7480

Need to test more keywords especially with regard to tx progress


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/2653

PR #2645, opened 2025-09-09
pppoe: backport test for pppoe decoder protocol alerts

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7826

### DIFF
--- a/tests/pppoe/test.yaml
+++ b/tests/pppoe/test.yaml
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  min-version: 7.0.12
 
 checks:
   - filter:


tests/pppoe/test.yaml
```diff
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  min-version: 7.0.13
```
victorjulien: should be 7.0.12

victorjulien (approved):

tests/pppoe/test.yaml
catenacyber: oops, fixed ;-)

PR #2646, opened 2025-09-09
Stream tests/v9

victorjulien:
## Ticket

changes to the ticket status:

Redmine ticket:

### DIFF
--- a/tests/bug-4702-01/test.yaml
+++ b/tests/bug-4702-01/test.yaml
@@ -1,5 +1,5 @@
 requires:
-  min-version: 8
+  min-version: 7.0.12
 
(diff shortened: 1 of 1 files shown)


tests/bug-4702-02/test.yaml
```diff
@@ -0,0 +1,17 @@
+requires:
+  min-version: 7
```
lukashino, line 2: Does this test pass already from 7 or from 7.0.12 only?

jufajardini: A while ago, I recall it being said that it was preferable to have these updated up to the minor version.

PR #2650, opened 2025-09-17
tests/etag: Basic test case for ETAG decode v2

AmazingPP:
Replaces: https://github.com/OISF/suricata-verify/pull/2642

Changes since v1:

- Added the redmine ticket in commit message.

I'm not sure which version this feature needs to be integrated in, so the minimum version is temporarily set to 8.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/3953


catenacyber:
Will soon be 9 : https://github.com/OISF/suricata/pull/13867 ;-)

tests/decode-etag-01/test.yaml
```diff
@@ -0,0 +1,12 @@
+requires:
+   min-version: 8
```
catenacyber, line 2: Now 9 is ready for SV PRs

PR #2651, opened 2025-09-17
tests: support 9 in requires-ok test

victorjulien:
## Ticket

changes to the ticket status:

Redmine ticket:

### DIFF
--- a/tests/requires-ok/test.yaml
+++ b/tests/requires-ok/test.yaml
@@ -5,10 +5,25 @@ pcap: ../eve-metadata/testmyids.pcap
 
 checks:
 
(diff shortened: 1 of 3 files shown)


tests/requires-ok/test.yaml
```diff
@@ -5,10 +5,25 @@ pcap: ../eve-metadata/testmyids.pcap
 
 checks:
 
+  # Check that for Suricata >= 9.0.0 we have one alert and that alert is sid 900.
```
catenacyber, line 8: "one alert" but the check is `count: 2` :-p

catenacyber: Because we also have one alert for sid 8 as expected

PR #2668, opened 2025-09-29
mime: adds test about quoted-printable

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7950

### DIFF
--- a/tests/smtp-file-data-02/test.yaml
+++ b/tests/smtp-file-data-02/test.yaml
@@ -45,7 +45,7 @@ checks:
       event_type: alert
       files[0].filename: NEWS.txt
       files[0].gaps: false
-      files[0].size: 10809
+      # files[0].size: 10807 TODO min-version 9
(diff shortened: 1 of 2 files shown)


tests/smtp-file-data-02/test.yaml
```diff
@@ -45,7 +45,7 @@ checks:
       event_type: alert
       files[0].filename: NEWS.txt
       files[0].gaps: false
-      files[0].size: 10809
+      # files[0].size: 10807 TODO min-version 9
```
victorjulien, line 48: when should this be addressed?

catenacyber: When finishing the backport to 7, this part of the check should be reinstated with the correct value

(instead of splitting the check in 2 based on version9 or below, then version 8 or below, then reuniting)

victorjulien (approved):

PR #2673, opened 2025-10-03
test/unknown: Use host order for ethtype check

jlucovsky:
Issue: 7855

Match the ethertype value using host order.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7855

### DIFF
--- a/tests/decode-unknown-2/test.yaml
+++ b/tests/decode-unknown-2/test.yaml
@@ -1,5 +1,3 @@
-requires:
-    min-version: 8
 
 args:
 - -k none
@@ -16,6 +14,16 @@ checks:
       decoder.unknown_ethertype: 1
   - filter:
       count: 1
+      min-version: 9
+      match:
+        event_type: anomaly
+        ether.ether_type: 64439
+        anomaly.type: decode
+        anomaly.event: decoder.ethernet.unknown_ethertype
+  - filter:
+      count: 1
+      min-version: 8
+      lt-version: 9
       match:
         event_type: anomaly
         ether.ether_type: 47099


jasonish:
Some version requirements needed until backport?

tests/decode-unknown-2/test.yaml
```diff
@@ -16,6 +14,15 @@ checks:
...
       count: 1
+      min-version: 9
+      match:
+        event_type: anomaly
+        ether.ether_type: 64439
+        anomaly.type: decode
+        anomaly.event: decoder.ethernet.unknown_ethertype
+  - filter:
+      count: 1
+      version: 8
```
inashivb: I'm not sure if this is correct. This will look exactly for version `8.0.0` and skip the test if not found so, this subtest will be skipped against `main-8.0.x` every time now.

jasonish (changes requested):
Version for 8 needs a fix. Probably `lt-version: 9` is best?

tests/decode-unknown-2/test.yaml
jlucovsky: Good catch. Thanks

tests/decode-unknown-2/test.yaml
```diff
@@ -1,5 +1,3 @@
-requires:
-    min-version: 8
```
jasonish, line 2: This means that the test will now also run on 7.0.x, are the checks that are not bound to a specific version OK for 7.0.x?

jasonish: So they do pass on 7. Probably OK to keep as-is then.

PR #2838, opened 2026-01-03
tests: add geoip-enrichment tests

ghost:
Add tests for the geoip-enrichment output feature

- geoip-enrichment: Tests that geoip fields are added to alert and flow events when
geoip-enrichment is enabled

- geoip-enrichment-disabled: Tests that NO events have geoip fields when geoip-enrichment is disabled

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6999


tests/geoip-enrichment/test.yaml
```diff
@@ -0,0 +1,25 @@
...
+        has-key: geoip_dst.ip
+        has-key: geoip_dst.geo.country_iso_code
+        geoip_dst.geo.country_iso_code: DE
+
+  # Check flow has geoip_dst enrichment
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        has-key: geoip_dst
```
jufajardini, line 24: nit: since we have the check below, this `has-key` isn't needed

jufajardini: (same for the previous check)

tests/geoip-enrichment-disabled/test.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+pcap: input.pcap
+
+checks:
+  # Verify NO alerts have geoip_dst when enrichment is disabled
+  - filter:
+      count: 0
+      match:
+        event_type: alert
+        has-key: geoip_dst
```
jufajardini, line 13: You can use `not-has-key`, instead.

jufajardini (changes requested):
Thanks for adding tests :)

We must also require a `min-version: 9.0` check, to ensure they're only tested against `main`.

In addition to the inline comments, could we adjust the test so that it would show all of the GeoIP enrichment that is being added? When I ran it locally, I only got:
```json
  "geoip_dst": {
    "ip": "82.165.177.154",
    "geo": {
      "country_iso_code": "DE"
    }
  },
```

jayhardee9:
Thank you for the quick feedback @jufajardini! When I submit my changes, do I open a new PR or can I push them to this one?

PR #2864, opened 2026-01-16
run.py: check for pattern on test's parent dir

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/2858

Changes since v2:
- instead of parent of basename, get parent from tests dir

### DIFF
--- a/run.py
+++ b/run.py
@@ -1247,15 +1247,16 @@ def main():
         if dirpath == tdir:
             continue
         basename = os.path.basename(dirpath)
+        parent = dirpath.replace(tdir, "")
         if args.skip_tests:
             skip_tests_opt = False
             patterns = args.skip_tests.split(",")
             for pattern in patterns:
                 if args.exact:
-                    if pattern == basename:
+                    if pattern == parent or pattern == basename:
                         skip_tests_opt = True
                         break
-                elif basename.find(pattern) > -1:
+                elif parent.find(pattern) > -1:
                     skip_tests_opt = True
                     break
             if skip_tests_opt:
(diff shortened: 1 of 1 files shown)


run.py
```diff
@@ -1277,6 +1277,11 @@ def main():
                         tests.append(dirpath)
                 elif basename.find(pattern) > -1:
                     tests.append(dirpath)
+                else:
```
catenacyber: can this just replace the `basename.find(pattern)` ? So I can run suricata-verify/run.py dcerpc/dcerpc-dce-iface

Also nit : strip prefix instead of replace ?

Also same logic needs to upgraded a few lines above for skip

catenacyber:
Thanks for doing this

run.py
inashivb: > can this just replace the basename.find(pattern) ? So I can run suricata-verify/run.py dcerpc/dcerpc-dce-iface

great idea!

ok to everything else 👍🏽

run.py
```diff
@@ -1247,15 +1247,16 @@ def main():
         if dirpath == tdir:
             continue
         basename = os.path.basename(dirpath)
+        parent = dirpath.replace(tdir, "")
```
catenacyber, line 1250: Strip prefix dirty like `parent = dirpath[len(tdir):]` ?

run.py
```diff
@@ -1247,15 +1247,16 @@ def main():
...
         if args.skip_tests:
             skip_tests_opt = False
             patterns = args.skip_tests.split(",")
             for pattern in patterns:
                 if args.exact:
-                    if pattern == basename:
+                    if pattern == parent or pattern == basename:
```
catenacyber, line 1256: I am not sure we should change the `exact` logic with parent name

jasonish (approved):
Behaviour seems fine to me.

victorjulien (approved):

victorjulien:
Merged in #3031, thanks!

PR #2894, opened 2026-01-29
http1: adds test with trailer

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8256

Passes with https://github.com/OISF/suricata/pull/14710


catenacyber:
Updated the test to show that we should alert only once per tx

PR #2896, opened 2026-01-31
workflows: only run the modified tests

inashivb:
Previous PR: https://github.com/OISF/suricata-verify/pull/2745

Changes since v5:
- Review comments addressed

Ticket: None. Should there be one?

Draft because: Yet to be tested.

Known TODOs:

- Fail if any of the changed tests are skipped (as follow up work? I have something.)

### DIFF
--- a/.github/workflows/builds.yml
+++ b/.github/workflows/builds.yml
@@ -9,7 +9,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  
+
   pcapng-check:
     name: PCAP Check
     runs-on: ubuntu-22.04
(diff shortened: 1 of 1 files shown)


inashivb:
Also need to present this better so it looks useful as Philippe did here: https://github.com/OISF/suricata-verify/pull/2745#issuecomment-3816434840 🙇🏽‍♀️

catenacyber (approved):
Cool :-)

.github/workflows/builds.yml
```diff
@@ -18,10 +18,51 @@ jobs:
...
+                if [[ $dir != tests/* ]]; then
+                    echo "$dir is not tests/"
+                    only_tests=0
+                    break
+                fi
+            done
+
+            if [ $only_tests == 1 ]; then
+                # Get the names of the innermost test dir
+                testnames=$(echo "$dirnames" | xargs -I {} basename "{}" | tr '\n' ' ')
```
catenacyber, line 47: Could you explain me the magic `xargs -I {} basename "{}"` ?

inashivb: I use this commonly. I understand it as xargs picking up as args that were the result of the cmd that was piped into it and sending them as args to the cmd that xargs is going to execute

inashivb: Is it wrong?

inashivb:
> Why a draft ?

Still gotta test it against my fork. I'm sure otherwise, I'd get a rejection from you if it was not Draft :P

.github/workflows/builds.yml
catenacyber: It looks to do the right thing

What is the difference with `xargs basename` ?

inashivb: Here it probably won't matter but I always use -I because it allows to put the args anywhere not just at the end. I can test and change it if you want

catenacyber: > I always use -I 

few lines above...

> `xargs dirname`

:-p

inashivb: :sob:

.github/workflows/builds.yml
```diff
@@ -18,10 +18,45 @@ jobs:
...
+          case '${{ github.event_name }}' in
+          pull_request)
+            firstCommit=${{ github.event.pull_request.base.sha }}
+            lastCommit=${{ github.event.pull_request.head.sha }}
+            # Get names of all dirs changed in a PR
+            dirnames=$(git diff --name-only "${firstCommit}" "${lastCommit}" | xargs dirname | sort -u)
+            echo "Dirnames: $dirnames"
+
+            # check iff tests/ dir has changed
+            only_tests=$(ls -la | grep -v "tests/*" | wc -l)
```
catenacyber, line 43: This seems wrong now, not checking `dirnames`

```
            # check iff tests/ dir has changed
            only_tests=1
            for dir in $dirnames; do
                if [[ $dir != tests/* ]]; then
                    echo "$dir is not tests/"
                    only_tests=0
                    break
                fi
            done

            if [ $only_tests == 1 ]; then
```

was correct, was it not ?

catenacyber: Maybe a small python script would be easier...

inashivb: sorry. this was WIP, was supposed to push it in another branch 🤦🏽‍♀️

.github/workflows/builds.yml
catenacyber: -I is needed to handle multi-line input !!!

PR #2906, opened 2026-02-09
Review tools v1

catenacyber:
## Ticket

No ticket, just sharing some tools to batch reviews


catenacyber:
I will close this draft in a month if there is not much interest...

catenacyber:
This is not specific to SV

PR #2909, opened 2026-02-11
tests: some overlapping ipv4 fragments are reassembled with a data hole

laubard:
Related to https://redmine.openinfosecfoundation.org/issues/6668 and https://redmine.openinfosecfoundation.org/issues/6673 issues


tests/ipv4-defrag-overlaps/test_407/test.yaml
```diff
@@ -0,0 +1,9 @@
+# *** Add configuration here ***
+
+args:
+- -k none
+
+checks:
+- shell:
+    args: grep '"payload_printable":".*\.\.\.\.\.\.\.\..*"' eve.json | wc -l | xargs
```
catenacyber, line 11: Why do you need a `shell` check instead of a `filter` one ?

jasonish:
Thanks @laubard. Due to some time off I probably won't get to full review along with the code for a week or so. In the mean time, could you add a `min-version: 9`, and perhaps a short README.md to each test, add a bit of detail as to what its showing? Thanks!

jufajardini (changes requested):
Changing the status as per Jason's comment.

laubard:
@jasonish I added a readme for each test as well as a `min-version`

catenacyber:
@laubard are you planning to rework this to get green CI ?

catenacyber:
Jason told me that CI is meant to be red now, and that we should create a tracking ticket for this work

laubard:
@catenacyber Is there something I need to do on my side?

catenacyber:
I guess you could create redmine ticket(s) to explain the problem(s)

PR #2932, opened 2026-02-21
tests: verify pcap_filename on alert and flow events - v2

oferda4:
Add a regression test that checks the pcap_filename field is present and correct in EVE JSON output for both alert events (packet-based) and flow events (non-packet-based).

These two code paths use different sources for the filename in OutputJsonBuilderBuffer: per-packet pfv->filename when a packet is available, and the global PcapFileGetFilename() fallback for flow/netflow events where no packet exists.

Previous PR: https://github.com/OISF/suricata-verify/pull/2927

v2:
    - Fix recursive test and readme.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5255
Suricata PR: https://github.com/OISF/suricata/pull/15116


tests/pcap-filename-alert-and-flow/test.yaml
```diff
@@ -0,0 +1,20 @@
...
+pcap: ../bug-7414-decoder-event-01/ip_secopt.pcap
+
+checks:
+  # Alert event - exercises the p != NULL path in OutputJsonBuilderBuffer
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 2200005
+        pcap_filename.__endswith: "ip_secopt.pcap"
```
catenacyber, line 13: What else could we get since there is only one pcap in this test ?

oferda4: Not sure I understand what you mean. We want to test the field shows the correct value. If you mean as why using `_endswith` it's because the field contains the absolute path.

catenacyber: Test pcap-filename-recursive has 2 packs, so I see we can be wrong for pcap_filename

But how could we be wrong when there is only one pcap like in this test ?

oferda4: This test is just a sanity the filename is properly added. Do you think the test is redundant?

catenacyber: Maybe indeed. What does it cover that pcap-filename-recursive  does not cover ?

oferda4: So regarding code paths it doesn't add any more coverage. It's just giving a simpler test for the basic functionality (that works even from older versions) to be easier to debug rather the complex recursive test. 
But I don't mind to remove the test entirely if we are trying to be more strict on tests usage.

catenacyber: Ok, I would say let's keep it, thanks

tests/pcap-filename-recursive/test.yaml
```diff
@@ -0,0 +1,36 @@
...
+      count: 1
+      match:
+        event_type: fileinfo
+        pcap_filename.__endswith: "http.pcap"
+
+  - filter:
+      count: 1
+      match:
+        event_type: flow
+        pcap_filename.__endswith: "http.pcap"
```
catenacyber, line 36: Do we want the stats event as well ?

tests/pcap-filename-recursive/test.yaml
```diff
@@ -0,0 +1,36 @@
+requires:
+  min-version: 9
+
+pcap: .
+
+args:
+- --pcap-file-recursive
+
+checks:
+  # Race-condition regression: alert.pcap is processed FIRST; http.pcap is
```
catenacyber, line 10: Why is alert processed first ? Alphabetical order ?

oferda4: The code in `source-pcap-file-directory-helper.c` sorts by modification time (CompareTimes), not alphabetically. After a git checkout, both alert.pcap and http.pcap get the same mtime (I verified they're identical), so the order is actually determined by readdir() - which is filesystem-dependent and not guaranteed.
                                                                                                                                                                                                            
However, the test assertions are order-independent: they just verify that each event has the correct pcap_filename regardless of which pcap was processed first. The comment here is misleading because it claims a specific processing order. I'll update the comment.

tests/pcap-filename-recursive/test.yaml
oferda4: Good idea - I've added stats coverage to the single-pcap test (pcap-filename-alert-and-flow). Stats events hit the p == NULL, f == NULL global fallback path in OutputJsonBuilderBuffer, which was the only path without test coverage. With one pcap the global is unambiguous so it's a clean verification.

PR #2943, opened 2026-03-02
tests: add ssh test for issue 8333

jasonish:
https://redmine.openinfosecfoundation.org/issues/8333


jasonish:
Fix looks simple, I think this is very similar to iprep. Should be able to do a firewall rule test as well and mark it SIGMATCH_SUPPORT_FIREWALL?

PR #2968, opened 2026-03-13
detect/byte: add cross-buffer byte variable tests

jlucovsky:
Continuation of #2497 

Multiple test cases checking that the byte_ variables can be used in a cross-buffer scenario.

Add 8 tests for per-transaction byte value caching (issue #7801).

Same-direction tests (01-04) verify cross-buffer byte_extract usage
within a single direction (e.g., http.header.raw -> file_data).

Bidirectional tests (01-04) verify byte_extract and byte_math across
direction boundaries using bidirectional (=>) rules with pipelined
HTTP/1.1 requests. A spoiler TX clobbers det_ctx->byte_values before
the first TX's toclient inspection, validating the per-TX cache.

Pipelined multi-TX tests (05-08) cover additional cache scenarios:
spoiler-first ordering, spoiler-second ordering, false positive
prevention with isdataat, and byte_math result caching.

  Issue: 7801

Updates:
- Moved from issue 1412 to issue 7801

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7801


tests/bug-7801/bug-7801-01/gen_pcap.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
```
catenacyber, line 1: Could we have the test from https://redmine.openinfosecfoundation.org/issues/1412 ?

jlucovsky: Will add.

catenacyber: Thanks for adding

catenacyber (approved):
Looks like a good complete suite of tests :-)

PR #2978, opened 2026-03-20
Flowbits tests/v3

inashivb:
Previous PR: #2936

Changes since v2:
- 5 more tests added

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7638

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,12 @@ checks:
         # Check if a string is contained within a JSON list.
         # Eg. "ftp":{"reply":["Opening BINARY mode data connection for temp.txt (1164 bytes).","Transfer complete."], }
         ftp.reply.__contains: 'Transfer complete.'
-
+  - filter:
+      # Check line number of an entry
+      count: 1
+      match:
+        __lineno: 5
+        alert.signature_id: 15
   - shell:
       # A simple shell check. If the command exits with a non-0 exit code the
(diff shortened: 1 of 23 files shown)


tests/test-ruleparse-etopen-01/emerging-all.rules
```diff
@@ -1430,7 +1430,7 @@ alert tcp $HOME_NET any -> $EXTERNAL_NET 1024: (msg:"ET MALWARE Perfect Keylogge
 
 alert tcp $HOME_NET any -> $EXTERNAL_NET $HTTP_PORTS (msg:"ET MALWARE Socks666 Connection Initial Packet"; flow:established,to_server; dsize:24; content:"|9a 02 06 00|"; offset:0; depth:4; flowbits:set,BS.BPcheckin; flowbits:noalert; classtype:trojan-activity; sid:2006395; rev:5; metadata:created_at 2010_07_30, updated_at 2019_07_26;)
```
inashivb, line 1431: Note there's a rule setting the flowbit first that is independent of the rules w cyclic dependencies. It should be evaluated sooner.
Should sid:2006396 and sid:2006397 still be considered cyclic?

Edit: I think having a rule that sets the flowbit should make it easier to define the order and as long as the set rule is active, these should not be considered cyclic dependencies.

PR #2982, opened 2026-03-22
tests/absent+pcrexform: Test cases for error_or/must_error coverage

jlucovsky:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7114

Add test cases for new absent options -- must_error and error_or that covers cross-buffer matching, validation rejection and functional behavior. Add must_error/error_or cases for pcrexform (another transform that can fail).


tests/detect-pcrexform-08/test.rules
```diff
@@ -0,0 +1,8 @@
+# pcrexform does not match -> error_or matches on error
+alert http any any -> any any (msg:"error_or pcrexform no match"; http.request_line; pcrexform:"(XXXNOMATCHXXX)"; absent: error_or; content:"nomatch"; sid:1;)
```
catenacyber, line 2: Why does this match ? 

`content:"nomatch";` should not match

`error_or; content:"/z4d4kWk.jpg";` should match

catenacyber: sorry my bad

tests/rules/absent-05/test.rules
```diff
@@ -0,0 +1,2 @@
+# error_or on a buffer with a transform that cannot fail should fail to load
+alert dns any any -> any any (msg:"error_or with non-failing transform"; dns.query; dotprefix; absent: error_or; content:".example.com"; sid:1;)
```
catenacyber: actually, dotprefix could fail if realloc failed

I would take strip_whitespace ;-)

tests/rules/absent-08/test.yaml
```diff
@@ -0,0 +1,15 @@
+requires:
+  min-version: 8
+
+pcap: ../absent-04/empty.pcap
+
+args:
+  - --init-errors-fatal
+
+checks:
+  # or_else with non-failing transform (dotprefix) should load without errors
```
catenacyber: Should it really ?

PR #2941, opened 2026-03-02
tests: add ja3 tests for issue 8336

jasonish:
https://redmine.openinfosecfoundation.org/issues/8336

PR #2983, opened 2026-03-22
tests: add ja3 tests for issue 8336

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8336

#2941 rebased with addition of transactional bidirectional signatures, ips tests, typo fix...


jasonish (approved):

catenacyber:
> Wonder if we should also have ssh-frames test with variants ips and encryption-handling=bypass

Test ssh-frames passes with current PR even if I add `--set app-layer.protocols.tls.encryption-handling=bypass` or `--simulate-ips`

victorjulien (approved):

victorjulien:
Merged in #3092, thanks!

PR #2992, opened 2026-03-28
tests: remove unneeded flowbit ops

inashivb:


inashivb:
> Why do we want to remove these ?

@catenacyber cleanup? They serve no purpose as they are never checked.

catenacyber (approved):
Ok for me

tests/datasets/datasets-lua-01/dataset-lua.rules
```diff
@@ -1 +1 @@
-alert tcp any any -> any any (flow:to_server; flowbits:isnotset,dataset_added; lua:dataset.lua; flowbits:set,dataset_added;sid:1;)
```
victorjulien, line 1: This would be a pattern I'd recommend to avoid the cost of calling into lua. Seems like a good idea to test it this way?

inashivb: ok. Thank you.

PR #2993, opened 2026-03-28
test/xor: Add xor test cases using a key

jlucovsky:
Issue: 7847

Add test cases for xor
- That specify an offset
- That specify a byte_extract variable

## Ticket

changes to the ticket status:


tests/detect-xor-key/detect-xor-key-01/test.yaml
```diff
@@ -0,0 +1,11 @@
+min-version: 9
```
catenacyber, line 1: directories are named bug-7847 but this is a feature, I would rather have a name like detext-xor-key-01

jlucovsky:
Continued in #3102

PR #3013, opened 2026-04-07
tests: verify pcap_filename on alert and flow events - v3

oferda4:
Add a regression test that checks the pcap_filename field is present and correct in EVE JSON output for both alert events (packet-based) and flow events (non-packet-based).

These two code paths use different sources for the filename in OutputJsonBuilderBuffer: per-packet pfv->filename when a packet is available, and the global PcapFileGetFilename() fallback for flow/netflow events where no packet exists.

Previous PR: https://github.com/OISF/suricata-verify/pull/2932

v2:
    - Fix recursive test and readme.

v3:             
  - Add stats event coverage to pcap-filename-alert-and-flow (exercises the p == NULL, f == NULL global fallback path in OutputJsonBuilderBuffer)                                 
  - Fix misleading comment in pcap-filename-recursive about processing order; order is by mtime, not guaranteed lphabetical.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/5255
Suricata PR: https://github.com/OISF/suricata/pull/15741


tests/pcap-filename-recursive/README.md
```diff
@@ -0,0 +1,17 @@
+Verify that pcap_filename in EVE JSON output reflects the file each
+connection's packets came from, not the last file the RX thread processed.
+
+Two pcap files are fed with --pcap-file-recursive.  alert.pcap is processed
```
catenacyber: You said

> the order is actually determined by readdir() - which is filesystem-dependent and not guaranteed.

So ? (or can we make the order reproducible)

catenacyber (approved):
Besides the nit comment, I think these tests are good

tests/pcap-filename-recursive/README.md
oferda4: You are right but the test outcome is deterministic:  regardless of order, one pcap's flow events would get the wrong pcap_filename if the global fallback were used instead of the per-flow pfv. I'll update the README to not claim a specific order 👍

oferda4:
> Besides the nit comment, I think these tests are good

@catenacyber How do you go with small fixes like that? Should I open a new PR version?

inashivb:
> > Besides the nit comment, I think these tests are good
> 
> @catenacyber How do you go with small fixes like that? Should I open a new PR version?

It's ok to force-push minor changes on suricata-verify PRs. Thanks for asking and following contribution guidelines! 🙇🏽‍♀️ 🌟

PR #3016, opened 2026-04-08
framework: improve error msg for unknown exception - v1

jufajardini:
If a parsing error happens in the `test.yaml` file, indicate which test cause it, and which sub-check in the test.yaml file, to be more user-friendly.

## Ticket

None yet, should I create one?

### DIFF
--- a/run.py
+++ b/run.py
@@ -189,7 +189,8 @@ def handle_exceptions(func):
                 print("===> {}: Sub test #{}: SKIPPED : {}".format(kwargs["test_name"], kwargs["test_num"], ue))
             kwargs["count"]["skipped"] += 1
         except Exception as err:
-            raise TestError("Internal runtime error: {}".format(err))
+            test_path = os.path.relpath(args[0].directory, args[0].cwd)
+            raise TestError("test.yaml parsing error: \"{}\" in test {} at check #{}".format(err, test_path, kwargs["test_num"]))
         else:
             if result:
               kwargs["count"]["success"] += 1


jasonish:
> If a parsing error happens in the test.yaml 

If we fail to parse the test.yaml, we won't make it to any check. What kind of parsing error are we trying to report on here?

jufajardini:
> > If a parsing error happens in the test.yaml
> 
> If we fail to parse the test.yaml, we won't make it to any check. What kind of parsing error are we trying to report on here?

Errors such as:
- yaml indentation issues -- leads to internal runtime error -- NoneType is not iterable (happens way too often for me, and the first few times it took me a while to figure out what and where the issue was)
- something like: `count: 1  2`  -- haven't really had this issue, but tried this one out, and also works for having a more meaningful error message.

jasonish:
> > > If a parsing error happens in the test.yaml
> > 
> > 
> > If we fail to parse the test.yaml, we won't make it to any check. What kind of parsing error are we trying to report on here?
> 
> Errors such as:
> 
> * yaml indentation issues -- leads to internal runtime error -- NoneType is not iterable (happens way too often for me, and the first few times it took me a while to figure out what and where the issue was)
> * something like: `count: 1  2`  -- haven't really had this issue, but tried this one out, and also works for having a more meaningful error message.

Ok. The the YAML is parsed all up front in one shot, before any test is run.  So I don't think the check # makes sense.

But I do see that an invalid type, like a NoneType error should hit this.  So the message is wrong. The YAML parsed successfully if we get here, just some code working with the parsed test.yaml is not setup to handle what was presented.. Internal error still makes sense, but adding the check # can then help you narrow it down.

jufajardini:
> > > > If a parsing error happens in the test.yaml
> > > 
> > > 
> > > If we fail to parse the test.yaml, we won't make it to any check. What kind of parsing error are we trying to report on here?
> > 
> > 
> > Errors such as:
> > 
> > * yaml indentation issues -- leads to internal runtime error -- NoneType is not iterable (happens way too often for me, and the first few times it took me a while to figure out what and where the issue was)
> > * something like: `count: 1  2`  -- haven't really had this issue, but tried this one out, and also works for having a more meaningful error message.
> 
> Ok. The the YAML is parsed all up front in one shot, before any test is run. So I don't think the check # makes sense.
> 
> But I do see that an invalid type, like a NoneType error should hit this. So the message is wrong. The YAML parsed successfully if we get here, just some code working with the parsed test.yaml is not setup to handle what was presented.. Internal error still makes sense, but adding the check # can then help you narrow it down.

It helps, to me, because if we're running many tests, we don't even get a clear idea of which test has failed. 
How can I improve this?
- Reword the part about being a parsing error
- Is there more?

catenacyber (changes requested):
There were some changes requested

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/3161

PR #3019, opened 2026-04-09
live: a new live testing harness

jasonish:
Where Suricata is run in AFP inline, NFQ inline, or IDS mode (using a Linux bridge) monitoring traffic between real clients and servers.

Just tossing this out there for early review.  

Some extra notes:
- I used an AI coding agent to help me with some of the tediousness of a test harness, but it's been heavily supervised
- I purposefully did not integrate into the existing test harness as I feel its complicated enough already, but some code like the checks was ported over.
- I'll likely circle back and pull out common code like the "checks" and "requires" (not yet supported here) into a common lib used by both harnesses
- I think @catenacyber used `live` as the directory for some other tests, so happy to hear about alternative nameing for these tests
- Heavily based on https://github.com/jasonish/suricata-ips-testing which is heavily based on the in-tree netns tests by @victorjulien 

It should work in GitHub CI, but I'm curious if anyone has issues running these in their own development environment. Podman is required.


live/tests/firewall-icmp/README.md
```diff
@@ -0,0 +1,13 @@
+---
+tags:
+- firewall
+- icmp
+---
```
jasonish, line 5: This allows for a `--tag firewall` option to run.py so you can run tests that match a tag. If multiple tags are provided, its an AND (a test must have all tags).

TODO: Document.

live/tests/firewall-icmp/test.yaml
```diff
@@ -0,0 +1,38 @@
...
+
+  podman run --cap-add=NET_RAW --rm --network ns:/var/run/netns/client ${id} bash -lc '
+    if ping -c 1 10.200.0.1; then
+        echo "error: ping should have failed" > /dev/stderr
+        exit 1
+    fi
+  '
+
+  cp firewall2.rules ${OUTDIR}/firewall.rules
+  ${SURICATASC} -c reload-rules ${OUTDIR}/suricata.socket
```
jasonish, line 17: The client does have access to do this.  In theory the server side does as well, but the client is where you have some "scripting" ability.

live/run.py
```diff
@@ -0,0 +1,1589 @@
...
+        )
+
+    return parser
+
+
+def need_root() -> None:
+    if os.geteuid() == 0:
+        return
+    print(
+        "ERROR: this script must be run as root. Use: sudo python3 run.py ...",
```
jasonish:
```suggestion
        "ERROR: this script must be run as root. Use: sudo run.py ...",
```

live/run.py
```diff
@@ -0,0 +1,1589 @@
...
+    if not m:
+        return None
+
+    major = int(m.group(1)) if m.group(1) else default_v
+    minor = int(m.group(2)) if m.group(2) else default_v
+    patch = int(m.group(3)) if m.group(3) else default_v
+    return SuricataVersion(major=major, minor=minor, patch=patch)
+
+
+class Version:
```
jasonish, line 993: Example of something duplicated from the main `run.py` that should be moved to a common lib.

.github/workflows/live.yaml
```diff
@@ -0,0 +1,167 @@
...
+        run: |
+          ./autogen.sh
+          ./configure --enable-nfqueue --enable-debug
+          make -j$(nproc)
+
+      - run: |
+          ../live/run.py run
+        working-directory: suricata
+
+  environments:
```
jasonish: This job can go, early testing just to figure out what worked in GitHub CI.

live/tests/simple-http/test.yaml
```diff
@@ -0,0 +1,20 @@
+# This test will run in NFQ, AF-PACKET and IDS modes.
+mode: [nfq, afp, ids]
+
+server: |
+  id=$(podman build -q --network=host server)
+  podman run --rm \
+    --network ns:/var/run/netns/server \
+    ${id}
+
```
jasonish, line 9: Podman is not required for the server, but then it would depend on whats installed on the host. We could safely assume Python 3.? is installed with its standard library which might be enough to implement some custom servers (and clients), but using Podman lets us use Dockerfile's for our servers and clients.

live/run.py
```diff
@@ -0,0 +1,1600 @@
...
+NFQ_SERVER_GW = "10.200.0.254"
+NFQ_QUEUE_NUM = "0"
+
+ALL_NAMESPACES = (CLIENT_NS, SERVER_NS, DUT_NS)
+ROOT_LINKS = (TMP_CLIENT_IF, TMP_SERVER_IF, TMP_DUT_CLIENT_IF, TMP_DUT_SERVER_IF)
+MODES = ("afp", "ids", "nfq")
+
+MODE_LABELS = {
+    "ids": "IDS",
+    "afp": "AF_PACKET",
```
jasonish, line 60: Mabye we should have `afp-ids` and `afp-tap`?

live/run.py
```diff
@@ -0,0 +1,1600 @@
...
+        need_cmd(cmd)
+
+    if action == "up":
+        UP_FUNCS[mode]()
+    elif action == "down":
+        do_down()
+    elif action == "status":
+        STATUS_FUNCS[mode]()
+    elif action == "shell":
+        do_shell(args.target)
```
jasonish, line 1731: TODO: This run.py can also be used to bring up and ids, nfq or afp environment and interactively work with it. Not sure if that functionality is needed here, at least not like this.  `suricata-ips-testing` is a little easier to use.

catenacyber:
So, what is the difference with https://github.com/OISF/suricata-verify/pull/2969 ?

live/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+This directory contains a test harness and tests for live tests. A
+live test is where Suricata is running in an IDS or IPS mode
+monitoring traffic between a real client and server in an isolated
+network environment.
+
+## Supported Modes
+
+- IDS: Runs Suricata in AF_PACKET IDS mode an interface that can see
+  traffic between the client and server namesapces. Internally a Linux
```
catenacyber: typo `namesapces`

live/README.md
```diff
@@ -0,0 +1,30 @@
+# Suricata-Verify Live Tests
+
+This directory contains a test harness and tests for live tests. A
+live test is where Suricata is running in an IDS or IPS mode
+monitoring traffic between a real client and server in an isolated
+network environment.
+
```
catenacyber, line 7: Is the goal to check all three outputs ?
- suricata
- client
- server

(otherwise, why cannot we tcpreplay a pcap ? )

jasonish:
> So, what is the difference with #2969 ?

Replay tests are good for IDS modes, but I don't think they're very good inline/ips modes.  When Suricata drops a packet in a replay, the replay keeps going. When actually inline, the client and/or server need to react to that, and we need to test and verify that we also react accordingly to their replays, etc.

live/README.md
jasonish: In some cases, it might be.  In others, like the ICMP test, it might just be good enough for the client to exit with an exit code.  For exampple if we know a curl to a specific URL should be blocked, but it happens to go through anyways.

jasonish:
@victorjulien I ported afp-fw-netns-bridge.sh to "afp-fw" if you could take a look.

Some notes though. This harness doesn't yet support multiple tpacket versions in the same test. I wonder if that makes sense. I guess it could do it like a matrix..

```yaml
tpacket-version: 2
```
```yaml
tpacket-version: [2,3]
```
(which would then do the complete test for each version)

The same would apply to workers mode.

Alternatively we allow these options, but only one per test and duplicated some tests.

jasonish:
This fails in CI here:
```
===> [afp/afp-fw]: FAIL
```
cause it depends on caddy being available on the host system. This was intentional for now to show a direct port for an existing. The fix will be to wrap it up in a Dockerfile and run with Podman which I'll do soon.

live/tests/afp-fw/test.yaml
```diff
@@ -0,0 +1,92 @@
+mode: [afp]
+
+before: |
+  cp ${TESTDIR}/firewall1.rules ${OUTDIR}/firewall.rules
+
+server: |
```
jasonish, line 6: There's nothing special about this being named `server`, as it's not run in the namespace for you. A better name might be `background`.

Alternatively, in the `client` script, you could background a server in the server namespace and then continue with your client calls.

live/tests/afp-fw/test.yaml
```diff
@@ -0,0 +1,92 @@
...
+  ip netns exec server tshark -i server -T json > ${OUTDIR}/tshark-server.json &
+  ip netns exec server caddy file-server browse
+
+client: |
```
jasonish, line 10: Again, nothing special with `client` here, maybe simply `script` is better here. This is what controls the life cycle of this test.  When `client` exits, the server script is terminated, then Suricata is shutdown.

live/run.py
victorjulien: seems to make sense. We'd want to test pcap with this as well eventually I think

live/run.py
victorjulien: I suppose it could be useful for debugging a failing test. Almost like creating a "breakpoint" when you can drop into the shell?

live/run.py
jasonish: I don't really see how pcap makes sense here.

jasonish:
Superseded by #3115 as the second version of this live-test runner work.

PR #3022, opened 2026-04-09
test(firewall): Add verify tests for untested firewall keywords

yashda:
Add suricata-verify tests for 7 keywords that emit 'has not been tested for firewall rules' warnings. Tests are consolidated into 4 test cases:

- firewall-keyword-icode: tests icode with ICMP echo traffic
- firewall-keyword-http: tests pcre, urilen, dataset with HTTP traffic
- firewall-keyword-dns: tests dns.opcode, datarep with DNS traffic
- firewall-keyword-tls: tests tls.cert_chain_len with TLS cert chain

These tests validate that the keywords function correctly in firewall mode and can be used to justify adding SIGMATCH_SUPPORT_FIREWALL to each keyword in the engine.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8387

### DIFF
--- a/tests/firewall/firewall-keyword-dns/test.yaml
+++ b/tests/firewall/firewall-keyword-dns/test.yaml
@@ -1,0 +1,33 @@
+requires:
+  min-version: 8
+
+pcap: ../../dns/dns-eve/input.pcap
+
+args:
+  - --simulate-ips
+  - -k none
+
+checks:
+# dns.opcode:0 matches all 4 DNS requests (all are standard queries)
+- filter:
+    count: 4
+    match:
+      event_type: alert
+      alert.signature_id: 1
+# datarep matches 3 queries (client-cf.dropbox.com x2 + block.dropbox.com)
+- filter:
+    count: 3
+    match:
+      event_type: alert
+      alert.signature_id: 2
+- filter:
+    count: 0
+    match:
+      event_type: drop
+- filter:
+    count: 1
+    match:
+      event_type: stats
+      stats.ips.accepted: 7
+      stats.ips.blocked: 1
+      stats.ips.drop_reason.default_packet_policy: 1
--- a/tests/firewall/firewall-keyword-dns/dns_rep.rep
+++ b/tests/firewall/firewall-keyword-dns/dns_rep.rep
@@ -1,0 +1,2 @@
+Y2xpZW50LWNmLmRyb3Bib3guY29t,255
+YmxvY2suZHJvcGJveC5jb20=,255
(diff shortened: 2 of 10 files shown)


tests/firewall/firewall-keyword-dns/test.yaml
```diff
@@ -0,0 +1,33 @@
+requires:
+  min-version: 8
```
jufajardini: Should we do as usual, and first enable these for 9 only, and once we're backporting the Warnings removal to 8, we enabled these, for 8, too?

jufajardini (changes requested):
Thanks for this work!

Some mostly minor comments:
- could you add tests for drops, too, not just accept?
- for the commit message to fully comply, we need to reduce the size of the commit title to be within 50 characters. I would say that something like:
`tests: check for untested firewall keywords` could work. Adding the ticket number as the last info in the commit body is also welcome! :)
- could you rename the tests themselves to refer to the keywords being tested? This makes them more self explanatory, at a glance
- normally, I would expect that we would first require that the tests pass for `min-version: 9`, but in this case, since we expect them to also work for 8, I'm guessing it's ok to keep it at 8, as the tests should pass, too.

jufajardini:
> Thanks for this work!
> 
> Some mostly minor comments:
> 

- [ ] >     * could you add tests for drops, too, not just accept?
- [x] >     * for the commit message to fully comply, we need to reduce the size of the commit title to be within 50 characters. I would say that something like:
>       `tests: check for untested firewall keywords` could work. Adding the ticket number as the last info in the commit body is also welcome! :)
- [x] >     * could you rename the tests themselves to refer to the keywords being tested? This makes them more self explanatory, at a glance
- [x] >     * normally, I would expect that we would first require that the tests pass for `min-version: 9`, but in this case, since we expect them to also work for 8, I'm guessing it's ok to keep it at 8, as the tests should pass, too.

Hi again! I'll work on the items checked above, so we can merge this quickly. If you could then submit a follow-up pr with the `drop` cases for these, that'd be great :)

`tls.cert_chain_len` has tests added with other PRs, so maybe you could review if these are missing something from your use cases, and if not, no need to add more for these.

jufajardini:
So, here is a consolidated version of this work and one with more rules for `tls.cert_chain_len`: https://github.com/OISF/suricata-verify/pull/3025

catenacyber:
So closing in favor of https://github.com/OISF/suricata-verify/pull/3032 and next versions, right ?

PR #3030, opened 2026-04-16
tests: add ntp tests - v1

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/8394
Ticket: https://redmine.openinfosecfoundation.org/issues/8488
Ticket: https://redmine.openinfosecfoundation.org/issues/8429
Ticket: https://redmine.openinfosecfoundation.org/issues/8431
Ticket: https://redmine.openinfosecfoundation.org/issues/8430


tests/firewall/ruletype-firewall-56-ntp/test.yaml
```diff
@@ -0,0 +1,72 @@
...
+#     count: 1
+#     match:
+#       event_type: alert
+#       alert.signature_id: 13
+# - filter:
+#     count: 4
+#     match:
+#       event_type: alert
+#       alert.signature_id: 22
+# - filter:
```
victorjulien: think it would be good to check stats as well

Additionally, can we have a drop case as well, so not accept a certain condition and see the drop happens as expected?

tests/firewall/ruletype-firewall-56-ntp/test.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+  - filter:
+      count: 1
+      match:
+        event_type: drop
+        pcap_cnt: 4
+
+  - stats:
+      ips.accepted: 3
+      ips.blocked: 1
```
jasonish, line 51: Added these to check for drops.  We don't have a rule to allow version 3 responses with "LOCL" refid, so these don't make it through.  Version 4 responses with "LOCL" are allowed. Does this do the job.

Oops, forced push here.

victorjulien (approved):

victorjulien:
Merged in #3038, thanks!

PR #3025, opened 2026-04-13
tests: check for more ruletypes in firewall mode - v3

jufajardini:
A follow-up Consolidating
- #3024 and
- #3022 

With improvements to:
- reduce noise;
- make test names more informative;
- and adhere to our contribution guidelines and workflow.

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/

https://redmine.openinfosecfoundation.org/issues/8387

PR #3032, opened 2026-04-17
tests: check for more ruletypes in firewall mode - v4

jufajardini:
Follow-up of https://github.com/OISF/suricata-verify/pull/3025

- Rebased, adjusting naming conflict after the last additions to firewalls tests
- adjusted dns tests, that now need a different dns hook to work, (since the Helper function registers the progress, now), and that now can't be in the same test to have alerts for both (as they're using the same dns hook for different keywords, and thus either I got alert for one, or for the other).

## Ticket

https://redmine.openinfosecfoundation.org/issues/8387

### DIFF
--- a/tests/firewall/ruletype-firewall-56-dns-opcode/firewall.rules
+++ b/tests/firewall/ruletype-firewall-56-dns-opcode/firewall.rules
@@ -1,0 +1,12 @@
+# Test dns.opcode keyword in firewall mode
+# dns-eve PCAP: 4 DNS queries (3 dropbox, 1 codemonkey.net), all standard queries (opcode 0)
+
+# Accept all UDP packets
+accept:hook udp:all any any -> any any (sid:100;)
+
+# Test dns.opcode: match standard query (opcode 0) and alert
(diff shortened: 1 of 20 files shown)


tests/firewall/ruletype-firewall-56-dns-opcode/firewall.rules
```diff
@@ -0,0 +1,12 @@
+# Test dns.opcode keyword in firewall mode
+# dns-eve PCAP: 4 DNS queries (3 dropbox, 1 codemonkey.net), all standard queries (opcode 0)
+
+# Accept all UDP packets
+accept:hook udp:all any any -> any any (sid:100;)
+
+# Test dns.opcode: match standard query (opcode 0) and alert
+accept:hook dns:request_complete any any -> any any (dns.opcode:0; alert; sid:1;)
```
victorjulien, line 8: this ruleset exposes a bug: the absence of a `dns:request_started` rule should have lead to a default drop, but didn't.

catenacyber:
Seems replaced by https://github.com/OISF/suricata-verify/pull/3039 right ?

PR #3037, opened 2026-04-20
test: add ftp too many tx comment test

alinse-pltzr:
# changes
test for: https://github.com/OISF/suricata/pull/15213/

test if  suricata.yaml loads properly

Im not quite sure what to test exactly, since the addded rule uses an event, which doesn't exist yet (see: https://redmine.openinfosecfoundation.org/issues/8489)


catenacyber:
I do not understand the purpose of this test

catenacyber:
> Im not quite sure what to test exactly, since the addded rule uses an event, which doesn't exist yet (see: https://redmine.openinfosecfoundation.org/issues/8470)

The test should contain a pcap that creates many txs (you can add a suricata.yaml to lower the max-tx limit) and the test should also use the rule put in the suricata PR to check that it triggers

catenacyber (changes requested):
See above

alinse-pltzr:
> > Im not quite sure what to test exactly, since the addded rule uses an event, which doesn't exist yet (see: https://redmine.openinfosecfoundation.org/issues/8470)
> 
> The test should contain a pcap that creates many txs (you can add a suricata.yaml to lower the max-tx limit) and the test should also use the rule put in the suricata PR to check that it triggers

ok, thanks for the clarity :)  I adjusted the test accordingly.

tests/ftp-max-tx-rule/test.yaml
```diff
@@ -0,0 +1,9 @@
+args:
```
jufajardini, line 4: We must add a:

```
requires:
   min-version: 9
```

To ensure that the CI checks pass for versions less than 9, so we could have the test merged.

tests/ftp-max-tx-rule/suricata.yaml
```diff
@@ -0,0 +1,6 @@
+%YAML 1.1
+---
+
+app-layer:
+  ftp:
+    max-tx: 2
```
catenacyber, line 6: Why should this pcap create more than 3 txs ?

alinse-pltzr: I ve been asuming that a pair of a request and a response are one transaction, because of this part of the documentation: https://docs.suricata.io/en/latest/devguide/extending/app-layer/transactions.html#general-concepts

counting the pairs of requests and responses in my pcap, there are more than 3 tx in my understanding.

what did I get wrong, and whats the correct defintion of a tx?

catenacyber: max-tx is about the maximum live transactions.

Once we see the response (and have run detection and logging on it), the transaction gets freed and does not longer count as a live translation

alinse-pltzr: ok thanks, this makes sense, i ll generate a pcap with open live transactions

catenacyber:
Is this PR superseded by https://github.com/OISF/suricata-verify/pull/3041 ? cc @jlucovsky

jlucovsky:
> Is this PR superseded by #3041 ? cc @jlucovsky

Yes

catenacyber:
So closing in favor of #3041

PR #3041, opened 2026-04-22
tests: add test cases for bug 8489

jlucovsky:
Issue: 8489

Three test cases:

- bug-8489-01: two commands with max-tx=1 (limit exceeded) raises the too_many_transactions anomaly and fires the corresponding alert.
- bug-8489-02: six commands under max-tx=10 raise no anomaly and no alert (negative regression).
- bug-8489-03: burst of commands exceeding the limit followed by a new command after a server response — verifies the flow keeps parsing after the event fires, so later commands are still logged.

Each test has a response_command_too_long rule alongside the too_many_transactions rule so the to_client FTP parsing path stays active; without a to_client app-layer-event signature Suricata skips response parsing and the max-tx check does not fire.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8489


tests/bug-8489/bug-8489-01/ftp-events.rules
```diff
@@ -0,0 +1,5 @@
+# response rule is required to keep Suricata's to_client FTP parsing active
+# so responses complete pending transactions; without it the max-tx check
+# does not fire in the to_server direction.
```
jufajardini, line 3: I wonder if running `inline` would make it possible to alert on sid 3 without sid 2...

jufajardini (commented):
It looks good to me -- I did miss README files to explain the `.syn` and `Makefile`  files, but overall, from the explanation, it looks good.

I labeled it as requiring a fix as I didn't see a Suricata PR yet.

victorjulien (approved):

victorjulien:
Merged in #3073, thanks!

PR #3042, opened 2026-04-23
detect-bytetest-02: fix alert count

inashivb:
There's just one packet in the pcap that matches the rules 4 and 5, there's no need for 2 alerts. However, with the way packet-stream rules are assessed by Suricata, the rules are inspected twice by packet and stream payload inspection engines thus creating 2 alerts.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8266

### DIFF
--- a/tests/detect-bytetest-02/test.yaml
+++ b/tests/detect-bytetest-02/test.yaml
@@ -1,5 +1,5 @@
 args:
-- --set stream.midstream=true
+  - --set stream.midstream=true
 
 checks:
 - filter:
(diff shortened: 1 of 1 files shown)


catenacyber:
There is not yet a suricata PR for this, right ?

inashivb:
> There is not yet a suricata PR for this, right ?

Indeed. In fact, it's looking like we may need to update some other tests to have the same behavior because it's hard to change how Suricata operates.

inashivb:
Closing for now.

PR #3052, opened 2026-04-27
Quality: Potential panic on non-UTF8 paths

tomaioo:
## Summary

Quality: Potential panic on non-UTF8 paths

## Problem

**Severity**: `Medium` | **File**: `eve-validator/src/main.rs:L50`

Converting paths with `instance.to_str().unwrap()` will panic if a path is not valid UTF-8, terminating validation unexpectedly.

## Solution

Avoid `unwrap()` here; use `instance.display()` for messages and pass `instance` directly to `File::open`.

## Changes

- `eve-validator/src/main.rs` (modified)


catenacyber:
> will panic if a path is not valid UTF-8

I think we want that, enforcing paths are valid utf-8

PR #3055, opened 2026-04-28
Detect dcerpc rust 8391 v3

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8391

Just moving some unit tests to SV : so tests pass already, but we need this when we merge the removal of unit tests in suricata PR (coming)

PR #2749, opened 2025-11-09
test/subslice: Add test cases for subslice

jlucovsky:
Issue: 7672

Add test cases for the subslice transform

## Ticket

If your pull request is related to a Suricata ticket, please provide
the full URL to the ticket here so this pull request can monitor
changes to the ticket status:

Redmine ticket: 7672

PR #3056, opened 2026-04-28
next/sv/1320/20260428/v1

victorjulien:
Staging:
 - #2749
 - #3055


tests/dcerpc/dcerpc-dce-stub-data-03/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+    match:
+      event_type: alert
+      alert.signature_id: 1
+checks:
+- filter:
+    count: 1
+    match:
+      event_type: alert
+      alert.signature_id: 1
+      pcap_cnt: 4
```
jasonish, line 16: Oops, duplicate `checks`. Only one is used, likely the last one. This is in the source PR as well.

catenacyber: No they are not duplicate : we check that we have only one alert, and that it happens at packet 4

Do you see a way to do that with only one check ?

catenacyber: Oh but `checks` is duplicated indeed

catenacyber: `git grep checks: | cut -d: -f1 | uniq -c | awk '$1 > 1' | grep test.yaml`

victorjulien: can we have CI check to avoid this? Something like `$(grep -c "^checks:" test.yaml) -gt 1`

catenacyber: I fixed the original PR

catenacyber: I will try something like https://github.com/yaml/pyyaml/issues/165 but in a new PR

victorjulien: I can accept this PR and we can do a follow up PR to fix this up. Or we can rerun staging. Either way is fine.

catenacyber: let's accept and fixup later

victorjulien: > let's accept and fixup later

Works for me. Need an approval then :)

catenacyber (approved):
Ok, to be fixed up in https://github.com/OISF/suricata-verify/pull/3057

PR #3074, opened 2026-05-09
dhcp: verify option 52 overload parsing

ssam18:
This adds an integration test for issue 8538, the request to make Suricata parse DHCP options carried inside the BOOTP sname or file fields when Option Overload (52) is set. The pcap captures two parallel flows with the same xid, one benign and one where the OFFER and ACK hide dns_servers, routers and domain inside an overloaded sname while keeping the inline options block looking innocuous. The test asserts that the EVE dhcp events for the overloaded OFFER and ACK now expose those values, and that the benign flow in the same capture still reports its inline values unchanged. This is the verify side companion to OISF/suricata#15340.


jasonish:
I'm worried this pcap from the ticket is very synthetic. Do you have any real pcaps showing this option being used?

ssam18:
> I'm worried this pcap from the ticket is very synthetic. Do you have any real pcaps showing this option being used?

Yes, it's synthetic. Option 52 overload is rare enough in the wild that I don't have a production capture either, but the Wireshark sample collection has `PRIV_bootp-both_overload.pcap` (https://wiki.wireshark.org/SampleCaptures), which has been the canonical reference capture for this feature for years.

I would suggest adding it as a second test directory rather than swapping out my current pcap. The Wireshark sample covers overload value 3 (both sname and file overloaded) on a single packet; my current synthetic covers overload value 2 (sname only) on an OFFER/ACK pair plus a benign control flow. Different overload modes, complementary coverage. Happy to just swap it instead if you would prefer the smaller test footprint.

catenacyber (approved):
Looks like a good SV test

jasonish:
> > I'm worried this pcap from the ticket is very synthetic. Do you have any real pcaps showing this option being used?
> 
> Yes, it's synthetic. Option 52 overload is rare enough in the wild that I don't have a production capture either, but the Wireshark sample collection has `PRIV_bootp-both_overload.pcap` (https://wiki.wireshark.org/SampleCaptures), which has been the canonical reference capture for this feature for years.
> 
> I would suggest adding it as a second test directory rather than swapping out my current pcap. The Wireshark sample covers overload value 3 (both sname and file overloaded) on a single packet; my current synthetic covers overload value 2 (sname only) on an OFFER/ACK pair plus a benign control flow. Different overload modes, complementary coverage. Happy to just swap it instead if you would prefer the smaller test footprint.

An additional test would be great.

jasonish (approved):

victorjulien (approved):

victorjulien:
Merged in #3149, thanks!

PR #3077, opened 2026-05-10
Backport8 8457 v1

catenacyber:
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8522

Backport for suricata branch: backport8-8457-v1

PR #3078, opened 2026-05-11
next/sv/1339/80x/20260510/v1

victorjulien:
Staging:
 - #3077


jufajardini (approved):
Consistent with original, approved PR (even if, as Shivani pointed out, the diff with "all changed" looks weird).

jasonish (approved):

PR #3027, opened 2026-04-14
Capture bypass stats test v1

adaki4:
The PR is based on changes made in https://github.com/OISF/suricata-verify/pull/2969

Changes:

- add a timeout option into the test framework to shut down Suricata after a certain number of seconds when run in live mode
- add live tests to test the issue from the ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8440

PR #3045, opened 2026-04-24
Xdp tunnel 7674 v10.1

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7674

https://github.com/OISF/suricata-verify/pull/2969 continuation :
- auto generate a dummy interface if non is provided
- cleaner commit history
- tcpreplay with `-t` top speed
- AFPACKET_TEST_REPLAY feature tested at init for live tests (instead of for each test)

PR #3079, opened 2026-05-11
Capture bypass stats test v2

adaki4:
The PR is based on changes made in https://github.com/OISF/suricata-verify/pull/2969

Changes:

- rebased to master

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8440

Previous PR: https://github.com/OISF/suricata-verify/pull/3027


live/bypass-stats-shutdown/test.yaml
```diff
@@ -0,0 +1,21 @@
+requires:
+  min-version: 8
+  os: linux
+  features:
+    - AF_PACKET
```
catenacyber, line 5: Does this not need feature EBPF ?

adaki4: Yes, it does need eBPF. I can add it to the list.

To clarify how to specify features, is it okay to specify it as a string 'eBPF' (as I noticed the check for features from `suricata --build-info` is case sensitive)? I am asking because I noticed that in this [change ](https://github.com/OISF/suricata-verify/pull/3045/changes/6e897ea9af733e74f9ac8bc29a0b7e2b03d23b79#diff-d92f934294cd3707d8301c721736f757371f69195f80b2b1acee76092a86fb7dR5) an uppercase 'EBPF' is used, but in my case, the test is skipped with the uppercase version (as the SV thinks the feature is missing).

The output of `suricata --build-info` for reference:

```
Suricata Configuration:
  AF_PACKET support:                       yes
  AF_XDP support:                          yes
  DPDK support:                            no
  eBPF support:                            yes

```

catenacyber: You should cherry-pick https://github.com/OISF/suricata/commit/75e452e204392ea8b00aeb870fc8735a98d45a83 in the suricata PR

then you can do like https://github.com/OISF/suricata-verify/pull/3045/changes#diff-d92f934294cd3707d8301c721736f757371f69195f80b2b1acee76092a86fb7dR5 in SV

catenacyber: Also `AF_PACKET` is not an existing feature...

adaki4:
Continues in #3194

PR #3010, opened 2026-04-06
tests: firewall stats - v1

jufajardini:
Changes:

- move ips as a firewall tests to their own folder (out of the "firewall" folder)
  - state in their READMEs that they're not _firewall-firewall_ tests
- add checks for firewall-dedicated stats counters

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/
https://redmine.openinfosecfoundation.org/issues/7699

PR #3090, opened 2026-05-14
tests: firewall stats - v2

jufajardini:
Previous PR:  #3010 

Changes from previous PR:

Besides adjusting the firewall stats checks to the format we'll most likely use, also some side-related work, such as the exception policy drop reason -- to differentiate from firewall drops, and thus count correctly.

Note: **I've avoided adding checks for *all* firewall tests that had stats, to slightly reduce review volume. For the final version, I should probably tackle that.**

- add tests checking for non-firewall stats in non-firewall mode
- check for drop-reason related specifically to exception policy

## Ticket

https://redmine.openinfosecfoundation.org/issues/7699


jufajardini:
Thought I had all the `min-version` checks, but clearly missed places.

catenacyber:
This needs a rebase now

jufajardini:
Replaced by: https://github.com/OISF/suricata-verify/pull/3104

PR #3102, opened 2026-05-19
test/xor: add tests for variable key (var <nbytes> <offset>) and offset syntax

jlucovsky:
Continuation of #2993 

Updates
- Changed variable reference format in `xor` from using a `byte_extract` variable to having the `offset` and `nbytes` value directly in `xor` option list.
 
Add 15 test cases for the 'var <nbytes> <offset>' inline variable key syntax and the optional 'offset' parameter of the xor transform.

Positive tests:
  01 - 1-byte var key at body offset 0, XOR from offset 1; 1 alert
  02 - same as 01, content absent in decoded data; 0 alerts
  03 - 4-byte var key at body offset 0, XOR from offset 4; 1 alert
  06 - extra whitespace in offset syntax accepted; 1 alert
  10 - xor preceded by content keyword on same buffer; 1 alert
  12 - two rules with different var key locations both alert;
       verifies independent buffer per key location
  13 - var key location beyond buffer end; transform skips, 0 alerts
  14 - var key on http.uri buffer; verifies transform is not limited
       to http.request_body; 1 alert
  15 - static key and variable key rules on the same buffer both alert;
       verifies independent transforms for each key type; 2 alerts

Error tests (rule load failures, exit code 1):
  04 - var 0 0 (zero nbytes)
  05 - var 1 (missing offset)
  07 - offset with non-numeric value
  08 - offset without comma or key
  09 - offset with empty value before comma
  11 - empty quoted hex key

[Issue](https://redmine.openinfosecfoundation.org/issues/7847)

## Ticket

changes to the ticket status:


catenacyber:
Is there not a suricata PR using this ? 

Should there be a min-version: 9 ?

jlucovsky:
> Is there not a suricata PR using this ?

I added the requires suricata pr label

> Should there be a min-version: 9 ?

Yes -- this logic is new for 9; `min-version: 9` is in each `test.yaml` file.

catenacyber:
Why is CI red for main-8 and main-7 ?

jlucovsky:
> Why is CI red for main-8 and main-7 ?

Thanks for flagging this -- I've pushed an update (i omitted the `requires:` section by mistake; it's added now)

catenacyber:
> 06 - extra whitespace in offset syntax accepted; 1 alert

nit : Could it be an extra rule in another test ?

tests/detect-xor-key/detect-xor-key-04/test.yaml
```diff
@@ -0,0 +1,9 @@
+requires:
+  min-version: 9
+
+pcap: ../detect-xor-key-01/input.pcap
+
+args:
+  - -k none
+
+exit-code: 1
```
catenacyber, line 9: Should we check a log error message ?

jlucovsky:
Continued in #3174 for issue 8671

PR #3105, opened 2026-05-20
dcerpc: add test for service and procedure name

inashivb:
## Ticket

changes to the ticket status:

Redmine tickets:
- https://redmine.openinfosecfoundation.org/issues/2727
- https://redmine.openinfosecfoundation.org/issues/8523


victorjulien:
If you run the complete SV suite, are there no other services? Quick grep should be able to tell you, right?

victorjulien:
Should fail on main only

inashivb:
> If you run the complete SV suite, are there no other services? Quick grep should be able to tell you, right?

there are def more even in the same pcap. I only added a sample. I'll add all that are there in the existing pcaps and update all tests.

> Should fail on main only

indeed. 🤦🏽‍♀️

PR #3115, opened 2026-05-25
live: add namespace-based live test runner

jasonish:
Second version of #3019.

This adds a live-test runner for exercising Suricata against locally generated
traffic between isolated Linux network namespaces. The runner supports AF_PACKET
inline, NFQUEUE inline, and IDS-style AF_PACKET monitoring modes.

Included tests:

- `simple-http`: three-mode smoke test with podman Caddy server and curl client.
- `firewall-icmp`: NFQUEUE firewall reload test for blocked then allowed ICMP.
- `afp-fw-nopodman`: AF_PACKET firewall test using host caddy/curl/wget/ping.
- `afp-fw-podman`: podman-based AF_PACKET firewall variant.
- `ruletype-firewall-96-lt-response-body-no-match`: live port of the firewall
  response-body no-match test using podman client and Caddy server containers.

Notes:

- The runner uses a non-blocking lock because all live tests share fixed
  namespace and interface names.
- Tests can declare `requires.commands` for host command dependencies.
- Firewall-only tests use local empty rules files to avoid loading the default
  ruleset while preserving live firewall rule reload support.
- `ruletype-firewall-96-lt-response-body-no-match` requires
  https://github.com/OISF/suricata/pull/15402.


catenacyber:
Should CI be red ?

jasonish:
Yeah, but not after some fixups for the latest firewall work merged.

PR #3109, opened 2026-05-21
tests: firewall stats - v4

jufajardini:
Previous PR: #3104 

Changes from previous pr: 
- rebase

Creating a new PR as the last one has now failures with 8.0.x checks that I could swear weren't there when I created the PR.

Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/
https://redmine.openinfosecfoundation.org/issues/7699

PR #3116, opened 2026-05-25
tests: move non-true-firewall tests to ips dir - v1

jufajardini:
These test the engines behavior with IPS acting as firewall-like, and with the introduction of the firewall mode, were ambiguous, in the firewall directory.

This was part of the firewall stats work (https://github.com/OISF/suricata-verify/pull/3109), but I want to isolate CI checks.

...
As the checks were persisting, here, I decided to investigate. I noticed that we were fetching `libhtp` for main and main-8.0.x, and split their workflows from main-7.0.x. I suspect that the way isolate-modified-tests works may also play a role when the tests changes are only related to renaming/ location changes, but still investigating this one.


.github/workflows/builds.yml
```diff
@@ -188,7 +185,6 @@ jobs:
       - uses: actions/checkout@v3
       - run: python3 ./run.py --self-test
       - run: git clone https://github.com/OISF/suricata -b ${{ matrix.branch }}
-      - run: git clone https://github.com/OISF/libhtp suricata/libhtp
```
jasonish, line 191: I believe a conditional can also be used, something like:
```yaml
        - name: Checkout libhtp                                                      
          if: ${{ matrix.branch ==  'main-7.0.x' }}        
          run: git clone https://github.com/OISF/libhtp suricata/libhtp              
```
If the broken-out tests are largely identical other than this line, I think it simplifies the build.yml.

jufajardini: 🙇🏽  will do.

jufajardini:
Feedback incorporated with https://github.com/OISF/suricata-verify/pull/3118

PR #3118, opened 2026-05-26
tests: move non-true-firewall tests to ips dir - v2

jufajardini:
Previous PR: #3116 

Changes from previous PR:
- make libhtp repo cloning conditional, for a simpler, cleaner patch

Move ips-as-a-firewall tests to `ips` dir. These test the engines behavior with IPS acting as firewall-like, and with the introduction of the firewall mode, were ambiguous, in the firewall directory.

This was part of the firewall stats work (https://github.com/OISF/suricata-verify/pull/3109), but I wanted to isolate CI checks.

...
As the checks were persisting, here, I decided to investigate. I noticed that we were fetching libhtp for main and main-8.0.x, and split their workflows from main-7.0.x. I suspect that the way isolate-modified-tests works may also play a role when the tests changes are only related to renaming/ location changes, but still investigating this one.


catenacyber (changes requested):
Needs a rebase

jufajardini:
rebased with https://github.com/OISF/suricata-verify/pull/3144

PR #3121, opened 2026-05-27
tests: add test for missing default policy alert

jasonish:
For PR: https://github.com/OISF/suricata/pull/15464


tests/firewall/ruletype-firewall-106-default-app-policy-accept-flow-alert/test.yaml
```diff
@@ -0,0 +1,16 @@
...
+  - --simulate-ips
+  - -k none
+
+checks:
+
+# We should have one http request.
+- filter:
+    count: 1
+    match:
+      event_type: http
```
jasonish, line 16: This check fails. Remove "alert" from the default policy and it passes.

PR #3122, opened 2026-05-27
test: ban replace & bypass keywords from firewall - v1

jufajardini:
Tests for `replace` and `bypass` keywords with firewall mode/ rules.

`replace` should be banned for both firewall mode as a whole, and firewall rules.
`bypass` is banned from firewall mode (right?).

Not sure if ticket https://redmine.openinfosecfoundation.org/issues/8459 should be the correct one for bypass, or another should be created.

## Ticket

https://redmine.openinfosecfoundation.org/issues/8551


victorjulien (approved):

victorjulien:
Merged in #3126, thanks!

PR #3123, opened 2026-05-27
tests: add test for missing default policy alert

jasonish:
Like test 104, but tests "accept:tx,alert". We are not seeing the alert
when we should be.


catenacyber:
#3121 is named the same, should #3121 be closed in favor of this ?

PR #3127, opened 2026-05-29
tests: test for delayed tx clean with pass rule

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/8619

A memory leak is not directly assertable, so the accumulation is made
observable through a parser that bounds its live transactions. The pcap
is a single CLDAP (LDAP-over-UDP) flow with four complete
request/response pairs plus three trailing requests with no response.
With ldap.max-tx lowered to 4, correct cleanup frees each completed pair
and the live-transaction list never exceeds the cap. Under the bug the
completed transactions are never freed, the list grows past max-tx, and
the LDAP parser raises too_many_transactions, logged as an anomaly
record independent of detection (which is why it survives the pass
rule).

The check asserts the too_many_transactions anomaly is absent: the test
fails on the buggy code and passes once cleanup is fixed.

If this smells of AI, you are correct.


tests/bug-8619/test.rules
```diff
@@ -0,0 +1 @@
+pass udp 192.168.0.0/16 any -> any any (msg:"pass udp"; sid:1; rev:1;)
```
jasonish, line 1: Comment out this rule, or convert to an alert and this test starts to pass - we no longer get the ldap.too_many_transactions anomaly event.

catenacyber: Yes, but this test is right : we want to test this `pass` behavior

victorjulien (approved):

victorjulien:
Merged in #3139, thanks!

PR #3132, opened 2026-06-02
http: adds test for http1 and http2 and uri encoding

catenacyber:
Ticket: 8130

## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8130


tests/http-uri-plus-8130/test.rules
```diff
@@ -0,0 +1,26 @@
+# rules on uri path
+# no special encoding
+alert http any any -> any any ( http.uri; content: "/toto" ; sid:1;)
+# percent encoding in uri path
+# need to url_decode for http2
+alert http any any -> any any ( http.uri; url_decode; content: "tata" ; sid:2;)
```
catenacyber, line 6: Suricata/HTTP1 does percent decoding

But not Suricata/HTTP2 (it is strange it is needed since HTTP2 is a binary protocol but the golang HTTP2 server does this)

But we can use `url_decode` for a consistent behavior...

Is that ok ?

tests/http-uri-plus-8130/test.rules
```diff
@@ -0,0 +1,26 @@
...
+alert http any any -> any any ( http.uri; url_decode; content: "space" ; sid:12;)
+# encoded value with plus as space
+alert http any any -> any any ( http.uri; url_decode; content: "enc=after space" ; sid:13;)
+
+# percent encoding in uri path, plus is plus in path even after url_decode
+# but plus in query is decoded
+alert http any any -> any any ( http.uri; url_decode: only_decode_plus_after_query; content: "toto+tata"; content: "after space"; sid:20;)
+
+# Should not match
+alert http any any -> any any ( http.uri; url_decode: only_decode_plus_after_query; content: "toto tata"; sid:21;)
```
catenacyber, line 26: I think we want to keep `url_decode` behavior but add this `only_decode_plus_after_query` option to match better reality : The plus->space decoding should only be done in the query part, not the path part

Is that ok ? Better name than `only_decode_plus_after_query` ?

victorjulien: `query-plusspace-decode` is the matching config name in our http1 config, right? Should we use that (and crosslink in docs)?

tests/http-uri-plus-8130/test.rules
victorjulien: It's unfortunate that we have mismatching behavior between http1 and 2 parsers. Wonder if we can still update http2 w/o breaking rules?

catenacyber: We can
- keep current behavior and document
- have HTTP2 change its behavior to do percent decoding and document this change

It is a breaking change if you have rules like `alert http2 any any -> any any (http.uri; content: "%XX";)` but I guess this rule should have used http.uri.raw since the beginning

tests/http-uri-plus-8130/test.rules
catenacyber: Ok for the name `query-plusspace-decode`

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/3217

PR #3134, opened 2026-06-02
Add namespace-based live test harness and tests - v4

jasonish:
This adds a self-contained harness for *live* tests — tests where Suricata
runs in IDS or IPS mode monitoring traffic between a real client and server
in an isolated network environment, all driven from `live/run.py`.

## Test environments

Tests use Linux network namespaces to build an isolated network with three
namespaces:

- **dut** — device under test, where Suricata runs
- **server** — runs the server application (e.g. an HTTP server)
- **client** — runs the user-controlled script (e.g. `curl`)

Three Suricata run modes are supported:

- **IDS** — AF_PACKET IDS mode on an interface that sees traffic between the
  client and server namespaces (a Linux bridge, conceptually a switch span
  port).
- **AFP** — AF_PACKET IPS mode, bridging all traffic between client and server.
- **NFQ** — NFQUEUE mode (`-q 0`) intercepting all traffic between client and
  server.

Tests can declare required host commands in `test.yaml` (e.g. `podman`,
`curl`) and are skipped when a requirement is missing. A GitHub Actions
workflow (`.github/workflows/live.yaml`) runs the suite on a privileged
Ubuntu 24.04 container.

## Tests migrated from the Suricata repo

Several tests were ported from `qa/live` in the Suricata repo into this
harness:

- **afp-fw** — from `afp-fw-netns-bridge`
- **ips-drop-icmp** — from `afp-ips-netns-bridge` (AFP) and
  `nfq-ips-netns-route` (NFQ); drops ICMP echo requests while HTTP passes
- **nfq-fw** — from `nfq-fw-netns-route`; NFQUEUE firewall with rule reload
- **ruletype-firewall-96-lt-response-body-no-match** — ported from the
  Suricata-Verify ruletype test as a live AF_PACKET firewall test

Additional example tests are included: **simple-http** (simple inline IPS)
and **firewall-icmp** (ICMP firewall with a mid-test rule reload).


jasonish:
To add, ultimately, this is 3 process spaces. Suricata - running as a daemon, a server space, and a client space.  The client space "drives" things.'

Suricata could be started in unix-socket mode. The client space would feed it a series of commands/pcaps, then exit. No server process space used. I think I'll look at that next.

catenacyber:
Should there be a suricata PR to run this new mode ?

catenacyber:
> The client space "drives" things.'

Why ? Just to avoid having a fourth process space of the driver ?

> No server process space used.

But you will not be able to check that the server correctly received the data sent by the client... I thought this was one purpose here

jasonish:
> > The client space "drives" things.'
> 
> Why ? Just to avoid having a fourth process space of the driver ?

I don't really follow.  The driver would then just wait for the client to exit and do nothing else. Having the life cycle driven by the client seems to make sense.


> > No server process space used.
> 
> But you will not be able to check that the server correctly received the data sent by the client... I thought this was one purpose here

Unix-socket mode is all about pcaps. So it doesn't make sense. In this case, the client space is just doing unix socket commands, feeding in pcaps, etc.

catenacyber:
> > > The client space "drives" things.'

Nevermind, your approach makes sense

> > > No server process space used.
> > 
> > 
> > But you will not be able to check that the server correctly received the data sent by the client... I thought this was one purpose here
> 
> Unix-socket mode is all about pcaps. So it doesn't make sense. In this case, the client space is just doing unix socket commands, feeding in pcaps, etc.

This already exists in SV without a new harness : see tests/unix-socket/hostbit-add-ip6/test.yaml

jasonish:
> Should there be a suricata PR to run this new mode ?

Once accepted.  Its already run against `main` in this PR though.

catenacyber:
But Suricata CI does not run live/run.py right ?

live/tests/ruletype-firewall-96-lt-response-body-no-match/server/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM ubuntu:24.04
```
catenacyber, line 1: Do we need Docker ? Does not seem check like other commands in live/run.py ?

jasonish:
> But Suricata CI does not run live/run.py right ?

No, because its not in main yet. But this shows enough that it would pass.

live/tests/ruletype-firewall-96-lt-response-body-no-match/server/Dockerfile
jasonish: Some of these tests need Podman.

In fact, Podman could just be a global requirement as I've eliminated the tests that call caddy, tshark, etc, on the host.

It was much easier to have Podman use an existing namespace than Docker, at least in my experiments.

jasonish: I guess `Containerfile` is the new tool agnostic filename for `Dockerfile`.

catenacyber:
> > But Suricata CI does not run live/run.py right ?
> 
> No, because its not in main yet. But this shows enough that it would pass.

Os, so to answer my question "Should there be a suricata PR to run this new mode ?"

The answer is yes, there should be ;-)

live/tests/ruletype-firewall-96-lt-response-body-no-match/server/Dockerfile
catenacyber: Should podman be checked by `need_cmd` then ?

jasonish:
> > > But Suricata CI does not run live/run.py right ?
> 
> > 
> 
> > No, because its not in main yet. But this shows enough that it would pass.
> 
> 
> 
> Os, so to answer my question "Should there be a suricata PR to run this new mode ?"
> 
> 
> 
> The answer is yes, there should be ;-)

I must be missing something. It's run here, is that not good enough for proof of concept?

live/tests/ruletype-firewall-96-lt-response-body-no-match/server/Dockerfile
jasonish: Every test checks for podman in its requirements statement.  I don't want to make it a mandatory requirement, yet at least.

jasonish:
> > > > The client space "drives" things.'
> 
> Nevermind, your approach makes sense
> 
> > > > No server process space used.
> > > 
> > > 
> > > But you will not be able to check that the server correctly received the data sent by the client... I thought this was one purpose here
> > 
> > 
> > Unix-socket mode is all about pcaps. So it doesn't make sense. In this case, the client space is just doing unix socket commands, feeding in pcaps, etc.
> 
> This already exists in SV without a new harness : see tests/unix-socket/hostbit-add-ip6/test.yaml

How scriptable is that? This would let you parse the response, send a follow up, parse the response, etc, etc.  Not trying to replace it, but I think the stock SV lacks the idea of a life cycle and lifetime, and a 'driver' that is done here which may allow for testing more complex scenarios.

jasonish:
@catenacyber Have you tried this? I'm curious if it works for others.

catenacyber:
> catenacyber Have you tried this? I'm curious if it works for others.

From running `python3 ../suricata-verify/live/run.py ` getting 
```
Traceback (most recent call last):
  File "/Users/catena/catena/prod/suricata/../suricata-verify/live/run.py", line 664, in <module>
    def render_test_include(test_dir: str, output_dir: str) -> str | None:
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

catenacyber:
> is that not good enough for proof of concept?

I have the feeling that this is beyond a POC :-) this is not a draft, this is ready for review, am I wrong ?

live/run.py
```diff
@@ -0,0 +1,1900 @@
+#!/usr/bin/env python3
```
catenacyber, line 1: You picked the same directory name as in https://github.com/OISF/suricata-verify/pull/3045 :-p

jasonish: Yeah. Noted in https://github.com/OISF/suricata-verify/pull/3019:

> I think @catenacyber used live as the directory for some other tests, so happy to hear about alternative nameing for these tests

I don't fully conisder replay tests as live tests as they are still somewhat synthetic.  Maybe this live could be network as they actually utilize the network? I don't know..

jasonish:
> > catenacyber Have you tried this? I'm curious if it works for others.
> 
> From running `python3 ../suricata-verify/live/run.py ` getting
> 
> ```
> Traceback (most recent call last):
>   File "/Users/catena/catena/prod/suricata/../suricata-verify/live/run.py", line 664, in <module>
>     def render_test_include(test_dir: str, output_dir: str) -> str | None:
> TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
> ```

Whats your `python3 --version`?

catenacyber:
Python 3.9.6

jasonish:
> Python 3.9.6

Pushed a fix for EOL Python 3.9 :)

catenacyber:
Looks like every test requires podman now

jasonish:
> Looks like every test requires podman now

Yeah, Podman is pretty much everywhere these days, so it means the user only needs it installed, and not the tools that make up the clients and servers.

catenacyber:
Not getting a summary in the end of how many tests pass/fail

And I have some tests failing like
```
===> [ids/simple-http] Setting up ids environment
===> [ids/simple-http] Starting Suricata
===> [ids/simple-http] Starting Server
===> [ids/simple-http] Running Client
===> [ids/simple-http] Stopping Server
===> [ids/simple-http] Stopping Suricata
===> [ids/simple-http] FAIL
  client script exited with code 7
  client stdout: /root/suricata-verify/live/tests/simple-http/output/ids/client.stdout
  client stderr: /root/suricata-verify/live/tests/simple-http/output/ids/client.stderr
```

Or
```
===> [nfq/ips-drop-icmp] FAIL
  Suricata did not become ready
  suricata stdout: /root/suricata-verify/live/tests/ips-drop-icmp/output/nfq/stdout
  suricata stderr: /root/suricata-verify/live/tests/ips-drop-icmp/output/nfq/stderr
```

jasonish:
> Not getting a summary in the end of how many tests pass/fail

Added.

Whats in those stdout and stderr files? It passes here on GitHub, and for me locally.

catenacyber:
> Error: suricata: NFQUEUE not enabled. Make sure to pass --enable-nfqueue to configure when building. [SCParseCommandLine:suricata.c:2036]

Should this be a feature tested ?

jasonish:
> > Error: suricata: NFQUEUE not enabled. Make sure to pass --enable-nfqueue to configure when building. [SCParseCommandLine:suricata.c:2036]
> 
> Should this be a feature tested ?

Yes, added.

catenacyber:
Now I have

```
===> [ids/simple-http] FAIL
  client script exited with code 7
  client stdout: /root/suricata-verify/live/tests/simple-http/output/ids/client.stdout
  client stderr: /root/suricata-verify/live/tests/simple-http/output/ids/client.stderr

PASS: 4
FAIL: 1
SKIP: 5

Failing tests:
- ids/simple-http
```

catenacyber:
Both client files are empty

catenacyber:
Also it takes some time even if it does
```
===> [afp/ips-drop-icmp] Setting up afp environment
===> [afp/ips-drop-icmp] Running before script
===> [afp/ips-drop-icmp] Starting Suricata
===> [afp/ips-drop-icmp] Starting Server
===> [afp/ips-drop-icmp] Running Client
===> [afp/ips-drop-icmp] Stopping Server
===> [afp/ips-drop-icmp] Stopping Suricata
===> [afp/ips-drop-icmp] OK
===> [nfq/ips-drop-icmp] SKIP: requires feature NFQ
```

jasonish:
> Both client files are empty

Can you do `.../run.py -v run --mode ids simple-http`... Maybe something will come to the terminal.

jasonish:
> Also it takes some time even if it does

How much? I'd say 5-10 seconds per test per mode, maybe more on slower machines, is probably normal. Some of the tests themselves have timeouts in them before steps. We also wait up to 5 seconds for a clean Suricata shutdown.

These tests aren't going to be fast, but faster than doing them manually.

catenacyber:
Passes once then fails once
```
python3 ../suricata-verify/live/run.py -v run --mode ids simple-http
===> [ids/simple-http] Setting up ids environment
===> [ids/simple-http] Starting Suricata
===> Suricata command: ip netns exec dut /root/suricata/src/suricata -c /root/suricata/suricata.yaml --include /root/suricata-verify/live/tests/simple-http/output/ids/include.yaml -l /root/suricata-verify/live/tests/simple-http/output/ids --set unix-command.filename=/root/suricata-verify/live/tests/simple-http/output/ids/suricata.socket --set classification-file=/root/suricata/etc/classification.config --set reference-config-file=./etc/reference.config --set threshold-file=./threshold.config -i br0
===> suricata stdout: Info: conf-yaml-loader: Configuration node 'default-rule-path' redefined. [ConfYamlParse:conf-yaml-loader.c:335]
===> suricata stdout: Info: conf-yaml-loader: Configuration node 'rule-files' redefined. [ConfYamlParse:conf-yaml-loader.c:335]
===> suricata stdout: Notice: suricata: This is Suricata version 9.0.0-dev (8a4eea44d8 2026-06-11) running in SYSTEM mode [LogVersion:suricata.c:1243]
===> suricata stdout: Notice: threads: Threads created -> W: 16 FM: 1 FR: 1   Engine started. [TmThreadWaitOnThreadRunning:tm-threads.c:1960]
===> [ids/simple-http] Starting Server
===> [ids/simple-http] Running Client
===> client stderr: 
===> client stderr: WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
===> client stderr: 
===> server stderr: {"level":"info","ts":1781703351.4199739,"msg":"using provided configuration","config_file":"/etc/caddy/Caddyfile","config_adapter":""}
===> server stderr: {"level":"warn","ts":1781703351.4221637,"msg":"Caddyfile input is not formatted; run the 'caddy fmt' command to fix inconsistencies","adapter":"caddyfile","file":"/etc/caddy/Caddyfile","line":2}
===> server stderr: {"level":"info","ts":1781703351.4271688,"logger":"admin","msg":"admin endpoint started","address":"localhost:2019","enforce_origin":false,"origins":["//localhost:2019","//[::1]:2019","//127.0.0.1:2019"]}
===> server stderr: {"level":"warn","ts":1781703351.4276726,"logger":"http","msg":"server is listening only on the HTTP port, so no automatic HTTPS will be applied to this server","server_name":"srv0","http_port":80}
===> server stderr: {"level":"info","ts":1781703351.4277856,"logger":"tls.cache.maintenance","msg":"started background certificate maintenance","cache":"0xc00013c380"}
===> server stderr: {"level":"info","ts":1781703351.4285216,"logger":"http.log","msg":"server running","name":"srv0","protocols":["h1","h2","h3"]}
===> server stderr: {"level":"info","ts":1781703351.428745,"logger":"tls","msg":"cleaning storage unit","description":"FileStorage:/root/.local/share/caddy"}
===> server stderr: {"level":"info","ts":1781703351.4288864,"logger":"tls","msg":"finished cleaning storage units"}
===> server stderr: {"level":"info","ts":1781703351.4290576,"msg":"autosaved config (load with --resume flag)","file":"/root/.config/caddy/autosave.json"}
===> server stderr: {"level":"info","ts":1781703351.4290893,"msg":"serving initial configuration"}
===> client stderr: 
===> client stderr: WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
===> client stderr: 
===> client stderr: debconf: delaying package configuration, since apt-utils is not installed
===> client stdout: uid=0(root) gid=0(root) groups=0(root)
===> [ids/simple-http] Stopping Server
===> server stderr: time="2026-06-17T13:36:02Z" level=error msg="Error forwarding signal 15 to container f159e1d8be2d4e25261722a832d7bb141da5ac2839b3e3cc84c1765700d15848: error sending signal to container f159e1d8be2d4e25261722a832d7bb141da5ac2839b3e3cc84c1765700d15848: `/usr/bin/crun kill f159e1d8be2d4e25261722a832d7bb141da5ac2839b3e3cc84c1765700d15848 15` failed: signal: terminated"
===> [ids/simple-http] Stopping Suricata
===> suricata stdout: Notice: suricata: Signal Received.  Stopping engine. [SuricataMainLoop:suricata.c:3020]
===> suricata stdout: Notice: device: br0: packets: 30, drops: 0 (0.00%), invalid chksum: 0 [LiveDeviceListClean:util-device.c:353]
===> [ids/simple-http] OK

PASS: 1
FAIL: 0
SKIP: 0
root@catena-suricata:~/suricata# python3 ../suricata-verify/live/run.py -v run --mode ids simple-http
===> [ids/simple-http] Setting up ids environment
===> [ids/simple-http] Starting Suricata
===> Suricata command: ip netns exec dut /root/suricata/src/suricata -c /root/suricata/suricata.yaml --include /root/suricata-verify/live/tests/simple-http/output/ids/include.yaml -l /root/suricata-verify/live/tests/simple-http/output/ids --set unix-command.filename=/root/suricata-verify/live/tests/simple-http/output/ids/suricata.socket --set classification-file=/root/suricata/etc/classification.config --set reference-config-file=./etc/reference.config --set threshold-file=./threshold.config -i br0
===> suricata stdout: Info: conf-yaml-loader: Configuration node 'default-rule-path' redefined. [ConfYamlParse:conf-yaml-loader.c:335]
===> suricata stdout: Info: conf-yaml-loader: Configuration node 'rule-files' redefined. [ConfYamlParse:conf-yaml-loader.c:335]
===> suricata stdout: Notice: suricata: This is Suricata version 9.0.0-dev (8a4eea44d8 2026-06-11) running in SYSTEM mode [LogVersion:suricata.c:1243]
===> suricata stdout: Notice: threads: Threads created -> W: 16 FM: 1 FR: 1   Engine started. [TmThreadWaitOnThreadRunning:tm-threads.c:1960]
===> [ids/simple-http] Starting Server
===> [ids/simple-http] Running Client
===> [ids/simple-http] Stopping Server
===> [ids/simple-http] Stopping Suricata
===> suricata stdout: Notice: suricata: Signal Received.  Stopping engine. [SuricataMainLoop:suricata.c:3020]
===> suricata stdout: Notice: device: br0: packets: 15, drops: 0 (0.00%), invalid chksum: 0 [LiveDeviceListClean:util-device.c:353]
===> [ids/simple-http] FAIL
  client script exited with code 7
  client stdout: /root/suricata-verify/live/tests/simple-http/output/ids/client.stdout
  client stderr: /root/suricata-verify/live/tests/simple-http/output/ids/client.stderr

PASS: 0
FAIL: 1
SKIP: 0

Failing tests:
- ids/simple-http
```

jasonish:
> Passes once then fails once

My first thought is maybe the client is starting to soon. I've pushed some changes:
- in general, give the server more time to start
- move the container build to get the id into the before script
- add more output to verbose mode.

Please give it shot when you get a chance. Thanks for testing!

catenacyber:
Passed 5 times in a row now :-)

live/run.py
```diff
@@ -0,0 +1,1936 @@
...
+    if not m:
+        return None
+
+    major = int(m.group(1)) if m.group(1) else default_v
+    minor = int(m.group(2)) if m.group(2) else default_v
+    patch = int(m.group(3)) if m.group(3) else default_v
+    return SuricataVersion(major=major, minor=minor, patch=patch)
+
+
+class Version:
```
catenacyber, line 1065: Should you not reuse code from run.py ?

catenacyber (changes requested):
Looks good.

I think we need some code factoring for the `requires` sections

And we need to settle down on directory naming : I can rename my PR to `replay`, what do you think ?

live/run.py
jasonish: Yes, that's the plan, once we've settled and are all OK with the shape of this, I'd refactor run.py and break out some parts like requires, and the checks into a reusable lib. But didn't want to get distracted by that while making this usable.

live/tests/afp-fw/empty.rules
```diff
@@ -0,0 +1 @@
+# Empty ruleset for live tests that only use firewall rules.
```
catenacyber, line 1: Why do we need this? Why are there multiple rules files ?

live/tests/afp-fw/README.md
```diff
@@ -0,0 +1,2 @@
+A port of the "afp-fw-netns-bridge" from the Suricata repo to this
```
catenacyber, line 1: do we need index.html ?

jasonish: It is served by a caddy, so it just needs to exist, but could be generated on the fly. Did you have more thoughts on why we might not need it?

live/tests/afp-fw/empty.rules
jasonish: The `include.yaml` is included, as an override to the `suricata.yaml` in the users build directory, and we don't want to pickup any rules thay may have installed, and we can't simply specify `rule-files:` as empty, Suricata has an error here (should probably be fixed). So this `empty.rules` is used to avoid pickup the users rules.

As for the `firewall1.rules` and `firewall2.rules`, its because we start with up, swap it out and test a rule reload.  So the AFP NETNS tests in Suricata main which this is a port of, mainly for demonstration purposes.

jasonish:
New version here: https://github.com/OISF/suricata-verify/pull/3190 with "checks" and "requires" factored out into modules used by both test runners.

live/tests/afp-fw/README.md
catenacyber: Is not the one in server directory used for that ?

live/tests/afp-fw/empty.rules
catenacyber: > and we don't want to pickup any rules thay may have installed

This does not happen inside a docker/podman ?

catenacyber:
So closing this version

live/tests/afp-fw/README.md
jasonish: Indeed it is. Just needed that extra context.

live/tests/afp-fw/empty.rules
jasonish: > > and we don't want to pickup any rules thay may have installed
> 
> This does not happen inside a docker/podman ?

No, Suricata runs on the host just like in normal S-V test.  Podman is only optional for the client and server so you can setup or depend on apps not installed on the host.

PR #3140, opened 2026-06-04
flowbits: error on invalid flowbit combinations

inashivb:


catenacyber (changes requested):
Needs rebase apparently

PR #3146, opened 2026-06-04
test(firewall): add failing test for dns:< auto-accept-prior breaking packet:filter

yashda:
Bug reproduction test for a firewall mode issue where `accept:flow dns:<state`
(auto-accept prior hooks syntax) corrupts the `packet:filter` table, causing ALL
packets to be dropped by the default packet policy even though packet-layer
accept rules are present.

**What this test does:**
- Same pcap as test-23 (`dns/dns-eve/input.pcap`)
- Same logical intent (accept DNS queries matching "dropbox", drop the rest)
- Uses `dns:<request_complete` / `dns:<response_complete` instead of explicit per-state rules

**Expected (once fixed):** sid:102 fires 3×, sid:202 fires 3×, non-dropbox queries hit app-layer drop. Same as test-23.

**Actual (bug present):** 0 alerts, 8 drops at `firewall default packet policy`. Packet-layer accept rules don't match.


catenacyber:
Is there a redmine ticket about this ?

yashda:
This is being implemented and discussed here: https://redmine.openinfosecfoundation.org/issues/8472#note-13

tests/firewall/ruletype-firewall-dns-auto-accept-prior-bug/test.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+checks:
+  # sid:102 should fire for DNS queries containing "dropbox" (request_complete)
+  - filter:
+      count: 3
+      match:
+        event_type: alert
+        alert.signature_id: 102
+
+  # sid:202 should fire for DNS responses containing "dropbox" (response_complete)
```
victorjulien, line 35: this is incorrect, as the request triggers accept:flow

victorjulien:
I've taken this test in to #3154. Closing this PR.

victorjulien:
> Is there a redmine ticket about this ?

Created https://redmine.openinfosecfoundation.org/issues/8645

PR #3151, opened 2026-06-07
config: test deterministic glob include ordering

ssam18:
> [!IMPORTANT]
**Depends on OISF/suricata#15574** (the `include:` glob feature, Redmine #8427). This test exercises that feature, so the CI here will stay red until it is merged into `OISF/suricata`: suricata-verify CI builds suricata from the official branches (`master`/`main-7.0.x`/`main-8.0.x`), none of which have the feature yet, so `suricata --dump-config` fails on the glob `include:`. The pair is validated in **#15574's CI** (which builds the feature and runs this  exact branch via `SV_BRANCH`) and locally.

Adds `config-includes-glob-order`, a test for the `include:` glob-pattern support added in OISF/suricata#15574 (Redmine #8427).

`suricata.yaml` includes `*-order.yaml`, which matches three drop-in files (`01-order.yaml`, `50-order.yaml`, `99-order.yaml`) that each set the same key `glob-order-test` to a different value. Because `glob(3)` returns matches in sorted (lexicographic) order, the files are processed in ascending numeric-prefix order and the highest-numbered file is applied last — matching the common `01..99` drop-in directory convention.

Using `--dump-config`, the test asserts:
- all three matched files were included (one unique marker key per file), and
- the final value of `glob-order-test` is `file-99` (last in sorted order), not the baseline or an earlier file.

This addresses review feedback on the suricata PR (victorjulien/jasonish) requesting the include tests live in suricata-verify, with an explicit check that include ordering is deterministic.

Ticket: https://redmine.openinfosecfoundation.org/issues/8427

Verified locally against a `9.0.0-dev` build with the glob feature: `===> config-includes-glob-order: OK`.

### DIFF
--- a/tests/config-includes-glob-order/test.yaml
+++ b/tests/config-includes-glob-order/test.yaml
@@ -1,0 +1,43 @@
+requires:
+  min-version: 9
+  # Glob support in include: is not merged yet (OISF/suricata#15574), so a
+  # min-version alone can't gate this test. Probe the source tree for the
+  # glob(3) expansion code instead and skip on builds that don't have it,
+  # the same way the eve-alert-metadata tests probe for METADATA_DEFAULTS.
+  script:
+    - grep globfree src/conf-yaml-loader.c > /dev/null
+
+pcap: false
+
+# Run suricata against glob.yaml explicitly rather than shipping a suricata.yaml.
+# suricata-verify always runs `suricata -c <test>/suricata.yaml --dump-config`
+# during test setup (before the min-version check), so a glob include in
+# suricata.yaml would make that setup step fail on Suricata versions without
+# glob support. Using a separate config exercised via `command:` keeps setup on
+# the default suricata.yaml and only runs the glob config on a supported build.
+command: |
+  ${SRCDIR}/src/suricata -c ${TEST_DIR}/glob.yaml -l ${OUTPUT_DIR} --dump-config
+
+checks:
+  # Every globbed drop-in file is included, not just one: the unique marker key
+  # from each of 01/50/99-order.yaml is present in the merged config.
+  - shell:
+      args: grep '^glob-order-seen-01 = present' stdout | wc -l
+      expect: 1
+  - shell:
+      args: grep '^glob-order-seen-50 = present' stdout | wc -l
+      expect: 1
+  - shell:
+      args: grep '^glob-order-seen-99 = present' stdout | wc -l
+      expect: 1
+
+  # Ordering is deterministic: glob(3) returns matches sorted, so the files are
+  # included in ascending numeric-prefix order. Include order is observable in
+  # --dump-config output because the config tree preserves insertion order, so
+  # the distinct per-file marker keys must appear in sorted-filename order.
+  # The test does not exercise override-on-duplicate-keys (each file uses a
+  # unique key on purpose). What this pins is just the sort order, which is
+  # what the numbered conf.d-style drop-in convention relies on in practice.
+  - shell:
+      args: grep -o '^glob-order-seen-[0-9][0-9]' stdout | tr '\n' ' '
+      expect: glob-order-seen-01 glob-order-seen-50 glob-order-seen-99
--- a/tests/config-includes-glob-order/01-order.yaml
+++ b/tests/config-includes-glob-order/01-order.yaml
@@ -1,0 +1,3 @@
+%YAML 1.1
+---
+glob-order-seen-01: present
--- a/tests/config-includes-glob-order/50-order.yaml
+++ b/tests/config-includes-glob-order/50-order.yaml
@@ -1,0 +1,3 @@
+%YAML 1.1
+---
+glob-order-seen-50: present
--- a/tests/config-includes-glob-order/99-order.yaml
+++ b/tests/config-includes-glob-order/99-order.yaml
@@ -1,0 +1,3 @@
+%YAML 1.1
+---
+glob-order-seen-99: present
--- a/tests/config-includes-glob-order/README.md
+++ b/tests/config-includes-glob-order/README.md
@@ -1,0 +1,38 @@
+# config-includes-glob-order
+
+Verifies that a glob pattern in an `include:` directive expands to all matching
+files and includes them in a deterministic (sorted) order.
+
+`glob.yaml` includes `*-order.yaml`, which matches three drop-in files
+(`01-order.yaml`, `50-order.yaml`, `99-order.yaml`). Each file defines its own
+unique marker key (`glob-order-seen-NN`). Because `glob(3)` returns matches in
+sorted (lexicographic) order, the files are merged in ascending numeric-prefix
+order.
+
+The test runs suricata against `glob.yaml` with `--dump-config` (via `command:`)
+and asserts that:
+
+- all three matched files were included (one unique marker key per file), and
+- the marker keys appear in the dump output in sorted-filename order. The
+  config tree preserves insertion order, so the position of each key in the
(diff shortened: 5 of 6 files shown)


ssam18:
The 3 failing builds here are expected for now. This test exercises glob expansion in the config include directive, which is added by Suricata PR OISF/suricata#15574 and is not in suricata master yet. When suricata-verify builds suricata from master, the include is read as a literal filename, the file is not found, and suricata exits 1.

That suricata PR sets SV_BRANCH to this branch, so its own CI builds the glob feature and runs this test against it, where it passes. Running the same command locally against the feature build also exits 0 with the expected lexicographic ordering. This PR is meant to land together with or right after #15574, and once the feature is in master these builds will pass with no further changes.

tests/config-includes-glob-order/test.yaml
```diff
@@ -0,0 +1,44 @@
...
+      args: grep '^glob-order-seen-50 = present' stdout | wc -l
+      expect: 1
+  - shell:
+      args: grep '^glob-order-seen-99 = present' stdout | wc -l
+      expect: 1
+
+  # Ordering is deterministic: glob(3) returns matches sorted lexicographically,
+  # so the highest numeric prefix (99) is included last and wins.
+  - shell:
+      args: grep '^glob-order-test = file-99' stdout | wc -l
```
catenacyber: How does this check test the order ?

catenacyber: Oh, this looks bad to have the test rely on duplicated keys behavior...

catenacyber: Why should we care about the order ?

catenacyber: Why did you mark this conversation as resolved without commenting ?

ssam18: Sorry for resolving without replying first. The test has been rewritten since those comments. Each drop-in file now sets its own unique marker key, and the final shell check greps the merged dump for those keys in sequence and expects them in sorted-prefix order. So the test now checks the actual glob expansion order without depending on which value wins when keys collide, and the README explains why ordered drop-ins matter for the numbered conf.d-style convention.

catenacyber: Why does order matter ?

Readme says

> Include order matters because settings from later includes override settings from earlier ones.

But how can that happen without duplicated keys ?

ssam18: You are right, the README was overclaiming. In this test the override behavior never actually fires because each file uses a unique key. The test only pins that glob expansion is sorted. I updated the README and the test.yaml comment to drop the override claim and call out that the duplicate key case is the real world reason users care about sort order, not something this specific test exercises.

catenacyber: So, we do not care about order at all, right ?
We should not test for it, right ?

ssam18: We do care about order? it's part of the feature's documented contract, not an implementation accident. The userguide section added in OISF/suricata#15574 states: "Patterns are expanded at startup via glob(3) and each matching file is loaded in lexicographic order."

That guarantee is what makes glob includes useful for conf.d/-style layouts, where numbered prefixes (01-*.yaml … 99-*.yaml) control which drop-in wins when the same key appears in more than one file, the same convention as systemd/rsyslog/logrotate drop-in directories. Without an ordering guarantee, that layout becomes a load-order lottery.

This test pins only the ordering contract, using unique keys per your earlier feedback that relying on duplicate-key override in the test would be bad. That's deliberate: it means a future change like GLOB_NOSORT or replacing glob() with a plain directory scan can't silently invalidate the documented behavior.

If the team's position is that Suricata should not guarantee include order, I'm happy to go the other way, but then the right change is to drop the "lexicographic order" sentence from the user guide in #15574 and reduce this test to "all matching files are included." I think guaranteeing order is the more useful contract for users, so I'd prefer to keep both.

jasonish: > If the team's position is that Suricata should not guarantee include order, I'm happy to go the other way, but then the right change is to drop the "lexicographic order" sentence from the user guide in #15574 and reduce this test to "all matching files are included." I think guaranteeing order is the more useful contract for users, so I'd prefer to keep both.

Predictable order should be a must. Probably much like udev rules, or older sysv init.  `50-zzz.yaml` should be included before `90-aaa.yaml` in a globbing pattern.

catenacyber: Why so @jasonish ?

For me the yaml is like a set of key/values without order (like a golang map)

jasonish: Thats true for pure YAML @catenacyber, but once you introduce includes, its no longer pure YAML.

So if you have one include:
```
sensor-name: foobar
```
and a second include
```
sensor-name: barfoo
```
the last include is going to overwrite the previous key, so deterministic with some level of user control would be desired when building up includes. This is already easily done by ordering the includes, but if we're going to glob them, we should carry over some way to order them as well.

ssam18:
@jasonish that's exactly the scenario this test is meant to protect. I kept the marker keys unique per @catenacyber's earlier feedback, so the test pins just the sort order without relying on override behavior, but sorted expansion is what makes your sensor-name example predictable when the includes come from a glob.

On the red CI: the main jobs were failing because the glob support itself (OISF/suricata#15574) isn't merged yet, not because of the test logic. I've updated the test with a requires script that greps conf-yaml-loader.c for the glob expansion code and skips on builds without it, same approach the eve-alert-metadata tests use for METADATA_DEFAULTS. Verified locally: skips against current main (no match on 7.0.x/8.0.x either), passes against a build of the 15574 branch.

tests/config-includes-glob-order/test.yaml
catenacyber: I do not agree with this desired behavior : I think we should fail if 2 includes redefine some key

But you may know better, so I defer the decision to you

jasonish: We already support this, and a warning is output when you redefine a key.

A subsequent, new object like `vars` redefines the whole previous `vars` section.  You can do `vars.address-groups.HOME_NET: ...` later in the suricata.yaml, or in an include to override just the field.

As this is in use in production these days, by integrators, it's not something we can change. It also means predictable ordering of includes can be very beneficial.

The work was done pre-SV so there aren't many tests about it, and its only lightly documented here: https://docs.suricata.io/en/suricata-8.0.6/configuration/suricata-yaml.html#splitting-configuration-in-multiple-files

jasonish (approved):
Test looks OK.

jasonish:
Depending on https://github.com/OISF/suricata/pull/15574#pullrequestreview-4774878259, this might need a not runs on Windows requirement.

ssam18:
Superseded by #3250 (v2), which adds the Windows requirement @jasonish asked about:

> Depending on OISF/suricata#15574 (review), this might need a not runs on Windows requirement.

It gates on the build rather than the platform — `grep '^#define HAVE_GLOB_H 1' src/autoconf.h` — so the test still runs on macOS and the BSDs, where `glob(3)` is available, and skips only where it is not.

Everything else is unchanged from this PR, including the ordering checks discussed above. The paired suricata PR is now OISF/suricata#15937 (v3 of OISF/suricata#15574), which is what `SV_BRANCH` points at.

A new PR rather than a push here, per the workflow: a branch should not be updated after its pull request is opened.

Closing in favour of #3250. Thanks @catenacyber and @jasonish for the review and the discussion on ordering.

PR #3155, opened 2026-06-10
tests: bsize applied as depth to content

jlucovsky:
Test Coverage for the bsize-as-depth optimization:

- bug-4226-01: engine-analysis confirms the bsize upper bound is applied as a content depth, and the attributing note is emitted
- bug-4226-02: matching is preserved across all bsize modes
- bug-4226-03: depth application is correct with offset, multi-content distance, and negated content
- bug-4226-04: engine-analysis suggests bsize where a single content is pinned to the whole buffer via startswith/endswith
- bug-4226-05 checks that a content longer than the tightest of several bsize bounds on a buffer is rejected at load, in either keyword order.
- bug-4226-06 checks that a lone content whose length equals an exact bsize is marked startswith+endswith, and that shorter content, a non-exact bsize, and multi-content buffers are not.

Issue: 4226

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/4226

### DIFF
--- a/tests/bug-4226/README.md
+++ b/tests/bug-4226/README.md
@@ -1,0 +1,78 @@
+# Description
+
+Tests for the `bsize` optimization from
+https://redmine.openinfosecfoundation.org/issues/4226
+
+When a buffer carries a `bsize` with a usable upper bound (`bsize:N`,
+`bsize:<N`, or a range), the bound is applied as a `depth` to the content
+matches in that buffer, so the engine bounds its search instead of scanning the
+whole buffer. This mirrors the existing `dsize` and `urilen` optimizations. The
+analyzer reports both the applied optimization and a suggestion to use `bsize`
+where a rule pins a buffer to a fixed length the long way.
+
+## bug-4226-01
+
+`--engine-analysis` check that the `bsize` upper bound is applied as a content
+`depth` (`bsize:10` -> `depth:10`, `bsize:<26` -> `depth:26`,
+`bsize:13<>34` -> `depth:34`) and that `bsize:>8` applies none. Also confirms the
+analysis note attributing the `depth` to `bsize` is emitted for the bounded
+rules only.
+
+## bug-4226-02
+
+Matching behaviour is preserved across all `bsize` modes (exact, less-than,
+range, greater-than) and a partial content. A `bsize` that mismatches the buffer
+length does not alert, proving the length check still runs alongside the depth
+optimization.
+
+## bug-4226-03
+
+Correctness of the depth application against `offset`, multi-content
+`distance`, and negated content -- the cases that interact with content limit
+propagation, since the depth is applied after `DetectContentPropagateLimits`.
+
+## bug-4226-04
+
+`--engine-analysis` check that the analyzer suggests `bsize` when a single
+content is pinned to the whole buffer with `startswith`/`endswith` (and the
+`isdataat:!1,relative` form), and stays quiet when only one end is anchored,
+when `bsize` already exists, or for a plain unbounded content.
+
+## bug-4226-05
+
+A buffer may carry more than one `bsize`; the tightest (smallest) bound
+applies. A content longer than that tightest bound can never match, so the
+signature is rejected at load -- in either keyword order.
+
+## bug-4226-06
+
+An exact `bsize` equal to a lone content's length means the content fills the
+buffer, so it is marked `startswith`+`endswith`. `--engine-analysis` confirms
+the marking happens for that case and not for a shorter content, a non-exact
+`bsize`, or a buffer with more than one content.
+
+## bug-4226-07
+
+Multiple `bsize` on a buffer (a lower + upper bound forming a range) are
+accepted; the rule loads, and engine-analysis suggests collapsing them into a
+single `bsize` range.
+
+## bug-4226-08
+
+Multiple `bsize` on a buffer whose bounds share no satisfiable length are
+rejected at load: two differing exact `bsize` values, and a lower bound above
+an upper bound, in either keyword order.
+
+## bug-4226-09
+
+Same-direction bounds (two upper or two lower `bsize`) are accepted but are not
+suggested as a range -- the fold hint fires only for a genuine lower+upper pair
+(contrast bug-4226-07).
+
+# PCAP
+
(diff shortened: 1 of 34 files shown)


tests/bug-4226/bug-4226-01/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+args:
+    - --engine-analysis
+    - --set engine-analysis.rules=yes
+
+# Each content is dumped three times in rules.json (rule, mpm, prefilter), so
+# the per-bsize depth appears three times. bsize:>8 (sid:4) yields no upper
+# bound, so it contributes no depth; the total count proves it added none.
+checks:
+    - shell:
+        args: grep -o '"depth":10' rules.json | wc -l | xargs
```
catenacyber, line 14: Could you not do a json check instead of a shell check ?

tests/bug-4226/bug-4226-01/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+    - shell:
+        args: grep -o '"depth":' rules.json | wc -l | xargs
+        expect: 9
+
+    # The optimization is attributed in the analysis notes: each of the three
+    # bounded rules carries the note twice (content dumped per match list),
+    # and bsize:>8 (sid:4) contributes none.
+    - shell:
+        args: grep -o "applied as a 'depth'" rules.json | wc -l | xargs
```
catenacyber, line 33: `"applied as a 'depth'"` is not self-explanatory

tests/bug-4226/bug-4226-01/test.rules
```diff
@@ -0,0 +1,6 @@
+# bsize is applied as a depth to the content matches in the same buffer when
+# it yields a usable upper bound. bsize:>N has no upper bound, so no depth.
+alert dns any any -> any any (msg:"bsize exact -> depth"; dns.query; content:"google.com"; bsize:10; sid:1;)
```
catenacyber, line 3: I would split the test per rule/case to make it clearer

- one rule with the bsize meaning perfect exact match
- `bsize:<26;` and `bsize:13<>34;` can go into the same test as upper bound
- `bsize:>8;`

tests/bug-4226/bug-4226-01/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+    - shell:
+        args: grep -o '"depth":34' rules.json | wc -l | xargs
+        expect: 3
+
+    - shell:
+        args: grep -o '"depth":' rules.json | wc -l | xargs
+        expect: 9
+
+    # The optimization is attributed in the analysis notes: each of the three
+    # bounded rules carries the note twice (content dumped per match list),
```
catenacyber, line 30: Could we add a direction to the rule to have once instead of twice ?

tests/bug-4226/bug-4226-04/test.yaml
```diff
@@ -0,0 +1,15 @@
...
+args:
+    - --engine-analysis
+    - --set engine-analysis.rules=yes
+
+# The two pinned rules (sids 1 and 2) each carry the suggestion note twice
+# (content dumped per match list); the four other shapes carry none, so the
+# total count of 4 proves the heuristic stays quiet on the negatives.
+checks:
+    - shell:
+        args: grep -o "pins a single content" rules.json | wc -l | xargs
```
catenacyber, line 14: How could we test that the suggestions are for the expected correct rules ?

tests/bug-4226/bug-4226-05/test.rules
```diff
@@ -0,0 +1,6 @@
+# A buffer may carry more than one bsize; the tightest (smallest) upper bound
+# applies. When the required content is longer than that tightest bound the
+# signature can never match, and it is rejected at load -- regardless of the
+# order the bsize keywords appear in.
+alert dns any any -> any any (msg:"tight bound first"; dns.query; content:"abcdefghij"; bsize:<8; bsize:<20; sid:1;)
```
catenacyber, line 5: Should we accept such a rule ?

tests/bug-4226/README.md
```diff
@@ -0,0 +1,60 @@
...
+## bug-4226-04
+
+`--engine-analysis` check that the analyzer suggests `bsize` when a single
+content is pinned to the whole buffer with `startswith`/`endswith` (and the
+`isdataat:!1,relative` form), and stays quiet when only one end is anchored,
+when `bsize` already exists, or for a plain unbounded content.
+
+## bug-4226-05
+
+A buffer may carry more than one `bsize`; the tightest (smallest) bound
```
catenacyber, line 43: > A buffer may carry more than one `bsize`

Should we let that happen ?

tests/bug-4226/bug-4226-05/test.rules
jlucovsky: @catenacyber 
back to back bsize. keywords or multiple bsize keywords?

I agree that the back-to-back keywords would be better disallowed. However, a recent ET/Pro ruleset has 837 instances of multi-bsize-keyword rules.

tests/bug-4226/bug-4226-01/test.yaml
jlucovsky: Yes -- will rework.

tests/bug-4226/bug-4226-01/test.rules
jlucovsky: Good point; will do.

tests/bug-4226/bug-4226-01/test.yaml
jlucovsky: will reword.

tests/bug-4226/bug-4226-04/test.yaml
jlucovsky: good point - will change.

tests/bug-4226/bug-4226-05/test.rules
catenacyber: Multiple bsize on the same buffer

At least this example `bsize:<8; bsize:<20;` should be rejected as it should be just `bsize: <8;`

PR #3156, opened 2026-06-10
test(firewall): add tcp.session: keyword verification fixtures

yashda:
Add 7 suricata-verify fixtures for the tcp.session: keyword:

- tcp-session-equivalence: asserts tcp.session:setup,established matches identical packet set as flow:not_established + flow:established
- tcp-session-stream-out-of-window-honored: out-of-window segment dropped regardless of tcp.session: accept rule
- tcp-session-stream-midstream-policy-honored: midstream pickup rejected regardless of tcp.session: accept rule
- tcp-session-decode-malformed-ipv4: decode-event emitted regardless of tcp.session: accept rule
- tcp-session-syn-flood-drop-precedence: explicit drop takes precedence over tcp.session: accept after threshold trips
- tcp-session-ipv6-stream-out-of-window-honored: IPv6 stream validation
- tcp-session-ipv6-decode-malformed-header: IPv6 decode event validation

Corresponding Suricata PR: https://github.com/OISF/suricata/pull/15705

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7704


tests/firewall/fw-intent-tcp-session-equivalence/test.yaml
```diff
@@ -0,0 +1,58 @@
...
+      stats.detect.engines[0].rules_failed: 0
+
+- filter:
+    count: 10
+    match:
+      event_type: alert
+      alert.signature_id: 9001
+
+- shell:
+    args: 'echo $(( $(grep -c ''"signature_id":9001'' eve.json) - ( $(grep -c ''"signature_id":9002'' eve.json) + $(grep -c ''"signature_id":9003'' eve.json) ) )) | xargs'
```
victorjulien: please use the yaml filter statements instead

victorjulien: (everywhere in the PR)

tests/firewall/fw-intent-tcp-session-equivalence/test.yaml
```diff
@@ -0,0 +1,58 @@
+requires:
+  min-version: 9
+
+category: novel-feature
```
victorjulien: this has no meaning for us and should be removed (everywhere in the PR)

victorjulien (changes requested):
what is the idea of the `fw-intent` prefix in test names?

victorjulien:
One thing that stands out to me here is that we essentially totally rely on features outside the new keyword to block bad packets. I think in the iptables+conntrack world we have the invalid state, which would mean these bad packets. I wonder if it makes sense to support that for tcp.session as well and (optionally?) allow these packets to reach the detection engine, where tcp.session:invalid could then drop them.

yashda:
> what is the idea of the `fw-intent` prefix in test names?

That was from the original idea of being able to write intent-based rules. Updated the test names now.

yashda:
> One thing that stands out to me here is that we essentially totally rely on features outside the new keyword to block bad packets. I think in the iptables+conntrack world we have the invalid state, which would mean these bad packets. I wonder if it makes sense to support that for tcp.session as well and (optionally?) allow these packets to reach the detection engine, where tcp.session:invalid could then drop them.

victorjulien (approved):

victorjulien (approved):

victorjulien:
Merged in #3296, thanks!

PR #3160, opened 2026-06-11
tests: test for filestore fd exhaustion

jasonish:
Test that we don't leak file descriptors when configured to not keep any
open.

### DIFF
--- a/tests/filestore-fd-exhaustion/README.md
+++ b/tests/filestore-fd-exhaustion/README.md
@@ -1,0 +1,38 @@
+# filestore-fd-exhaustion
+
+Regression test for the file-store open-file-descriptor accounting at the
+default `max-open-files: 0`.
+
+The default is documented as "files get closed after each write to the file",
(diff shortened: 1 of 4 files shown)


catenacyber:
Is this meant to pass @jasonish ?

jasonish:
> Is this meant to pass @jasonish ?

Yes. It tests a regression in one version of https://github.com/OISF/suricata/pull/15634, and should prevent that regression from happening again.

catenacyber:
Test does not fail locally at 8a4eea44d8a4b754a368555850769a7aec1fc975 for me

jasonish:
> Test does not fail locally at 8a4eea44d8a4b754a368555850769a7aec1fc975 for me

It fails on this PR: https://github.com/OISF/suricata/pull/15607

But the idea is still sound. That PR started to leak file descriptors, and we didn't have a test to catch that. This test will ensure future refactors won't.

catenacyber (approved):
Looks good thanks Jason

PR #3161, opened 2026-06-11
framework: improve error msg for unknown exception - v2

jufajardini:
Previous PR: #3016 

If an internal error happens while reading the test.yaml file, indicate which test caused it as well as which sub-check in the file, to be more user-friendly.

Changes from previous PR:
- improve error message
- improve commit message.

Samples:

Yaml examples:
```yaml
checks:
  - filter:
      count: 0
      match:
      event_type: drop
      drop.reason: "firewall default packet policy"
```

```yaml
  - filter:
      count: 3 2
      match:
        event_type: alert
        alert.signature_id: 102
```

Output: 

Before:
```
===> ruletype-firewall-110-dns-lte: FAILED: Internal runtime error: Unexpected key in filter check: event_type
```

```
===> ruletype-firewall-110-dns-lte: FAILED: Internal runtime error: %d format: a real number is required, not str
```

With patch:
```
===> ruletype-firewall-110-dns-lte: FAILED: Internal runtime error: "Unexpected key in filter check: event_type"  with test.yaml, check #3. Full test path: ../sv-fix-nonetype/tests/firewall/ruletype-firewall-110-dns-lte.
```

```
===> ruletype-firewall-110-dns-lte: FAILED: Internal runtime error: "%d format: a real number is required, not str"  with test.yaml, check #1. Full test path: ../sv-fix-nonetype/tests/firewall/ruletype-firewall-110-dns-lte.
```

### DIFF
--- a/run.py
+++ b/run.py
@@ -189,7 +189,8 @@ def handle_exceptions(func):
                 print("===> {}: Sub test #{}: SKIPPED : {}".format(kwargs["test_name"], kwargs["test_num"], ue))
             kwargs["count"]["skipped"] += 1
         except Exception as err:
-            raise TestError("Internal runtime error: {}".format(err))
+            test_path = os.path.relpath(args[0].directory, args[0].cwd)
+            raise TestError("Internal runtime error: \"{}\"  with test.yaml, check #{}. Full test path: {}.".format(err, kwargs["test_num"], test_path))
         else:
             if result:
               kwargs["count"]["success"] += 1


run.py
```diff
@@ -189,7 +189,8 @@ def applicator(*args, **kwargs):
                 print("===> {}: Sub test #{}: SKIPPED : {}".format(kwargs["test_name"], kwargs["test_num"], ue))
             kwargs["count"]["skipped"] += 1
         except Exception as err:
-            raise TestError("Internal runtime error: {}".format(err))
+            test_path = os.path.relpath(args[0].directory, args[0].cwd)
+            raise TestError("Internal runtime error: \"{}\"  with test.yaml, check #{}. Full test path: {}.".format(err, kwargs["test_num"], test_path))
```
inashivb, line 193: I'd be cautious as the exception caught here is the generic "all exceptions".

catenacyber: Cautious about what ?
The patch does not change the formatting of the exception (just adds some nice stuff around it)

inashivb: It says the error happened with test.yaml, samples are also about malformed yaml. There could be other types of errors caught in the catch-all exception?

catenacyber: I think it is still an improvement even if it is not perfect

inashivb: sure. I just shared a concern. Please feel free to accept. :)

catenacyber (approved):
Looks ok to me, @jasonish thoughts ?

run.py
jufajardini: I've only ever encountered these with the test.yaml files 😅 
Would it be better if I raised specific exception errors, and handled them here? Or at least remove the `test.yaml` hard coded mention, and just leave the full path?

inashivb: yes, something like that would have been my preference but since it's ok for Philippe, it's OK as-is, I guess. :)

jasonish (approved):

victorjulien:
Merged in #3371, thanks!

PR #3162, opened 2026-06-12
conf: adds test that we do not crash on yaml null

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8651


jasonish:
Test looks fine, but any idea why Windows still passes?

catenacyber:
> Test looks fine, but any idea why Windows still passes?

No, maybe something like strlen(NULL) is safer on windows ?

jasonish:
> > Test looks fine, but any idea why Windows still passes?
> 
> No, maybe something like strlen(NULL) is safer on windows ?

Yes, apparently the Windows C runtime does have some extra parameter validations: https://learn.microsoft.com/ga-ie/cpp/c-runtime-library/parameter-validation?view=msvc-170

victorjulien (approved):

victorjulien:
Merged in #3163, thanks!

PR #3167, opened 2026-06-16
ftp: tests to show some firewall issues - v1

jasonish:
Adds some baseline tests then some firewall tests.

Shows had ftpdata_command: retr is not possible to match against in firewall
model as ftpdata_command makes some assumptions on direction that don't always
hold up.


tests/firewall/ruletype-firewall-112-ftp-download-passive/test.yaml
```diff
@@ -0,0 +1,35 @@
...
+      count: 1
+      match:
+        event_type: ftp_data
+        ftp_data.command: RETR
+
+  # ftpdata_command: retr matches in ftp-data:response_started
+  - filter:
+      count: 1
+      match:
+        alert.signature_id: 5000
```
victorjulien, line 35: check for 5010 is missing, think it would be good to add for completeness

victorjulien:
The firewall tests both are positive tests: matching the rules, allowing the traffic. Can you also add tests for negative rulesets (explicit drop, or drop by rule omission). In addition, I think it would be good to test the default policy side as well.

victorjulien:
Not expecting much difference, but both fw tests use `retr`, would be good to complete it by adding 2 `stor` based tests too?

tests/firewall/ruletype-firewall-112-ftp-download-passive/test.yaml
jasonish: Added, makese sense.

jasonish:
> The firewall tests both are positive tests: matching the rules, allowing the traffic. Can you also add tests for negative rulesets (explicit drop, or drop by rule omission). In addition, I think it would be good to test the default policy side as well.

Those are in another branch; this was more specific to show the issues in https://github.com/OISF/suricata/pull/15656. But it would be nice to clear this out first.

jasonish:
> Not expecting much difference, but both fw tests use `retr`, would be good to complete it by adding 2 `stor` based tests too?

Make sense.

jasonish:
> Not expecting much difference, but both fw tests use `retr`, would be good to complete it by adding 2 `stor` based tests too?

Added, just mirror the download tests for now.

jasonish:
> > The firewall tests both are positive tests: matching the rules, allowing the traffic. Can you also add tests for negative rulesets (explicit drop, or drop by rule omission). In addition, I think it would be good to test the default policy side as well.
> 
> Those are in another branch; this was more specific to show the issues in [OISF/suricata#15656](https://github.com/OISF/suricata/pull/15656). But it would be nice to clear this out first.

Mainly because other fixes are required to do something like allow certain commands and disallow others.

victorjulien (approved):

victorjulien:
Merged in #3170, thanks!

PR #3186, opened 2026-06-23
Xdp tunnel 7674 v10.2

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7674

https://github.com/OISF/suricata-verify/pull/3045 continuation :
- rename to replay (instead of live)
- use of veth instead of dummy interface


run.py
```diff
@@ -853,7 +855,19 @@ def run(self, outdir):
 
                 p = subprocess.Popen(
                     args, shell=shell, cwd=self.directory, env=env,
-                    stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE, start_new_session=True)
+
+                if self.suricata_config.interface != None:
+                    argsl = ["tcpreplay", "-t", "-i", self.suricata_config.interface]
```
adaki4: Does the tcpreplay have to replay packets at max speed (`-t` flag)? While trying this version of the framework to test XDP bypass, the filter did not catch anything because Suricata did not have enough time to register the bypass.

jasonish:
```
Live tests require feature AFPACKET_TEST_REPLAY
```
Last time I tried, we didn't need this. Is it really necessary, would a timeout not do, even as a fallback without this feature being compiled in?

catenacyber:
> ```
> Live tests require feature AFPACKET_TEST_REPLAY
> ```
> 
> Last time I tried, we didn't need this.

That was marked as a TODO then

>  Is it really necessary, would a timeout not do, even as a fallback without this feature being compiled in?

There is already a timeout fallback (even if it is 1200 seconds I think)

This seems necessary to me so that Suricata knows when to stop when it processed X packets (the number in the replayed pcap)
This seems much more precise and reproducible and fast than a timeout...
What do you think ?

run.py
catenacyber: It can be a per test option indeed

catenacyber: added

run.py
```diff
@@ -853,7 +855,22 @@ def run(self, outdir):
 
                 p = subprocess.Popen(
                     args, shell=shell, cwd=self.directory, env=env,
-                    stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                    stdout=subprocess.PIPE, stderr=subprocess.PIPE, start_new_session=True)
+
+                if self.suricata_config.interface != None:
+                    argsl = ["tcpreplay"]
+                    for arg in self.config["tcpreplay_args"]:
```
adaki4, line 862: When running the test without any `tcpreplay_args` I get an error:
```
===> xdp-bypass-vlan: FAILED: Unexpected exception: 'tcpreplay_args'
Traceback (most recent call last):
  File "../suricata-verify/run.py", line 1204, in run_test
    results = test_runner.run(outdir)
  File "../suricata-verify/run.py", line 874, in run
    for arg in self.config["tcpreplay_args"]:
KeyError: 'tcpreplay_args'
```

Perhaps you can add a NULL check so we can run the test without any tcpreplay arguments?

catenacyber: oops right

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/3206

PR #3187, opened 2026-06-23
tests: add DHCP single-direction transaction tests

jlucovsky:
DHCP is a stateless parser where each datagram is its own standalone, single-direction transaction. Verify that a request/reply exchange alerts once per datagram rather than once per direction, and that directional rules only match the transaction for their own direction.

bug-8621-01 checks a generic dhcp rule fires exactly once on the request and once on the reply. bug-8621-02 checks a flow:to_server rule only matches the request and a flow:to_client rule only matches the reply.

Issues: 8621, 8658

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8621
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8658


victorjulien (approved):

victorjulien:
Merged in #3197, thanks!

PR #3188, opened 2026-06-23
tests: ipv6 hop-by-hop option length past option area

uwezkhan:
Companion test for OISF/suricata#15695.

The hop-by-hop option loop in `DecodeIPV6ExtHdrs` gated each option on `(offset + 1 + ip6_optlen) > optslen`, which let an option whose data runs one byte past the option area through; the Router Alert memcpy then read one byte past the area. The fix bounds it on `offset + 2 + ip6_optlen`.

The pcap is an IPv6 packet (payload len 8, next header hop-by-hop) carrying an 8 byte hop-by-hop header with three Pad1 options and a Router Alert that declares 2 data bytes while only 1 fits in the 6 byte option area, so the option is rejected as `decoder.ipv6.exthdr_invalid_optlen`.

Branch named to match the suricata PR so the verify run is picked up.


catenacyber:
Closing as suricata PR got closed

PR #3190, opened 2026-06-24
Live tests using Linux namespaces - v5

jasonish:
A harness and set of tests that uses Linux namespaces to create a server (http, etc), a client, and Suricata running as the device under test that must pass the packets between client and server in IPS/firewall modes. Optionally a Linux bridge can be used and Suricata will run in IDS mode effectively using the bridge as a network tap for passive live tests.

Previous PR: https://github.com/OISF/suricata-verify/pull/3134

This PR factors out the "checks" and "requires" into a common module used by both runtimes.


live/tests/afp-ips-xdp-bypass/README.md
```diff
@@ -0,0 +1,11 @@
+---
+tags:
+- bypass
+- xdp
+---
+
+# AF-PACKET IPS with XDP bypass test
```
jasonish, line 7: @catenacyber Curious what you think of this test.

live/tests/afp-ips-xdp-bypass/test.yaml
```diff
@@ -0,0 +1,54 @@
...
+        event_type: alert
+        alert.signature_id: 1
+        metadata.flowbits[0]: "bypassed"
+
+  # SID 2 should match once on the non-bypassed flow.
+  - filter:
+      count: 1
+      match:
+        event_type: alert
+        alert.signature_id: 2
```
catenacyber, line 36: Hard to tell that this signature is on the not-bypassed flow

live/tests/afp-ips-xdp-bypass/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Test AF-PACKET IPS with XDP bypass/offload. Test checks that packets are
+bypassed by Suricata, but uses a client and server to verify that the traffic
+still flows.
```
catenacyber, line 11: Where do we test this part ?

live/tests/afp-ips-xdp-bypass/test.yaml
jasonish: The comment doesn't help? After something else?

live/tests/afp-ips-xdp-bypass/README.md
jasonish: The client will fail if it doesn't receive back from the server what it sent (simple echo). That will cause the test to fail.

jasonish:
CI is failing on this test, `ids/afp-xdp-bypass-stats-shutdown`. This is currently expected as its testing for issue https://redmine.openinfosecfoundation.org/issues/8440.  Running with PR https://github.com/OISF/suricata/pull/15732 will show this test as passing.

live/tests/afp-ips-xdp-bypass/README.md
catenacyber: Deserves a comment as I expected to see something more explicit in test.yaml ;-)

live/tests/afp-ips-xdp-bypass/test.yaml
catenacyber: How can you be sure from test.yaml on which flow the signature matched ?
You have 2 flows that trigger this but one is bypassed
Maybe you can add some flowbits ;-)

live/tests/afp-ips-xdp-bypass/README.md
catenacyber: Commented on different lines  ;-)

tests/alert-testmyids-frames/test.yaml
```diff
@@ -1,3 +1,7 @@
+requires:
+  command:
+    - xargs
```
catenacyber, line 3: Eheh, where do we not have xargs ?

Could be its own PR ;-)

run.py
```diff
@@ -25,6 +25,9 @@
 from __future__ import print_function
 
 import sys
+
+sys.dont_write_bytecode = True
```
catenacyber, line 29: What is this ?

live/run.py
```diff
@@ -0,0 +1,1617 @@
...
+        stderr_log=stderr_log,
+        stdout_thread=stdout_thread,
+        stderr_thread=stderr_thread,
+    )
+
+
+SuricataVersion = namedtuple("SuricataVersion", ["major", "minor", "patch"])
+
+
+def parse_suricata_version(
```
catenacyber, line 1051: Should this code be still factorized ?

live/tests/demo-afp-xdp-bypass-stats-shutdown/client.py
```diff
@@ -0,0 +1,33 @@
...
+"""
+
+import socket
+import sys
+import time
+
+HOST = "10.200.0.1"
+PORT = 7000
+
+PAYLOAD = b"\x17\x03\x03" + b"\x00" * 100
```
catenacyber, line 14: This feels weird

live/tests/demo-afp-xdp-bypass-stats-shutdown/test.rules
```diff
@@ -0,0 +1 @@
+alert ip4 any any -> any any (msg:"Dropping all IPv4 traffic"; bypass; sid:1; rev:1;)
```
catenacyber, line 1: we do not need tls in fact ?

live/tests/demo-afp-xdp-bypass-stats-shutdown/test.yaml
```diff
@@ -0,0 +1,39 @@
+mode: [ids]
+
+requires:
+  command:
+    - python3
```
catenacyber, line 5: Do we need to explicitly require python3 when we run with python3 ?

live/README.md
```diff
@@ -0,0 +1,43 @@
...
+Tests can declare required host commands in `test.yaml`. If any command
+is missing, the test is skipped. Tests run in NFQ mode also implicitly
+require Suricata's `NFQ` build feature.
+
+```
+requires:
+  command:
+    - podman
+    - curl
+```
```
catenacyber, line 43: How do we specify tests for client/server ?
Just have the client/server exit with code != 0 if we want the test to fail ?

live/tests/demo-afp-xdp-bypass-stats-shutdown/test.rules
jasonish: Again, from https://github.com/OISF/suricata-verify/pull/3194.

live/tests/demo-afp-xdp-bypass-stats-shutdown/client.py
jasonish: From https://github.com/OISF/suricata-verify/pull/3194, just for demo purposes.

live/README.md
jasonish: The test will fail if the server exits non-zero before the client is done. Once the client is done, the server return code is ignored. The client script (which is the test driver) could check server logs and still exit. We could tune this with an expected exit code I suppose.

The client script needs to exit successfully; otherwise, the test fails.

live/tests/demo-afp-xdp-bypass-stats-shutdown/test.yaml
jasonish: Just for completeness I guess, if anyone decides to rewrite the runner in Rust :)

run.py
jasonish: Prevents Python from dropping __pycache__ files. As this runs as root, its annoying to have these root owned artifacts.

live/tests/afp-ips-xdp-bypass/test.yaml
jasonish: > How can you be sure from test.yaml on which flow the signature matched ?

I guess for this test I don't care.  I know we have 2 flows, and only one triggered the tripwire. Thats all I'm testing here.

live/run.py
jasonish: Do you mean should it be factorized?  Yeah, there's probably a lot more than could be factorized. Still focusing on the shape of the tests, though.

run.py
jasonish: Well, this is the main runner which I don't run as root.  These files still bother me though.

live/run.py
catenacyber: Yes :-)

live/README.md
catenacyber: Ok, but that needs to be in the live/README.md doc ;-)

live/tests/demo-afp-xdp-bypass-stats-shutdown/test.yaml
catenacyber: 🤷

live/README.md
jasonish: Updated the README with some test life cycle docs.

jasonish:
Replace by https://github.com/OISF/suricata-verify/pull/3212.

PR #3194, opened 2026-06-25
Capture bypass stats test v3

adaki4:
Changes:
- Change pcap for one generated by scapy program
- Add scapy program
- Split commit with changes from different commits and cherry-picked them: https://github.com/OISF/suricata-verify/commit/1b52a649803589fc19ed44e26ed717e8ea1588ce https://github.com/OISF/suricata-verify/commit/e544957a90c8fe7dd5b0f287fb4183c0927f0fc6 https://github.com/OISF/suricata-verify/commit/51f9f096810a87e011af332fd8acf7116285c467 from https://github.com/catenacyber/suricata-verify/tree/xdp-tunnel-7674-v10.2
- Removed `-t` (top-speed) flag from tcpreplay
- Add a timeout for bypassed flows larger than the test duration
- Rebased to master

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8440

Previous PR: https://github.com/OISF/suricata-verify/pull/3079


replay/bypass-stats-shutdown/test.yaml
```diff
@@ -0,0 +1,20 @@
+requires:
+  min-version: 8
+  os: linux
+  features:
+    - EBPF
+files:
+  - ebpf/bypass_filter.bpf
```
jasonish, line 7: Should this be `xdp_filter.bpf`?

jasonish:
@adaki4 @catenacyber I wonder if this test I added to the "live" framework is equivalent to this test, and a test for Redmine issue 8440.

https://github.com/OISF/suricata-verify/pull/3190/changes/12004e84f3ed1884847a5982bbe2ce616a8e5b4c (on PR https://github.com/OISF/suricata-verify/pull/3190/)

Instead of a reply, its a live server and client where the server sends the same byte sequence that is in this tests pcap/scapy script. The server just sinks the data.

We then do some checks that packets were actually bypassed, and bypassed outside of Suricata:
```yaml

checks:

  - filter:
      count: 1
      match:
        event_type: flow
        dest_port: 7000
        flow.bypassed.pkts_toserver.__gt: 0
        flow.bypassed.pkts_toclient.__gt: 0
        flow.bypassed.bytes_toserver.__gt: 0
        flow.bypassed.bytes_toclient.__gt: 0
        flow.state: bypassed
        flow.bypass: capture

  - stats:
      flow.end.state.capture_bypassed: 1
      flow_bypassed.pkts.__gt: 0
      flow_bypassed.bytes.__gt: 0
```

The purpose of the "live" framework is to test real live scenarios, not pcaps or replays, but it seems to lend itself well to testing this scenario as well.

catenacyber:
I have the feeling both framework can be ok for this issue.

And I think I prefer pcap replay to have exact stats ;-)

replay/bypass-stats-shutdown/test.yaml
adaki4: Yes it should be. I will change it in the next iteration.

adaki4:
@jasonish The live server framework also seems plausible for this purpose. Maybe it would also work well for https://github.com/OISF/suricata/pull/15773 (ticket [8442](https://redmine.openinfosecfoundation.org/issues/8442)), where we need to test the concurrency of 2 threads?

adaki4:
Continues in https://github.com/OISF/suricata-verify/pull/3209

PR #3200, opened 2026-06-30
test: adds test for list-keyword json option

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8675


catenacyber:
Next version in #3219

PR #3201, opened 2026-06-30
tests: update ftp tests for 8.0.6

jasonish:
With the backport of the FTP firewall work we can set the version on
these tests to 8.0.6.


victorjulien:
Why are some of the tests removed?

jasonish:
> Why are some of the tests removed?

The .2 versions of the tests were just temporary as main-8.0.x and main diverged in compatibility between the initial merge, and this work. With this backported, we only need the .2 variants, so I renamed them back to the original names. Should have noted that in the commit message, as they would have become identical.

victorjulien (approved):

victorjulien:
Merged in #3203, thanks!

PR #3202, opened 2026-06-30
firewall/smtp/v1

jasonish:
- **tests: add trivial smtp firewall test**
  

- **test: simple smtp firewall test with send/rcpt validation**
  Only allows mail from/rcpt to specific addresses.
  

- **tests: smtp firewall test with blocked sender**
  

- **tests: smtp firewall test with blocked rcpt to**


jufajardini (changes requested):
Looks good, only commits for test 122 could be in a single commit, it seems.

victorjulien:
Can you add some tests where unexpected things happen? RSET, QUIT commands before and after RCPT TO, other more interactive commands, etc.

jasonish:
> Can you add some tests where unexpected things happen? RSET, QUIT commands before and after RCPT TO, other more interactive commands, etc.

Add new commit with an allow focused test that has an interaction like:
```
<--- 220 mail.example.com ESMTP Postfix

---> helo client.example.com
<--- 250 mail.example.com

---> NOOP
<--- 250 2.0.0 Ok

---> VRFY bob
<--- 252 2.0.0 bob

---> EXPN staff
<--- 500 5.5.2 Error: command not recognized

---> HELP
<--- 500 5.5.2 Error: command not recognized

---> mail from:<alice@example.com>
<--- 250 2.1.0 Ok

---> NOOP
<--- 250 2.0.0 Ok

---> rcpt to:<bob@example.com>
<--- 250 2.1.5 Ok

---> data
<--- 354 End data with <CR><LF>.<CR><LF>

---> (message body)
---> .
<--- 250 2.0.0 Ok: queued as <id>

---> QUIT
<--- 221 2.0.0 Bye
```

I've got more pcaps with more scenarios made, but still need to get tests built around them.

victorjulien:
Will need a clean PR that passes on 7/8.

PR #3204, opened 2026-07-01
Xdp bypass vlan v1

adaki4:
## PR tests correct bypass of VLAN-tagged packets with xdp bypass

Changes:
- Test if a flow with VLAN-tagged packets is correctly bypassed
- Uses **live** test feature from https://github.com/OISF/suricata-verify/pull/3186
- Cherry picked commits 7c32eefb83336bc6f49e897bc37a6c56463bcbc2 908e5c0265e9f7ed72d10aff3443a2b28dd3210d and 306630e352640d086ac5e3d7006eeac25829d18b from https://github.com/OISF/suricata-verify/pull/3186
- Add tcpreplay argument`-q` to prevent the test from crashing


run.py
```diff
@@ -855,6 +864,12 @@ def run(self, outdir):
                     args, shell=shell, cwd=self.directory, env=env,
                     stdout=subprocess.PIPE, stderr=subprocess.PIPE)
 
+                if "live-shutdown-after" in self.config:
```
catenacyber, line 867: Why do we need this ?

I would rather change `r.join(timeout=PROC_TIMEOUT)` to use a configurable value if needed, what do you think ?

catenacyber (changes requested):
Needs last version of my commits after your review : see https://github.com/OISF/suricata-verify/pull/3206

run.py
adaki4: I tried to make `PROC_TIMEOUT` configurable, but I noticed that the value does not affect when Suricata shuts down and the test hangs. Do you have any other suggestions on how to shut down Suricata after a certain number of seconds?

adaki4: Also, if we agree on some way to implement this feature, it would possibly be beneficial to merge it independently, as https://github.com/OISF/suricata-verify/pull/3208 and https://github.com/OISF/suricata-verify/pull/3209 also depend on it. What do you think?

catenacyber: See my latest version in https://github.com/OISF/suricata-verify/pull/3243 ;-)

PR #3207, opened 2026-07-01
Live tests using Linux namespaces - v6

jasonish:
- Cleanups from last commits (remove the xargs change)
- Add a tcpreplay test


live/run.py
```diff
@@ -0,0 +1,1619 @@
...
+
+
+def get_include_args(
+    mode: str, script_dir: str, test_include: str | None = None
+) -> list[str]:
+    """Return Suricata --include args for the selected mode and test."""
+    args = []
+    if mode == "afp":
+        args += ["--include", os.path.join(script_dir, "etc", "af-packet.yaml")]
+    if test_include:
```
lukashino, line 701: I've noticed that even in AFP tests you include configs, so for a cleaner include path, the default af-packeet.yaml should only be included if test_include  is not specified?

live/tests/afp-fw/test.yaml
```diff
@@ -0,0 +1,108 @@
...
+  if run ping -c 10 -i 0.05 -W 1 10.200.0.1; then
+    echo "error: ping should have failed"
+    errors="yes"
+  fi
+
+  cp ${TESTDIR}/firewall2.rules ${OUTDIR}/firewall.rules
+  ${SURICATASC} -c reload-rules ${OUTDIR}/suricata.socket
+
+  # Give the rule reload a brief moment to settle.
+  sleep 1
```
lukashino: would be better to monitor e.g. suricata.log for full reload to make the test less flakyy-prone

live/tests/afp-ips-xdp-bypass/server.py
```diff
@@ -0,0 +1,64 @@
...
+        print(f"conn {conn_id}: received {total} bytes", flush=True)
+
+
+def main():
+    srv = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    srv.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+    srv.bind((HOST, PORT))
+    srv.listen(16)
+    print(f"listening on {HOST}:{PORT}", flush=True)
+    global conn_no
```
lukashino, line 55: why do we need conn_no as global?

live/tests/simple-http/test.yaml
```diff
@@ -0,0 +1,32 @@
+# This test will run in NFQ, AF-PACKET and IDS modes.
+mode: [nfq, afp, ids]
```
lukashino, line 2: I would suggest to name the modes per their capture mode + runmode, so:
`nfq-ips, afp-ips, afp-ids`

live/run.py
```diff
@@ -0,0 +1,1619 @@
...
+
+
+def get_mode_requires(requires: dict, mode: str) -> dict:
+    """Return test requirements with implicit mode-specific features."""
+    if mode != "nfq" or not isinstance(requires, dict):
+        return requires
+
+    mode_requires = dict(requires)
+    features = list(mode_requires.get("features", []))
+    if "NFQ" not in features:
```
lukashino, line 1454: i don't quite understand exclusion of nfq (L1442) and why you add it here.

lukashino (commented):
I generally like this work.

It doesn't require anything like "packets/time-based shutdown", supports multistep actions and checks through unix socket (in the firewall test), supports PCAP replay...

I can see this being extended with tests for e.g. Hyperscan caching or DPDK.

live/tests/afp-fw/test.yaml
jasonish: The suricatasc command is blocking, and aligns with that. I'm not even sure this is needed anymore.

jasonish:
> I can see this being extended with tests for e.g. Hyperscan caching or DPDK.

Can DPDK use Linux veth devices? Or some other "virtual" type of network adapter?

live/tests/afp-ips-xdp-bypass/server.py
jasonish: Its defined at the top, if not marked global here, where its used below will fail. Not sure if this is needed.  I've been heavy on the AI as a helper for these tests, other than making sure this script is not evil, and doing what I want it to do, I haven't given the implementation much thought.

live/run.py
jasonish: Probably an older artifact. I'm still more concerned with the shape of the tests.  This is very much a work in progress still, without much time to dedicate to it at the moment.

live/run.py
jasonish: `afp-fw` has an include but it doesn't touch `af-packet` settings.  This automatic afp include gets you in a working state.  The per test include.yaml lets you override other thigs, like firewall config, rule vars. I think they are both needed.

live/run.py
jasonish: Ah, NFQ mode requires the nfq features.  If we put requires NFQ in the test requirements, and NFQ is not built-in, we'll also skip the AFP tests, as requires is scoped at the test level, not the mode level. So this just slipos in the NFQ feature as a requirement only when needed.

jasonish: Added function comments to that effect.

live/tests/afp-fw/test.yaml
jasonish: I don't think this was needed.

jasonish:
> > I can see this being extended with tests for e.g. Hyperscan caching or DPDK.
> 
> Can DPDK use Linux veth devices? Or some other "virtual" type of network adapter?

Or could DPDK run on top of a bridge interface? As that is how `ids` mode is implemented, nothing about it is specific to AF_PACKET.

live/tests/simple-http/test.yaml
jasonish: Yeah, could use some work..  Note that testing `afp-ids` or `libpcap-ids` still uses the same environment. Nothing about the `ids` environment is tied to AF_PACKET other than thats how we choose to launch Suricata, but that could be overrided with args, or a way to provide a custom suricata start command.  Testing PF_RING in IDS mode, should we choose to would still use the same IDS environment.

Note to self: Add libpcap example to see just what is needed.

lukashino:
> > > I can see this being extended with tests for e.g. Hyperscan caching or DPDK.
> > 
> > 
> > Can DPDK use Linux veth devices? Or some other "virtual" type of network adapter?
> 
> Or could DPDK run on top of a bridge interface? As that is how `ids` mode is implemented, nothing about it is specific to AF_PACKET.

Yeah, DPDK should be able to DPDK-ize a kernel interface -- https://doc.dpdk.org/guides/nics/af_packet.html

I also experimented with virtual functions and was quite successful with that -- that could be a native way how to run DPDK on the interface itself, but it would require a tighter integration with the deployment site, as it requires at least one physical NIC (but requires no specific cabling and would allow the tests for e.g. rte_flow (HW offload)). By deployment site, I mean, e.g.,  a stage in the local CI.

lukashino:
As an FYI, I tried to see how this could run on DPDK.

We have two drivers available:

- https://doc.dpdk.org/guides/nics/af_packet.html
- https://doc.dpdk.org/guides/nics/tap.html

Both DPDK-ize (allow DPDK app to work on top of Kernel interface)
Neither of those steals the interface from the kernel in the sense that it would not be visible in the kernel afterwards.

But, as the TAP driver hooks onto the interface, it redirects the traffic in a way that it is only delivered to the DPDK app.
Should be available in the distro DPDK package in both Ubuntu and RHEL.

AF_PACKET vdev is not available in e.g. Rocky 9, but should be on e.g., Fedora 43.

Exact commands which worked for me:
```
sudo dpdk-testpmd --vdev=net_tap0,remote=enxf4a80d05a368 --no-huge --no-pci -m 128 -- -i --total-num-mbufs=2048
```

```
sudo dpdk-testpmd --vdev=net_af_packet0,iface=enxf4a80d05a368 --no-huge --no-pci -m 128 -- -i --total-num-mbufs=2048
```

TAP was promising that it supports rte_flow rules too -- but does not support COUNT action which we use in the on-going rte_flow bypass work. 
AF_PACKET therefore seems more approachable atm.

jasonish:
Replaced by https://github.com/OISF/suricata-verify/pull/3212.

PR #3210, opened 2026-07-02
Xdp bypass vlan v2

catenacyber:
## PR tests correct bypass of VLAN-tagged packets with xdp bypass

Changes:
- Test if a flow with VLAN-tagged packets is correctly bypassed

Alternative to https://github.com/OISF/suricata-verify/pull/3204


catenacyber:
This indeed needs a timeout option... will try to work on it again...

PR #3212, opened 2026-07-03
Live tests using Linux namespaces - v8

jasonish:
A harness and set of tests that uses Linux namespaces to create a server (http, etc), a client, and Suricata running as the device under test that must pass the packets between client and server in IPS/firewall modes. Optionally a Linux bridge can be used and Suricata will run in IDS mode effectively using the bridge as a network tap for passive live tests.

Main changes from previous PRs:
- Don't allow multi-mode tests, each test is a specific mode/environment
- Test must add its "runmode" args in its own args, for example `-i br0`, `--af-packet`, `-q0`. As the environment shouldn't look the run mode. For example, the `ids` mode/environment is suitable for af-packet, libpcap, pf-ring, dpdk, etc.  Additionally, the `afp` (ips environment) is suitable for dpdk ips (if that can use veth devices).

This does make tests for verbose, but we still need to add the multi-network tests, bond tests, and multi network bond tests with different thread counts per network pairs, and I was suffering from premature abstractions here.


jasonish:
@lukashino Please see https://github.com/OISF/suricata-verify/pull/3212/commits/69827cc3a2c1ae8a220b38c723dd969302d26f7d (the last commit as of this comment) for a live DPDK test.

It uses `net_af_packet0` in the IDS environment, where it just listens on an interface, passing traffic.

live/README.md
```diff
@@ -0,0 +1,238 @@
...
+- inline: Creates an inline topology, Suricata bridging all traffic between the
+  client and server. This is useful for AF_PACKET type IPS testing where
+  Suricata creates the bridge between 2 ethernet interfaces.
+
+- nfq: Creates an NFQUEUE routed topology, intercepting all traffic between the
+  client and server.
+
+The test's `environment` selects the network topology. The test's `args`
+select the Suricata capture/runmode, such as `--pcap=br0`, `--af-packet`,
+`--dpdk`, or `-q 0`. This allows a DPDK virtual PMD to run inside a tap or
```
lukashino, line 25: perhaps the last sentence is redundant, otherwise, what is `this` and why DPDK is highlighted.

live/README.md
```diff
@@ -0,0 +1,238 @@
...
+inline topology.
+
+## Linux Network Namespaces
+
+Linux network namespaces are use to provide an isolated network, the
```
lukashino, line 30: are used

lukashino: The sentence deserves full rewrite

live/run.py
```diff
@@ -0,0 +1,1884 @@
...
+
+    if action == "up":
+        with RunnerLock():
+            UP_FUNCS[environment]()
+    elif action == "down":
+        with RunnerLock():
+            do_down()
+    elif action == "status":
+        STATUS_FUNCS[environment]()
+    elif action == "shell":
```
lukashino, line 1879: I am not sure if the use is right but this command crashed it. Noting it here for now

```
sudo python3 ../live/run.py inline
usage: run.py inline [-h] {up,down,status,shell} ...

positional arguments:
  {up,down,status,shell}
    up                  bring lab up
    down                tear lab down
    status              show lab status
    shell               open shell in namespace

options:
  -h, --help            show this help message and exit


sudo python3 ../live/run.py inline shell
Cannot open network namespace "client0": No such file or directory
```

lukashino: generally I like the idea, it can be great for reproducibility, it would just need a bit more docs.

live/run.py
```diff
@@ -0,0 +1,1884 @@
...
+        if unknown:
+            raise ValueError(
+                f"{prefix} has unknown key(s): {', '.join(sorted(map(str, unknown)))}"
+            )
+        if "client" not in raw_network or "server" not in raw_network:
+            raise ValueError(f'{prefix} requires explicit "client" and "server" CIDRs')
+        bond = raw_network.get("bond", False)
+        if not isinstance(bond, bool):
+            raise ValueError(f'{prefix} "bond" must be true or false')
+        bond_mode = raw_network.get("bond-mode")
```
lukashino, line 156: just a thought, we can have bond-mode `none` to avoid specifying bond two times.

live/tests/dpdk-ids-afpacket-pmd/include.yaml
```diff
@@ -0,0 +1,47 @@
...
+      threads: 1
+      multicast: false
+      mempool-size: auto
+      mempool-cache-size: auto
+      rx-descriptors: 16
+      tx-descriptors: 16
+      copy-mode: none
+      copy-iface: none
+
+# Required when using DPDK?
```
lukashino, line 39: yes, it is.

lukashino (commented):
What else are we missing before the merge. I went through the PR, tried it locally, tests pass.

The README docs will need grammar polishing, I stopped after few comments.

Can this be a regular PR?

live/run.py
jasonish: This interactive stuff has fallen behind, might be better as a separate tool

live/run.py
jasonish: Not sure I follow? Specified 2x where?

lukashino: In the test.yaml -- https://github.com/OISF/suricata-verify/pull/3212/changes#diff-0325e9b4697454e3479b338bf0e08352152306879a4587300dc6d31c5adcd402R13

You specify:
```
topology:
  mtu: 9000
  networks:
    - client: 10.200.0.2/24
      server: 10.200.0.1/24
      bond: true
      bond-mode: balance-rr
    - client: 10.200.1.2/24
      server: 10.200.1.1/24
      bond: true
      bond-mode: balance-rr
```

Whereas, if you don't want bond-mode, you could just do:
```
topology:
  mtu: 9000
  networks:
    - client: 10.200.0.2/24
      server: 10.200.0.1/24
      bond: true
      bond-mode: balance-rr
    - client: 10.200.1.2/24
      server: 10.200.1.1/24
      bond-mode: none
```

live/run.py
lukashino: As a  separate tool that would be used both by the framework and users? Nice idea. This doesn't seem to be a high priority to me but it could be useful when developing new tests.

live/run.py
jasonish: Could just leave off bond as well I think.

PR #3215, opened 2026-07-04
tests: ipv6 hop-by-hop option length past option area

uwezkhan:
Companion test for OISF/suricata#15695 (reopened as a fresh PR per the workflow).

The hop-by-hop option loop in `DecodeIPV6ExtHdrs` gated each option on `(offset + 1 + ip6_optlen) > optslen`, which let an option whose data runs one byte past the option area through; the Router Alert memcpy then read one byte past the area. The fix bounds it on `offset + 2 + ip6_optlen`.

The pcap is an IPv6 packet (payload len 8, next header hop-by-hop) whose 8-byte hop-by-hop header carries three Pad1 options then a Router Alert declaring 2 data bytes where only 1 fits the option area.

Before the fix: the option is accepted and the RA read runs past the option area.
After the fix: the option is rejected with `ipv6.exthdr_invalid_optlen`.


catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/3305

PR #3223, opened 2026-07-08
Tag tests with protocol metadata

jasonish:
- Adds a meta.yaml to each test that has a "protocols" array
- Adds a utility to update this meta.yaml for every test using tshark
- Update run.py to tag a --meta-protocols command option

To run tests that have been identified by tshark to have smtp:

  .../run.py --meta-protocols smtp

or that have smtp or imap:

  .../run.py --meta-protocols smtp,imap

To update the metadata:

  ./util/update-meta-protocols.py

This was AI assisted then tuned up some.  Protocol renames/mappings might need some refining.

### DIFF
--- a/tests/alert-no-3whs-established/meta.yaml
+++ b/tests/alert-no-3whs-established/meta.yaml
@@ -1,0 +1,1 @@
+protocols: [data, eth, ip, tcp]
--- a/run.py
+++ b/run.py
@@ -1241,6 +1241,25 @@ def build_eve_validator():
             "{} build --release".format(cargo), cwd=os.path.join(TOPDIR, "eve-validator"),
             shell=True, env=env)
 
+def parse_protocols(values):
+    protocols = set()
+    for value in values or []:
+        for protocol in value.split(","):
+            protocol = protocol.strip()
+            if protocol:
+                protocols.add(protocol)
+    return protocols
+
+def get_test_protocols(dirpath):
+    meta_yaml = os.path.join(dirpath, "meta.yaml")
+    if not os.path.exists(meta_yaml):
+        return set()
+    config = yaml.safe_load(open(meta_yaml, "rb")) or {}
+    protocols = config.get("protocols", []) or []
+    if isinstance(protocols, str):
+        protocols = [protocols]
+    return set(str(protocol) for protocol in protocols)
+
 def main():
     global TOPDIR
     global args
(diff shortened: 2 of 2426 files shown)


catenacyber:
> each test that has a "protocols" array

What does that mean ?

> Adds a utility to update this meta.yaml for every test using tshark

Does CI check that meta.yaml is good for new tests ?

> To run tests that have been identified by tshark to have smtp:

Why use tshark instead of suricata ?


What is the value over `run.py smtp` ? and Suricata CI running every test no matter what...

jasonish:
> > each test that has a "protocols" array
> 
> What does that mean ?
> 

Each test has a `meta.yaml`, like:
```
protocols: [eth, dns]
```
so an array of protocols.

> > Adds a utility to update this meta.yaml for every test using tshark
> 
> Does CI check that meta.yaml is good for new tests ?

No, not yet. Worried about conflicting versions of tshark at the moment. So that would be the next step.

> 
> > To run tests that have been identified by tshark to have smtp:
> 
> Why use tshark instead of suricata ?

We've often wanted this feature to find stuff Suricata does not support.

> 
> What is the value over `run.py smtp` ? and Suricata CI running every test no matter what...

In that case, `smtp` has be in the test name. This would let you run every test where `smtp` is present in the pcap. For example, your just looking for some SMTP that might be untested yet, your looking for a protocol you've just started implementing.

catenacyber:
> > > each test that has a "protocols" array
> > 
> > 
> > What does that mean ?
> 
> Each test has a `meta.yaml`, like:
> 
> ```
> protocols: [eth, dns]
> ```
> 

Ok, you are adding this, but `Adds a meta.yaml to each test that has a "protocols" array` sounds cyclic dependency then

You rather meant `Adds a meta.yaml to each test and fill in there a "protocols" array`, am I right ?

Ok for tshark

tests/alert-no-3whs-established/meta.yaml
```diff
@@ -0,0 +1 @@
+protocols: [data, eth, ip, tcp]
```
catenacyber, line 1: Should it be named tshark-protocols ?

catenacyber:
> > Does CI check that meta.yaml is good for new tests ?
> 
> No, not yet. Worried about conflicting versions of tshark at the moment. So that would be the next step.

Waiting for this next step then.
I guess the tshark installed in CI will be the authoritative version

tests/alert-no-3whs-established/meta.yaml
jasonish: That might be a bit much?  If I were to do that, I'd probably want to drop the mappings we have for some common protocols so its literal to tshark-protocols, where we just want the common protocol name.

And we can expand one thark name into many for ease of use, for example:
```
    "xmpp": ["xmpp", "jabber"],
    "x509ce": ["x509", "x509ce"],
    "x509sat": ["x509", "x509sat"],
```

jasonish:
> Waiting for this next step then.

Only if I get some level of consensus this is useful.

catenacyber:
> > Waiting for this next step then.
> 
> Only if I get some level of consensus this is useful.

I do not think I would use it, but I see your use case, so I am ok

PR #3232, opened 2026-07-13
test: adds test for lua http2 library

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/6409

Suricata PR still a draft with some questions here


tests/lua/lua-http2lib-01/http-lua.rules
```diff
@@ -0,0 +1,2 @@
+alert http2:stream:request_headers any any -> any any (msg: "Test HTTP2 Lua"; lua: http2.lua; sid:1;)
+alert http1:request_line any any -> any any (msg: "Test HTTP2 Lua"; lua: http2.lua; sid:2;)
```
catenacyber, line 2: @jasonish what would be the good way to run `lua: http2.lua;` whatever the HTTP protocol ?

tests/lua/lua-http2lib-01/http2.lua
```diff
@@ -0,0 +1,19 @@
+-- simple http match on request_headers_raw module
+local http = require("suricata.http")
```
catenacyber, line 2: I think we will want for lua what we have for rules in suricata : something like :
- `http` means http whatever version
- `http1` only http1
- `http2` only http2

Should I add a function a bit like `LuaFlowAppLayerProto` to get the HTTP version of the flow alproto ?

jasonish: I think that makes sense.  Right now `suricata.http` is specifically `http1` right? Not sure we'd want to break that, even on a major version upgrade.

tests/lua/lua-http2lib-01/http-lua.rules
jasonish: I'm not sure I know enough about how the `http` abstraction over `http1` and `http2` works to comment.  I guess we need to decide if there should even be a common abstraction, or if Lua scripts should be specific enough to the version of http being processed. I guess for the command things, URLs, etc. a common abstraction would be ideal.

Would/do existing scripts need to be updated to validate what version of http they are working against to avoid making a bad call?

catenacyber: > I guess we need to decide if there should even be a common abstraction

Indeed, asked in the suricata draft PR ;-)
I think there should be one

> Would/do existing scripts need to be updated to validate what version of http they are working against to avoid making a bad call?

We can do multiple scenarios here :
- either make lua `http` work for http2, like we did `alert http` now works transparently for http2, and add functions that could check the http version later in script
- introduce a new `get_tx_whatever_the_http_version` function, and keep the `get_tx` for HTTP1 only, so current scripts remain http1 only, but it is is easy to upgrade them to do both HTTP versions

tests/lua/lua-http2lib-01/http2.lua
catenacyber: I think we want to break/extend this for 9...

catenacyber:
Replaced by https://github.com/OISF/suricata-verify/pull/3315

PR #3236, opened 2026-07-17
tests: add firewall bypass test

jasonish:
This tests that a "bypass" in a firewall rule fails out and Suricata
exits with exit code 0.

It also contains the checks for if bypass in firewall mode were to be
allowed. Based on a simple revert of commit
a78300740818dfa634d4e19a50d7051544ee655a in Suricata.

Ticket: #8459

PR #3237, opened 2026-07-17
tests: allow bypass keyword for firewall rules - v1

jufajardini:
Changes:
- adjust the test from https://github.com/OISF/suricata-verify/pull/3236 for bypass enabled for firewall
- rename firewall tests that had duplicated numbers
- update bypass ban test's error message

TODOs:
- add test with `accept: packet` -> `bypass` shouldn't be valid in such a case
- test with a single packet and td drop rule?

## Ticket

https://redmine.openinfosecfoundation.org/issues/8459


jufajardini:
Followed by: https://github.com/OISF/suricata-verify/pull/3257

PR #3239, opened 2026-07-21
tests: bdat last message followed by a new message

jasonish:
The final BDAT chunk and its reply must complete the transaction so a
following MAIL FROM starts a new transaction instead of merging into the
previous one. Covers a non-empty final chunk, a zero-length final chunk,
and a multi-chunk message.

Ticket: #8741


victorjulien (approved):

victorjulien:
Merged in #3273, thanks!

PR #3243, opened 2026-07-22
Xdp bypass vlan v2.2

catenacyber:
## Ticket

Also implements https://redmine.openinfosecfoundation.org/issues/7413

Alternative to https://github.com/OISF/suricata-verify/pull/3204


replay/xdp-bypass-vlan/suricata.yaml
```diff
@@ -0,0 +1,17 @@
...
+---
+
+outputs:
+  - eve-log:
+      enabled: yes
+      filetype: regular
+      filename: eve.json
+      types:
+        - stats
+        - flow
```
catenacyber, line 11: flow log are better reproducible than stats

replay/xdp-bypass-vlan/suricata.yaml
```diff
@@ -0,0 +1,17 @@
...
+
```
catenacyber, line 12: no need to decrease the tcp byapssed timeout thanks to suricata commit https://github.com/OISF/suricata/pull/15889/changes/cf66c321d0f8846c54d858c02dce97dadad50ea3

replay/xdp-bypass-vlan/test.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - --set af-packet.0.xdp-filter-file=${SRCDIR}/ebpf/xdp_filter.bpf
+
+timeout: 1
+
+checks:
+  - filter:
+      count: 0
+      match:
+        event_type: flow
+        flow.bypassed.pkts_toserver: 0
```
catenacyber, line 22: @jasonish @inashivb do we have a better SV way to test an integer key is `>0` ?

jasonish: No, to my live work I've added __gt, __gte, etc as being live something __gte is what you need.  But it would make sense to add to S-V IMO.

catenacyber: Is there a commit to cherry-pick ?

jasonish: No, it was done in the live runner, then refactored out to a common module.  So that branch has it, but not in a cherry-pickable way.

jasonish: https://github.com/OISF/suricata-verify/pull/3246

PR #3246, opened 2026-07-23
run.py: allow gt, gte, lt and lte comparisons

jasonish:
From the README updates, one can now do:

```yaml
  - stats:
      # Check values in the last stats event in eve.json. Keys are relative to
      # the stats object. Values must match exactly unless a comparison suffix
      # is used.
      decoder.pkts: 42
      decoder.ethernet: 42
      # Numeric comparisons can also be used:
      decoder.bytes.__gte: 128 
```


catenacyber (approved):
Looks a good addition to me

victorjulien (approved):

victorjulien:
Merged in #3289, thanks!

PR #3250, opened 2026-07-26
config: test deterministic glob include ordering (v2)

ssam18:
**Depends on OISF/suricata#15937** (the `include:` glob feature, Redmine #8427), which supersedes #15574. This test exercises that feature, so CI here will stay red until it is merged into `OISF/suricata`: suricata-verify CI builds suricata from the official branches (`master`/`main-7.0.x`/`main-8.0.x`), none of which have the feature yet. The pair is validated in the suricata PR's CI (which builds the feature and runs this exact branch via `SV_BRANCH`) and locally.

Previous PR: #3151

Adds `config-includes-glob-order`, a test for the `include:` glob-pattern support added in OISF/suricata (Redmine #8427).

`glob.yaml` includes `*-order.yaml`, which matches three drop-in files (`01-order.yaml`, `50-order.yaml`, `99-order.yaml`). Each defines its own unique marker key (`glob-order-seen-NN`). Because `glob(3)` returns matches in sorted (lexicographic) order, the files are merged in ascending numeric-prefix order. The common `01..99` drop-in directory convention.

Using `--dump-config`, the test asserts:
- all three matched files were included (one unique marker key per file), and
- the marker keys appear in sorted-filename order, observable because the config tree preserves insertion order.

No key is defined twice, so the test does not depend on duplicate-key override behaviour. What it pins is the sort order, which is what the numbered drop-in convention relies on in practice.

The glob config is kept in `glob.yaml` and run via `command:` rather than a shipped `suricata.yaml`, because suricata-verify runs `suricata -c suricata.yaml --dump-config` during test setup, before the `min-version` check.

This addresses review feedback on the suricata PR (victorjulien/jasonish) requesting the include tests live in suricata-verify, with an explicit check that include ordering is deterministic.

### Changes since #3151

- Added a second `requires: script:` probe, `grep '^#define HAVE_GLOB_H 1' src/autoconf.h`, so the test is skipped on builds without `glob(3)`.

  Glob expansion needs `glob(3)`, which Windows builds do not have. Per review on the suricata side (jasonish), that platform now skips an include pattern with a warning and loads none of the matched files, so the ordering checks cannot hold there. Probing the *build* rather than the platform keeps the test running everywhere `glob(3)` is available, including macOS and the BSDs, unlike a `requires: os: linux` gate.

  Previously the test avoided failing on Windows only incidentally: `requires: script:` runs under `cmd.exe` there, where the `> /dev/null` redirect fails and makes any probe return false. That is fragile and would break the moment suricata-verify fixed the redirect.

- Documented the second probe in `README.md`.

No change to the test's configs, checks, or what it proves.

Ticket: https://redmine.openinfosecfoundation.org/issues/8427

Verified locally against a `9.0.0-dev` build with the glob feature: `===> config-includes-glob-order: OK`. Verified the skip path by undefining `HAVE_GLOB_H` in a local build: `===> config-includes-glob-order: SKIPPED: requires script returned false`.

### DIFF
--- a/tests/config-includes-glob-order/01-order.yaml
+++ b/tests/config-includes-glob-order/01-order.yaml
@@ -1,0 +1,3 @@
+%YAML 1.1
+---
+glob-order-seen-01: present
--- a/tests/config-includes-glob-order/50-order.yaml
+++ b/tests/config-includes-glob-order/50-order.yaml
@@ -1,0 +1,3 @@
+%YAML 1.1
+---
+glob-order-seen-50: present
--- a/tests/config-includes-glob-order/99-order.yaml
+++ b/tests/config-includes-glob-order/99-order.yaml
@@ -1,0 +1,3 @@
+%YAML 1.1
+---
+glob-order-seen-99: present
--- a/tests/config-includes-glob-order/README.md
+++ b/tests/config-includes-glob-order/README.md
@@ -1,0 +1,44 @@
+# config-includes-glob-order
+
+Verifies that a glob pattern in an `include:` directive expands to all matching
+files and includes them in a deterministic (sorted) order.
+
+`glob.yaml` includes `*-order.yaml`, which matches three drop-in files
+(`01-order.yaml`, `50-order.yaml`, `99-order.yaml`). Each file defines its own
+unique marker key (`glob-order-seen-NN`). Because `glob(3)` returns matches in
+sorted (lexicographic) order, the files are merged in ascending numeric-prefix
+order.
+
+The test runs suricata against `glob.yaml` with `--dump-config` (via `command:`)
+and asserts that:
+
+- all three matched files were included (one unique marker key per file), and
+- the marker keys appear in the dump output in sorted-filename order. The
+  config tree preserves insertion order, so the position of each key in the
+  output reflects the order the files were merged. No key is defined twice, so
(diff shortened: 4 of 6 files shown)


ssam18:
This is v2 of #3151, which @jasonish approved there ("Test looks OK"). A new PR was needed because the branch of an open PR should not be updated.

The only delta from the approved version is your own follow-up suggestion:

> Depending on OISF/suricata#15574 (review), this might need a not runs on Windows requirement.

Rather than `requires: os: linux`, it probes the build: `grep '^#define HAVE_GLOB_H 1' src/autoconf.h`. Windows is what actually lacks `glob(3)`, so probing the build keeps the test running on macOS and the BSDs, where it works. Plus a README paragraph explaining it.

The drop-in configs, `glob.yaml`, and every check are unchanged from the version approved in #3151, the ordering assertions @catenacyber and @jasonish discussed there are untouched.

Verified locally: `===> config-includes-glob-order: OK` against a build of OISF/suricata#15937, and `SKIPPED: requires script returned false` against the same build with `HAVE_GLOB_H` undefined.

ssam18:
Superseded by #3281: same test rebased onto current master (no changes to the test itself), resubmitted as a new PR per the workflow since this branch had to be updated. The companion suricata PR is now OISF/suricata#16031.

PR #3252, opened 2026-07-28
firewall: test the default policy v2

lukashino:
Follow-up of https://github.com/OISF/suricata-verify/pull/3251

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8712

Describe changes:
- Update to the new config format
- Add extra tests

PR #3229, opened 2026-07-10
Fw backports/v7.2

victorjulien:
Not for merge: backport branch for tx substate work related tests. Also includes #3227.

PR #3255, opened 2026-07-29
Backport/8.0.x: firewall: test the default policy v1

lukashino:
Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8770

Patchset covers new commits from #3252 but is rebased onto #3229, and enables 8.0.7 for the new tests.


catenacyber:
Why is main CI red ?

PR #3285, opened 2026-08-18
workflows: eol 7.0.x; use rust 1.89 - v2

jasonish:
Builds on https://github.com/OISF/suricata-verify/pull/3283.

- Update Ubuntu to 24.04
- Use Rust 1.89
- Update checkout action version

PR #3289, opened 2026-08-19
next/140/20260819/v1

victorjulien:
Staging:
 - #3236
 - #3246
 - #3285


jufajardini (approved):
Consistent with original, approved PRs.
CI passes.

If we want to merge bypass tests showing bypass in fw rules would fail right now, before we add that feature in firewall, then all looks good to me.

PR #3292, opened 2026-08-20
logging: test for multi-threaded double free in deinit v1

lukashino:
Suricata crashes on double free without the proposed fix as multiple threads free the same shallow-copied variables.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8861


victorjulien:
For merge we need a test that passes on 8.0.x.

lukashino:
follow-up in https://github.com/OISF/suricata-verify/pull/3301

PR #3294, opened 2026-08-20
run: allow suricatasc to be specified with an environment var

jasonish:
Needed for a new Rust+ASAN CI build configuration.

### DIFF
--- a/run.py
+++ b/run.py
@@ -1383,13 +1383,15 @@ def main():
     HAS_SURICATA_SC = None
     global suricatasc_bin
     if not args.skipunixsocket:
-        sc_path = [".", "rust", "target", "release", "suricatasc"]
-        if "DEBUG" in suricata_config.features:
-            sc_path[3] = "debug"
-        cargo_build_target = os.environ.get("CARGO_BUILD_TARGET")
-        if cargo_build_target != None :
-            sc_path.insert(3, cargo_build_target)
-        suricatasc_bin = os.path.join(*sc_path)
+        suricatasc_bin = os.environ.get("SURICATASC")
+        if suricatasc_bin is None:
+            sc_path = [".", "rust", "target", "release", "suricatasc"]
+            if "DEBUG" in suricata_config.features:
+                sc_path[3] = "debug"
+            cargo_build_target = os.environ.get("CARGO_BUILD_TARGET")
+            if cargo_build_target != None :
+                sc_path.insert(3, cargo_build_target)
+            suricatasc_bin = os.path.join(*sc_path)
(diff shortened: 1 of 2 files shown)


run.py
```diff
@@ -1383,13 +1383,15 @@ def main():
...
-        cargo_build_target = os.environ.get("CARGO_BUILD_TARGET")
-        if cargo_build_target != None :
-            sc_path.insert(3, cargo_build_target)
-        suricatasc_bin = os.path.join(*sc_path)
+        suricatasc_bin = os.environ.get("SURICATASC")
+        if suricatasc_bin is None:
+            sc_path = [".", "rust", "target", "release", "suricatasc"]
+            if "DEBUG" in suricata_config.features:
+                sc_path[3] = "debug"
+            cargo_build_target = os.environ.get("CARGO_BUILD_TARGET")
```
jasonish, line 1391: Setting `CARGO_BUILD_TARGET` is another way, but in this case, I don't want to alter the cargo build environment as it can affect the build of the included eve validation program.

inashivb (approved):
Seems to be picked up and working as intended. 🚀

catenacyber:
Why do you need this ?
Do not you want to use the suricatasc you just built along the suricata you just built ?

jasonish:
> Why do you need this ? Do not you want to use the suricatasc you just built along the suricata you just built ?

When Rust is built with ASAN, it requires using a --target string which differs from the default.  So it is to run with the suricatasc/suricata we just built.

run.py
catenacyber: Could you use a `SURICATASC_BUILD_TARGET` ? if this is just about the target.

I find rust's way hard to find where my binary was built :-/

jasonish: This is how we pick up the target between release and debug builds. Its not related to this PR.

catenacyber (approved):
I do not like this, as it is not elegant but it does the work

And the problem I do not like is how rust picks different subdirectories to build binaries

So, approving here since I do not have a better idea

jasonish:
> I do not like this, as it is not elegant but it does the work
> 
> And the problem I do not like is how rust picks different subdirectories to build binaries
> 
> So, approving here since I do not have a better idea

I think it's the auto-picking you don't like, which is pre-existing.  This new one is to handle a specific case where the caller specifically knows where the binaries exist that cannot automatically be determined.  I think that will be clearer when I can do the Suricata side of the PR. Your point is valid, but I don't see how it relates to this PR.

victorjulien:
Merged in #3371, thanks!

PR #3298, opened 2026-08-21
Live testing framework - v9

jasonish:
Rework of https://github.com/OISF/suricata-verify/pull/3212 remove the
interactive features that lagged and didn't work properly. They may be
re-introduced as a side-car tool.

Address README and DPDK comments.

Previous PR: https://github.com/OISF/suricata-verify/pull/3212

Should cover all the live tests we do in GitHub CI from the main Suricata repo.


lukashino:
Passes well. Some tests were flaky but e.g. disabling IPv6 helps make them reproducible (tried 10 times in row). I'd like to refine the commands so IPv6 doesn't need to disabled fully but we only disable the noisy parts.

jasonish:
> Passes well. Some tests were flaky but e.g. disabling IPv6 helps make them reproducible (tried 10 times in row). I'd like to refine the commands so IPv6 doesn't need to disabled fully but we only disable the noisy parts.

Flaky where. Locally? On Github?

jasonish:
> Passes well. Some tests were flaky but e.g. disabling IPv6 helps make them reproducible (tried 10 times in row). I'd like to refine the commands so IPv6 doesn't need to disabled fully but we only disable the noisy parts.

And where are you disabling IPv6 to get more consistent results?

lukashino:
This test failed.

```
~/suricata-verify/suricata $ sudo ../live/run.py 
===> [inline/afp-fw] Setting up inline environment
===> [inline/afp-fw] Running before script
===> [inline/afp-fw] Starting Suricata
===> [inline/afp-fw] Starting Server
===> [inline/afp-fw] Running Client
===> [inline/afp-fw] Stopping Server
===> [inline/afp-fw] Stopping Suricata
===> [inline/afp-fw] FAIL ❌
  shell command failed with exit code 1: 'jq -e -s \'\n  [.[] | select(.event_type == "stats") | .stats][-1] as $stats\n  | ($stats.firewall.accepted + $stats.firewall.blocked) as $verdicts\n  | ($stats.firewall.blocked > 0)\n    and ($stats.capture.kernel_packets == $verdicts)\n\' ./eve.json >/dev/null\n'
```

lukashino:
btw, your latest made the test consistent in my environment, thanks.

Btw2:
From operational perspective, I liked the idea of "test once, run multi-topology" what was in v8, I think. Right now, if I got it correctly, you specify a test for each topology, am i correct (e.g. simple-http-*)? 

If that is true, then I would avoid specifying the topology mode in the squared brackets as you cannot then just double click and run that specific test again.

```
===> [inline/afp-fw] Setting up inline environment
===> [inline/afp-fw] Running before script
===> [inline/afp-fw] Starting Suricata
===> [inline/afp-fw] Starting Server
===> [inline/afp-fw] Running Client
===> [inline/afp-fw] Stopping Server
===> [inline/afp-fw] Stopping Suricata
===> [inline/afp-fw] OK ✅

PASS: 1
FAIL: 0
SKIP: 0

$ sudo python3 ../live/run.py inline/afp-fw

PASS: 0
FAIL: 0
SKIP: 0
```

jasonish:
Replaced by https://github.com/OISF/suricata-verify/pull/3378.

PR #3302, opened 2026-08-23
tests: add test for bug 8856

victorjulien:
Test reproduces DecodePPPOESession over-read with 7-byte PPPoE session header and non-zero pppoe_length. Verifies decoder.event.pppoe.pkt_too_small stat is raised.

Ticket: #8856.

## Ticket

changes to the ticket status:


catenacyber:
Why is CI green ?

victorjulien:
> Why is CI green ?

It only triggers on an ASAN enabled build.

inashivb:
> > Why is CI green ?
> 
> It only triggers on an ASAN enabled build.

We do build Suricata on ubuntu-24 job with ASAN.

catenacyber:
I do not reproduce locally with ASAN on macOS

victorjulien:
Must have been mixing things up. I can't either, and neither with valgrind.

victorjulien:
So this doesn't trigger the bug but it's still useful to have a test for this small packet condition.

tests/bug-8856-pppoe-session/test.rules
```diff
@@ -0,0 +1 @@
+# empty rules for this test
```
catenacyber, line 1: no need for this file

catenacyber (approved):
This test is ok, passes and tests a good behavior, even if it is not really about the suricata PR linked (in the sense that we do not need the suricata PR to make it pass)

jasonish (approved):

victorjulien:
Merged in #3341, thanks!

PR #3313, opened 2026-08-25
add tls sni auto-accept-prior-hook repro bug validation

yashda:
## Ticket
Add a firewall-mode test showing that the auto-accept-prior-states (`<`) notation on a catch-all accept does not cover the prior app-layer hook when a lower-SID same-hook drop rule with a matching prefilter leads the candidate list.
    
The rules drop TLS SNI "www.google.com" (sid:200) and accept all other SNI via `accept:flow tls:<client_hello_done` (sid:201), without an explicit `accept:hook tls:client_in_progress`. The expected result is a drop by sid:200 with an alert; the actual result is a default-app-policy drop at client_in_progress with sid:200 never firing.
    
The test asserts the expected behaviour and therefore currently fails, documenting the defect. Reuses tests/tls/tls-client-hello-frag-01.

Redmine: https://redmine.openinfosecfoundation.org/issues/8944


catenacyber:
Is there a redmine ticket that goes with this test ? A suricata PR ?

yashda:
Updated the description with the Redmine ticket.

PR #3319, opened 2026-08-25
test(firewall): add IMAP app-layer firewall regression test

yashda:
IMAP is detection-only in Suricata: src/app-layer-imap.c registers protocol detection patterns but no parser, transactions or progress states. A firewall rule that hooks an IMAP app-layer state therefore loads without error yet can never match, and the app-layer firewall never engages for an IMAP flow.

This test loads a drop:flow rule on imap:request_started intending to block every IMAP flow and asserts the expected behaviour:
  - the flow is still detected as app_proto imap (control), and
  - the drop rule fires and blocks the flow (alert sid 2000, action blocked),
  - the flow is actually dropped (firewall drop_reason.rules and blocked >= 1).

It fails on affected builds (reproducing the bug) and will pass once the underlying issue is fixed.

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8948

### DIFF
--- a/tests/firewall/firewall-imap/firewall.rules
+++ b/tests/firewall/firewall-imap/firewall.rules
@@ -1,0 +1,17 @@
+# Proof that firewall-mode app-layer rules are inert for IMAP.
+#
+# IMAP in Suricata is detection-only. src/app-layer-imap.c registers protocol
+# detection patterns but no parser, no transactions and no progress states.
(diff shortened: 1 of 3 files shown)


catenacyber:
Thanks, is there a ticket about this test ?

PR #3324, opened 2026-08-26
tests/quic: add five-tuple reuse SNI test

jiangshaoqi:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8776

## Description

Add a Suricata-Verify test for QUIC SNI inspection when a UDP five-tuple is
reused for a new QUIC connection.

The first ClientHello in the PCAP contains SNI `Google.com` but does not
progress to a client Handshake packet. A second connection on the same
five-tuple contains SNI `www.dongbeirain.fun`, progresses to the Handshake
stage, and exchanges application data.

Without the fix, Suricata logs only the first SNI. It misses both the QUIC
event for the second SNI and the matching alert. The corresponding rebased
fix is OISF/suricata#16209.

## Validation

- OISF/suricata `main`: the second-SNI and alert checks fail as expected.
- Patched Suricata: the new test passes.
- Patched Suricata: all QUIC tests pass (20 passed, 0 failed, 1
  version-dependent skip, including this new test).


catenacyber:
Why are there the 2 last packets with reused stream offset ?

jiangshaoqi:
> Why are there the 2 last packets with reused stream offset ?

Those last two packets are PTO probes from the server for the first, unfinished connection. Since its Initial was not acknowledged, the server retransmits the same CRYPTO data (offset 0, length 90) in two new Initial packets with packet numbers 1 and 2. That is why Wireshark reports a reused offset; it is expected retransmission behavior.

PR #3329, opened 2026-08-27
Bug/8954 action scope verify v2

lukashino:
Follow-up of https://github.com/OISF/suricata-verify/pull/3323

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8954

Describe changes:
v2:
- tests for `pass` keyword (in)compatibility added

PR #3331, opened 2026-08-28
tests: tests for scope inheritance in rules/policies + pass&config action updates v3

lukashino:
Follow-up of https://github.com/OISF/suricata-verify/pull/3329

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8954

Describe changes:
v3:
- rebased
- added a test for `config` action in default policies to make sure it is banned by Suricata

v2:
- tests for pass keyword (in)compatibility added

### DIFF
--- a/tests/firewall/firewall-action-pass-requires-accept-policies-invalid/test.yaml
+++ b/tests/firewall/firewall-action-pass-requires-accept-policies-invalid/test.yaml
@@ -1,0 +1,13 @@
+requires:
+  min-version: 9
+  pcap: false
+
+args:
+  - --engine-analysis
+
+exit-code: 1
+
+checks:
+  - shell:
+      args: "grep -c \"'pass' is only supported as a secondary action for 'accept'\" suricata.log | xargs"
+      expect: 1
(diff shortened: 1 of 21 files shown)


tests/firewall/firewall-action-pass-requires-accept-policies-invalid/test.yaml
```diff
@@ -0,0 +1,13 @@
...
+  pcap: false
+
+args:
+  - --engine-analysis
+
+exit-code: 1
+
+checks:
+  - shell:
+      args: "grep -c \"'pass' is only supported as a secondary action for 'accept'\" suricata.log | xargs"
```
victorjulien, line 12: we see this fail in CI for Macos and FreeBSD. In general lets try to avoid grep and friends as much possible for portability.

lukashino: well, the problem was that I used reject policy/rules without feature-specifying it in the test itself.
Leaving it for now as it is matching an error message.

victorjulien (approved):

victorjulien:
Merged in #3355, thanks!

PR #3333, opened 2026-08-28
tests: test dcerpc.iface any_frag first fragment handling - v2

jasonish:
Ticket: https://redmine.openinfosecfoundation.org/issues/8577


tests/dcerpc/issue-8577-dcerpc-iface-any-frag/test.yaml
```diff
@@ -0,0 +1,112 @@
...
+pcap: input.pcap
+
+args:
+  - -k none
+
+checks:
+  # Without any_frag the request PDU must have PFC_FIRST_FRAG set, the
+  # flags of the BIND are not relevant. 40001 (BIND with pfc_flags 0),
+  # 40002 (fragmented request) and 40005 match, 40003 (request with
+  # pfc_flags 0) and 40004 (request with only PFC_LAST_FRAG) do not.
```
catenacyber, line 13: I do not understand the purpose of this `any_frag` keyword

I would simply remove it and consider every `dcerpc.iface` keyword as if it had any_frag (and I would add a dcerpc.flags keyword)

The pcap is synthetic and I wonder how a real server would answer to a request with only PFC_LAST_FRAG

PR #3335, opened 2026-08-29
tests: verify unix command socket removal in hostbit-add-ip6

e-cite:
As suggested in review of OISF/suricata#16117: add a shell check to the existing hostbit-add-ip6 test verifying that the command socket file is removed after the clean 'suricatasc -c 'shutdown'' that the test harness itself issues at the end of a unix-commands test.

This covers the 'suricatasc shutdown' path. Covering a SIGTERM-triggered shutdown requires additional runtime support in the test framework and is thus not part of this test.

Ticket: 8799

## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8799

### DIFF
--- a/tests/unix-socket/hostbit-add-ip6/test.yaml
+++ b/tests/unix-socket/hostbit-add-ip6/test.yaml
@@ -10,3 +10,10 @@ checks:
     count: 1
     match:
       return: OK
+
+# The test harness sends "shutdown" over the unix socket after the
+# commands above. By the time this check runs, Suricata has fully
+# exited, so the command socket file must be gone.
+# Ticket: https://redmine.openinfosecfoundation.org/issues/8799
+- shell:
+    args: test ! -e socket
--- a/tests/unix-socket/hostbit-add-ip6/README.md
+++ b/tests/unix-socket/hostbit-add-ip6/README.md
@@ -1,6 +1,7 @@
 # Test Description
 
-Test `add-hostbit` unix-socket command with ip6 address
+- Test `add-hostbit` unix-socket command with ip6 address
+- Test socket file to be properly closed and removed after Suricata shutdown
 
 ## PCAP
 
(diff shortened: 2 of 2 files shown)


tests/unix-socket/hostbit-add-ip6/test.yaml
```diff
@@ -10,3 +10,10 @@ checks:
     count: 1
     match:
       return: OK
+
+# The test harness sends "shutdown" over the unix socket after the
+# commands above. By the time this check runs, Suricata has fully
+# exited, so the command socket file must be gone.
+# Ticket: https://redmine.openinfosecfoundation.org/issues/8799
+- shell:
+    args: test ! -e socket
```
catenacyber, line 19: This requires min-version 9

catenacyber: Otherwise, test looks good

catenacyber (changes requested):
Needs the min-version 9 to keep CI green for 8

tests/unix-socket/hostbit-add-ip6/test.yaml
e-cite: Thanks for the review, Philippe. I will add that tomorrow.

Just one question in advance: There is a Suricata 8.0.x backport ticket [#8802](https://redmine.openinfosecfoundation.org/issues/8802) linked. Doesn't this mean min-version would be 8 instead of 9 or do I get things wrong?

jasonish:
> Needs the min-version 9 to keep CI green for 8

We always merge to `main` first, so there is a window where the fix is not in `8.0`, and both versions run against the same `suricata-verify` branch, which we like to keep passing.  When backports are done, the tests are then updated.

e-cite:
> > Needs the min-version 9 to keep CI green for 8
> 
> We always merge to `main` first, so there is a window where the fix is not in `8.0`, and both versions run against the same `suricata-verify` branch, which we like to keep passing. When backports are done, the tests are then updated.

Thanks for the quick reply and your explanation. I will add the min-version 9 tomorrow.

e-cite:
Follow-up PR now at: https://github.com/OISF/suricata-verify/pull/3346

PR #3336, opened 2026-08-29
backport: support issue 8845 tests for 8.0.x 

jlucovsky:
Add bug-8845-03, gated with min-version 8.0.7 and lt-version 9.0. It loads the same three rules as bug-8845-02, and on 8.0.x all three load while the two shifting by a literal 64 draw a warning naming sid:1 and sid:2. bug-8845-02 stays at min-version 9.0, where those two rules are rejected and the run exits 1. The two cases need separate directories because exit-code is one value for the whole test.

Lower bug-8845-01 from min-version 9.0 to 8.0.7 now that the right shift guard it exercises is on 8.0.x as well. Add a check there that a rule whose rvalue names a variable draws no warning at load: that count comes from a byte_extract and is known only once the rule runs.

Issue: 8845
Issue: 8902

## Ticket

changes to the ticket status:

### DIFF
--- a/tests/bug-8845/bug-8845-01/test.yaml
+++ b/tests/bug-8845/bug-8845-01/test.yaml
@@ -1,7 +1,15 @@
+# The runtime guard is on 8.0.x as well, so this runs on both branches.
 requires:
-  min-version: 9.0
+  min-version: 8.0.7
 
 checks:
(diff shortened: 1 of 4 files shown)


victorjulien (approved):

victorjulien:
Merged in #3337, thanks!

PR #3338, opened 2026-08-31
reference: cover reference key handling

jlucovsky:
Add tests for reference keys that are not in reference.config. reference-05 checks that such a key logs "undefined" joined to the reference value. reference-07 checks the ordering behind it: the first rule to name a key registers it, so a second rule reusing that key resolves through the ordinary lookup rather than the fallback, and the two paths have to agree.

reference-06 checks that --strict-rule-keywords=reference rejects the same key. reference-08 covers three rejections -- a key of 64 characters, which cannot fit REFERENCE_SYSTEM_NAME_MAX with its terminator, a reference with no value, and one with no comma. reference-09 takes the accepting side of that bound at 63 characters, separate from reference-08 because run.py passes --init-errors-fatal, so one run cannot hold both a rule that fails to load and a rule that alerts.

Issue: 8849

## Ticket

changes to the ticket status:

### DIFF
--- a/tests/reference-05/suricata.yaml
+++ b/tests/reference-05/suricata.yaml
@@ -1,0 +1,13 @@
+%YAML 1.1
+---
+
+outputs:
+  - eve-log:
+      enabled: yes
+      filetype: regular #regular|syslog|unix_dgram|unix_stream|redis
+      filename: eve.json
+      types:
+        - alert:
+            metadata:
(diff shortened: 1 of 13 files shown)


victorjulien (approved):

victorjulien:
Merged in #3345, thanks!

PR #3340, opened 2026-08-31
Rfb test/v2

victorjulien:
## Ticket

changes to the ticket status:


tests/rfb-max-8731-toskip/make_pcap.py
```diff
@@ -0,0 +1,146 @@
+#!/usr/bin/env python3
+"""
+Pcap generator for the rfb-max-8731-toskip SV test (ticket #8731, PR 16145).
+
+Traffic:
+  - RFB handshake (versions 3.8)
+  - server sends security-types count = 0 followed by a failure reason of
+    REASON_LEN (5000) bytes in a SINGLE segment (larger than the
+    max-string-length cap of 1024 used in test.yaml)
+  - server sends a second, well-formed failure reason of 10 bytes
```
catenacyber, line 10: Not sure this is the valid behavior for Suricata

RFC states

> The server closes the connection after sending the reason-string.

For a better test, I think we should have 2 different flows : one with a big reason, and another with a small reason

tests/rfb-none-secresult/make_pcap.py
```diff
@@ -0,0 +1,172 @@
...
+
+Bug: the RFB parser skips the security result on the "None" (security
+type 1) path. After the client selects type 1 it expects the client's
+shared flag and then the ServerInit, but a spec-compliant RFB 3.8 server
+sends a security result first (0 = OK, 1 = failure + reason). The
+unconsumed result bytes collide with the parser's state machine (the
+result arrives on ToClient while the state is TSClientInit, a
+ToServer-only state), trigger the "Invalid state for response"
+catch-all, log a confused_state anomaly and discard the rest of the
+handshake (shared flag + ServerInit).
```
catenacyber, line 13: If this is true, it deserves its own ticket

catenacyber:
I propose #3342

tests/rfb-max-8731-toskip/make_pcap.py
victorjulien: I think the point of the test is to show that the big reason is miscounted, leading to the reason following it be mishandled?

catenacyber: > I think the point of the test is to show that the big reason is miscounted

yes

> leading to the reason following it be mishandled

I do not think so : I think this rather a side effect of our state machine.
A server RFB implementation will close the connection after the error reason. Suricata does not, and we keep the state `TCFailureReason` and we should maybe go to `RFBGlobalState::Skip`

tests/rfb-max-8731-toskip-vncauth/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. one security type (VNC auth, 2); the client selects it; a 16-byte
+   challenge and 16-byte response complete the VNC authentication;
+2. a security result of `1` (failure), sent as a 4-byte big-endian
+   value — the layout both dissectors model: Suricata's parser reads it
+   with `be_u32` and tshark's VNC dissector with `tvb_get_ntohl`
+   (packet-vnc.c, VNC_SESSION_STATE_SECURITY_RESULT);
+3. a failure reason of 5000 bytes (cap 1024 ⇒ truncated,
+   `rfb.too_long_string`) in the SAME segment as the result;
+4. a well-formed short failure reason (`0123456789`, 10 bytes) in a
```
catenacyber, line 28: Same as for rfb-max-8731-toskip : I am not sure this is a valid test

RFC 6143 section [7.1.3](https://www.rfc-editor.org/info/rfc6143/#section-7.1.3) states

> If unsuccessful, the server sends a string describing the reason for the failure, and then closes the connection:

So, I think the best behavior for Suricata would be to change the state machine to a new state like `Do NotExpectAnyMoreData` and raise an event if there is more data (instead of remaining in the SecurityResult state and having many of these)

tests/rfb-none-secresult/README.md
```diff
@@ -0,0 +1,63 @@
...
+  only present when the handshake was parsed end to end;
+- zero `anomaly` events of type `confused_state`.
+
+Against a parser with the bug, the first check matches 0 events and the
+second matches 1 (the logged `rfb` event contains only the versions and
+`security_type`).
+
+## Related issues
+
+Bug found during review of PR 16145 (2026-08); no Redmine ticket yet.
```
catenacyber, line 60: This deserves its own ticket independent of the security issue

tests/rfb-max-8731-reason-coalesced/README.md
```diff
@@ -0,0 +1,71 @@
...
+auth — spec-compliant path, independent of the separate None-path
+security-result issue): version exchange, 1 security type (VNC auth)
+offered and selected, 16-byte challenge / 16-byte response, then ONE
+server segment containing:
+
+- security result: failure (1), 4-byte big-endian — the width modeled
+  by both the parser's `be_u32` and tshark's VNC dissector;
+- first failure reason: 2048 bytes of 'A' (> cap 1024 set in
+  test.yaml): truncated, `rfb.too_long_string`;
+- second failure reason: 10 bytes "0123456789", well-formed — must be
```
catenacyber, line 42: As for other tests, I do not think Suricata should allow 2 failure reason in one flow

tests/rfb-none-secresult/README.md
catenacyber: https://redmine.openinfosecfoundation.org/issues/9037

catenacyber:
#3352 was merged and https://redmine.openinfosecfoundation.org/issues/9037 is some follow-up work which references this PR

PR #3344, opened 2026-09-01
tests/firewall: cover LTE candidate policy regressions - v1

jasonish:
Tests for https://github.com/OISF/suricata/pull/16156.

The idea is that these pass on `main`, but fail with the PR.


catenacyber:
> The idea is that these pass on main, but fail with the PR.

So the PR needs a fix ?

PR #3311, opened 2026-08-25
tests/ldap: add BindRequest keyword tests

0x-0ddc0de:
Adds tests for the LDAP BindRequest detection keywords of Redmine #7536:

BindRequest protocol version
authentication choice (simple / sasl)
SASL mechanism
SASL credentials

The tests use generated pcaps and include positive and negative cases to verify that the SASL sticky buffers inspect only the corresponding SASL fields.

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7536

PR #3354, opened 2026-09-04
tests/ldap: add BindRequest keyword tests

0x-0ddc0de:
Previous [PR](https://github.com/OISF/suricata-verify/pull/3311). This PR only amends the commit message (no code changes).

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/7536


tests/detect-ldap-bind-auth/test.yaml
```diff
@@ -0,0 +1,51 @@
+requires:
+  min-version: 9
+
+args:
+  - -k none
+
+pcap: ldap-bind-auth.pcap
+rules: test.rules
```
jufajardini, line 8: If the pcap file is in the same directory as the test itself, the `pcap` option isn't needed here (it's used when a test uses another test's pcap). Same applies to the rules file.

tests/detect-ldap-bind-auth/test.yaml
```diff
@@ -0,0 +1,51 @@
+requires:
+  min-version: 9
```
jufajardini, line 2: This has to be present in all test.yaml added by these tests, if this is a feature initially added only to suricata 9.

jufajardini (changes requested):
Thanks for the tests!

We do have to add the `min-version` requirement before they can be added to SV, though, or this would fail the suricata 8 branch CI.

tests/detect-ldap-bind-auth/test.yaml
0x-0ddc0de: removed

tests/detect-ldap-bind-auth/test.yaml
0x-0ddc0de: added

0x-0ddc0de:
superseded by https://github.com/OISF/suricata-verify/pull/3357

PR #3364, opened 2026-09-15
security-8642: use new max value

glongo:
## Ticket

changes to the ticket status:

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8914


jasonish (approved):

catenacyber:
This looks wrong as it does not test the previous bug

jasonish:
> This looks wrong as it does not test the previous bug

Do we need 2 tests?

catenacyber:
> > This looks wrong as it does not test the previous bug

Confirmed, suricata at commit 52490b8c26909772801afed98df63a6d62f3af18 just before fix d9fae18432501674338e28bd33e35788c95d4a98 passes this new test, but crashes the current one

> 
> Do we need 2 tests?

I am afraid so

catenacyber (changes requested):
I propose https://github.com/OISF/suricata-verify/pull/3365 instead to keep the original test for 8.0x with x<7

catenacyber:
Replaced by #3366

PR #3339, opened 2026-08-31
tests: remove 7.0 only tests and checks

jasonish:
- **tests: remove tests for versions less than 8**
- **tests: remove checks targeting Suricata versions < 8**

PR #3365, opened 2026-09-15
swf: split test before 8.0.7 and post 8.0.7

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/8642

https://github.com/OISF/suricata-verify/pull/3364 alternative


victorjulien:
I don't see the point of this, why would we still have <8.0.7, something we'll never exercise?

catenacyber:
That is what Jason did in https://github.com/OISF/suricata-verify/pull/3339

Alternative is to completely remove the test

catenacyber:
> I don't see the point of this, why would we still have <8.0.7, something we'll never exercise?

Do you like https://github.com/OISF/suricata-verify/pull/3366 better ?

jasonish:
> That is what Jason did in #3339

Disagree. I didn't retarget any tests to be < 8.0.7. I only targeted tests that were less than 8. Any that were left as < 8.0.7 were as they were before.

catenacyber:
> Any that were left as < 8.0.7 were as they were before.

Yes that is what I mean.

You are keeping tests in https://github.com/OISF/suricata-verify/pull/3339 that are never run in current CI

I am fine with either approach : always skipping tests that are obsolete, or keep the 8.0.x history

catenacyber:
As you changed your #3339, I am closing this PR in favor of #3366

PR #3375, opened 2026-09-21
Http1 100 continue/v2a

victorjulien:
Mishandling of http1 status line if a client expects a 100 continue and the server provides it.

PR #3381, opened 2026-09-22
Http1 continue progress 9096 v1

catenacyber:
## Ticket

Redmine ticket: https://redmine.openinfosecfoundation.org/issues/9096

https://github.com/OISF/suricata-verify/pull/3375 fixed test cases


catenacyber:
Note that in the supplied pcaps, the client does not send its data after getting the 100 continue

